PR #9, opened 2020-12-28
Add secretsmanager-jdbc

eddumelendez:
Provide dependency management for `aws-secretsmanager-jdbc` and
`aws-secretsmanager-caching-java`. Also, docs are updated.

### DIFF
--- a/docs/src/main/asciidoc/rds.adoc
+++ b/docs/src/main/asciidoc/rds.adoc
@@ -408,3 +408,43 @@ Sample IAM policy granting access to RDS:
         },
 }
 ----
+
+=== SecretsManager credentials
+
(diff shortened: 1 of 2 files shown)


maciejwalkowiak (commented):
I am not convinced we should add it at all since do not provide anything except saving users from specifying a dependency version. Since it's in the RDS section it could be even misleading as secretsmanager-jdbc module will not work with Spring Cloud AWS RDS integration. 

So either we should be very explicit about incompatibilities or just not add it at all.

eddumelendez:
what about a new jdbc section?

maciejwalkowiak:
Sorry, it is actually "Data access with JDBC section". Still, I believe that if we add it we need to be very explicit about incompatibilities to save users from trying to get both jdbc secrets manager together with read replicas or failover.

PR #56, opened 2021-02-12
Auto-configure credentials and region with AWS SDK v2.

### LINKED ISSUE #54: Autoconfigure CredentialsProvider and RegionProvider with AWS SDK v2
Auto-configure module should auto-configure `AwsCredentialsProvider` and `AwsRegionProvider` with AWS SDK v2. Feature wise - same logic and same options should be provided as there are in Spring Cloud AWS v2.3.

maciejwalkowiak:
Fixes #54

### DIFF
--- a/v3/spring-cloud-aws-v3-autoconfigure/src/main/java/io/awspring/cloud/v3/autoconfigure/RegionProviderAutoConfiguration.java
+++ b/v3/spring-cloud-aws-v3-autoconfigure/src/main/java/io/awspring/cloud/v3/autoconfigure/RegionProviderAutoConfiguration.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2013-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.awspring.cloud.v3.autoconfigure;
+
+import io.awspring.cloud.v3.autoconfigure.properties.AwsRegionProperties;
+import io.awspring.cloud.v3.core.region.StaticRegionProvider;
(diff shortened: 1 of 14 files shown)


v3/spring-cloud-aws-v3-autoconfigure/src/main/java/io/awspring/cloud/v3/autoconfigure/RegionProviderAutoConfiguration.java
```diff
@@ -0,0 +1,56 @@
...
+	private final AwsRegionProperties properties;
+
+	public RegionProviderAutoConfiguration(AwsRegionProperties properties) {
+		this.properties = properties;
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public AwsRegionProvider awsRegionProvider() {
+		if (properties.isStatic()) {
```
maciejwalkowiak, line 48: Similar to credentials, we should also support profile region provider.

arunkpatra (commented):
@maciejwalkowiak 
You might need to add `CredentialsProviderAutoConfiguration` and `RegionProviderAutoConfiguration` to a `spring.factories` file in the classpath. 

PS: The file exists but probably we should not be adding to the existing `spring.factories` file for now. Once V3 stabilizes, probably then. Until then a new file should be good in `spring-cloud-aws-v3-autoconfigure/src/main/resources/META-INF/spring.factories` IMO:

```
org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
io.awspring.cloud.v3.autoconfigure.CredentialsProviderAutoConfiguration,\
io.awspring.cloud.v3.autoconfigure.RegionProviderAutoConfiguration
```

`RegionProviderAutoConfigurationTests#customRegionConfigured` passes right now because of the way the auto-config is loaded right now:

```
private final ApplicationContextRunner contextRunner = new ApplicationContextRunner()
			.withConfiguration(AutoConfigurations.of(CredentialsProviderAutoConfiguration.class));
```

But in the general case when Spring Boot apps will load the auto-config just by virtue of the auto-config class being in the class path (standard auto-config design), this has given me problems without the registration in the spring.factories file earlier. If some lower level infra code in Spring Boot has altered this behavior, then probably I am wrong on my suggestions above and you won't need the spring.factories way of registration.

eddumelendez:
I have created a [discussion](https://github.com/awspring/spring-cloud-aws/discussions/63) related to spring cloud aws 3 and moving to aws sdk v2.

maciejwalkowiak:
Thanks @arunkpatra for a review! I forgot about spring.factories, good catch!

PR #87, opened 2021-02-25
Fixes #81 by reading the MessageGroupId out of the attibutes

### LINKED ISSUE #81: FIFO queues support does not guarantee FIFO order
Under the hood, SQS integration uses `AmazonSQSBufferedAsyncClient` to reduce amount of requests sent to AWS and same time reduce the costs. Unfortunately, [it does not support FIFO queues](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-client-side-buffering-request-batching.html). 

Ideally in case of FIFO queue we should fall back to regular SQS client.

tekener:
It is now reading the MessageGroupId out of the attribute. Had to fix the test as well because it was using a ArrayList in a multithreaded environment. Because of that the tests were sometimes failing.

- [ x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Will read the MessageGroupId out of the attributes because it is not part of the MessageAttributes

NullpointerException as described in https://github.com/awspring/spring-cloud-aws/issues/81

Modified the existing test.

## :pencil: Checklist

- [x ] I added tests to verify changes
- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/spring-cloud-aws-messaging/src/main/java/io/awspring/cloud/messaging/listener/SimpleMessageListenerContainer.java
+++ b/spring-cloud-aws-messaging/src/main/java/io/awspring/cloud/messaging/listener/SimpleMessageListenerContainer.java
@@ -373,7 +373,7 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 
 		private List<MessageGroup> groupByMessageGroupId(final ReceiveMessageResult receiveMessageResult) {
 			return receiveMessageResult.getMessages().stream()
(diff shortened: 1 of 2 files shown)


fmalhe:
Any plans to include this in a RC3 build or something like that?
Some of my production services would benefit greatly from this fix (currently i have a hack in place to read only one msg at the time from fifo queues)

maciejwalkowiak (approved):

PR #93, opened 2021-03-01
Mention in docs a need to use not-buffered SQS client with FIFO queues.

### LINKED ISSUE #81: FIFO queues support does not guarantee FIFO order

maciejwalkowiak:
Fixes #81

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -45,9 +45,9 @@ build.
 
 Or reach out directly to individual team members:
 
-* Maciej Walkowiak (https://twitter.com/maciejwalkowiak[Twitter]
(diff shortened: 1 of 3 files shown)


eddumelendez (approved):
Let's make this PR just change `docs/src/main/asciidoc/sqs.adoc`

PR #113, opened 2021-04-06
Provide ClientHttpRequestInterceptor to make aws-sig-4 signed request with the RestTemplate

joshiste:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Provide a ClientHttpRequestInterceptor which signs requests using the AWS Signature V4 which is needed for example to call various AWS services (like API Gateway)

Currently imho it is tedious and lacking a good documentation on how to make signed request using the RestTemplate to the API Gateway. 
Alternative to the RestTemplate you can use generated SDKs or subclassing the `AmazonWebServiceClient`.

Imho being able to make signed requests using the RestTemplate has the benefit for the spring developers to work with a familiar API.

There are two test verifying the signatures of requests from the aws-sig-v4-test-suite.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
As this is more or less just a preview of the feature and also lacking support for the WebClient, I am happy to discuss if this is a desirable feature and if yes what further steps are needed to see this feature integrated.

### DIFF
--- a/spring-cloud-aws-web/src/main/java/org/springframework/cloud/aws/web/RequestSigner.java
+++ b/spring-cloud-aws-web/src/main/java/org/springframework/cloud/aws/web/RequestSigner.java
@@ -1,0 +1,29 @@
+package org.springframework.cloud.aws.web;
+
+import com.amazonaws.auth.AWS4Signer;
+import com.amazonaws.auth.AWSCredentialsProvider;
+import org.springframework.http.HttpRequest;
+
+public class RequestSigner {
+	private final String regionName;
+	private final String serviceName;
+	private final AWSCredentialsProvider awsCredentialsProvider;
+
+	public RequestSigner(AWSCredentialsProvider awsCredentialsProvider, String regionName, String serviceName) {
+		this.regionName = regionName;
+		this.serviceName = serviceName;
+		this.awsCredentialsProvider = awsCredentialsProvider;
+	}
+
+	public void signRequest(HttpRequest request, byte[] body) {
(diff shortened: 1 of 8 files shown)


spring-cloud-aws-web/src/main/java/org/springframework/cloud/aws/web/RequestSigner.java
```diff
@@ -0,0 +1,29 @@
...
+		this.awsCredentialsProvider = awsCredentialsProvider;
+	}
+
+	public void signRequest(HttpRequest request, byte[] body) {
+		AWS4Signer signer = this.createAws4Signer();
+		signer.sign(new SpringSignableRequest(request, body), this.awsCredentialsProvider.getCredentials());
+	}
+
+	protected AWS4Signer createAws4Signer() {
+		AWS4Signer signer = new AWS4Signer(false);
```
maciejwalkowiak, line 24: Why `doubleUrlEncoding` is set to `false`? What are actually use cases for both values?

maciejwalkowiak (commented):
Thanks @joshiste for the PR! Indeed I remember it was a pain to call AWS Elasticsearch from Spring app with `RestTemplate`. This out-of-the-box support makes it easier. 

I left one question plus we need also reference docs for new module. Would you like to contribute it too?

maciejwalkowiak:
Closed due to lack of feedback. We can get back to this topic if there is a will from the community.

Issue #122, opened 2021-04-25
QueueMessageChannel.receive() implementation doesn't align with PollableChannel 's definition


jorgenota:
**Type**: Bug

**Component**:
SQS

**Describe the bug**

- According to javadoc of PollableChannel, MessageChannel.INDEFINITE_TIMEOUT should be used to represent indefinite timeout (instead of 0).
- The timeout parameter of PollableChannel.receive() method represents the timeout in milliseconds (so it should be transformed into seconds when building the ReceiveMessageRequest).

jorgenota:
Suggested fix in PR #123

maciejwalkowiak:
As mentioned in #123 - we are not going to fix it to not introduce breaking change.

PR #123, opened 2021-04-25
Fix timeout handling in QueueMessageChannel

jorgenota:
The timeout parameter of PollableChannel.receive() method represents the timeout in milliseconds (so it should be transformed into seconds when building the ReceiveMessageRequest). Besides, it should use MessageChannel.INDEFINITE_TIMEOUT to represent indefinite timeout (not 0).

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Minor changes to better handle timeout parameter when receiving messages from an SQS queue.

I suggest this fixes to align with PollableChannel interface:
- According to javadoc of PollableChannel, `MessageChannel.INDEFINITE_TIMEOUT` should be used to represent indefinite timeout (instead of 0). Besides, no args `PollableChannel.receive()` method should "block indefinitely if necessary".
- The `timeout` parameter of `PollableChannel.receive(timeout)` method represents the timeout in milliseconds (so it should be transformed into seconds when building the ReceiveMessageRequest).

Fixes issue #122 

## :pencil: Checklist

- [X ] I reviewed submitted code
- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/spring-cloud-aws-messaging/src/main/java/io/awspring/cloud/messaging/core/QueueMessageChannel.java
+++ b/spring-cloud-aws-messaging/src/main/java/io/awspring/cloud/messaging/core/QueueMessageChannel.java
@@ -34,6 +34,7 @@ import com.amazonaws.services.sqs.model.SendMessageRequest;
 import com.amazonaws.services.sqs.model.SendMessageResult;
 
 import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessageDeliveryException;
 import org.springframework.messaging.MessageHeaders;
 import org.springframework.messaging.PollableChannel;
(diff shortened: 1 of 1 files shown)


maciejwalkowiak (commented):
Thanks @jorgenota for the PR. Since the unit changes, we need to consider this as a breaking change. 

According to SQS spec, to achieve long polling, `receiveMessage` should have `waitTimeoutSeconds` set to `0` - which should be the case when `receive` is called with `MessageChannel.INDEFINITE_TIMEOUT`, or am I missing something?

jorgenota:
> According to SQS spec, to achieve long polling, `receiveMessage` should have `waitTimeoutSeconds` set to `0` - which should be the case when `receive` is called with `MessageChannel.INDEFINITE_TIMEOUT`, or am I missing something?

The problem is that, according to the javadoc of `PollableChannel.receive(timeout)` you should use the value `MessageChannel.INDEFINITE_TIMEOUT` (i.e. -1) to set an indefinite timeout. Then, the implementation of `QueueMessageChannel` should translate this (-1) into whatever SQS requires (in this case, a 0).

maciejwalkowiak:
Sorry for late reply.

That makes sense, but I don't see this translation from -1 into 0..? Changing seconds to seconds to millis - good catch - it's a breaking change though, we would have to wait with that likely till 3.0

jorgenota:
Regarding the -1 and 0: the translation is basically between Spring's API and Amazon's API. The first one uses -1 and the latter 0.

maciejwalkowiak:
@jorgenota apologies for a delay. We do not want to introduce breaking changes for versions that use AWS SDK v1. We will keep your change in mind for 3.0. Thanks!

PR #138, opened 2021-05-20
Execute all received sqs messages

LeonieAdis:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

After the SimpleMessageListenerContainer has received SQS messages from a queue it does not check again, that the queue is still running and delivers all messages to their handlers even if the queue has stopped in the meantime.

If a SimpleMessageListenerContainer is stopped, e.g. because the service instance is shut down, it might happen that while stopping, the container still receives some SQS messages from a queue, but doesn't distribute those messages to their handlers.
As a result, the SQS messages are invisible for other service instances that are listening to the same queue until the visibility timeout of the messages is reached. This causes a delay of the message delivery.
With this fix, all messages that are pulled from the queue - and thus set invisible - are delivered to their handlers and can therefore be processed, even if the  SimpleMessageListenerContainer is stopped or the service is shut down.

Added a unit test:
- Start a SimpleMessageListenerContainer
- Send a SQS message
- Stop the SimpleMessageListenerContainer
- Verify that the message was received by its handler

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


maciejwalkowiak (approved):
Looks good to me, thanks for submitting PR!

@eddumelendez would you like to take a look?

birnbuazn:
Any news on this?

PR #147, opened 2021-06-09
fix prefix not being validated for other slashes

### LINKED ISSUE #146: Secrets Manager aws.secretsmanager.prefix does not allow addition segments
**Type**: Bug

**Component**:
Secrets Manager

**Describe the bug**
The old library and current documentation says:

Prefix indicating first level for every property loaded from the Secrets Manager. Value must start with a forward slash followed by one or more valid path segments or be empty.

The new regex does not allow for additional path segments, only validates that it starts with slash:
Old regex: (/[a-zA-Z0-9.\\-_]+)*
New regex: (/)?([a-zA-Z0-9.\\-_]+)*

I have not validated, but I would guess that Parameter Store would have the same issue.

**Sample**
In this impl only "/foo" validates, not "/foo/bar/baz".

MatejNedic:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
FIxes #146

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/spring-cloud-aws-parameter-store-config/src/main/java/io/awspring/cloud/paramstore/AwsParamStoreProperties.java
+++ b/spring-cloud-aws-parameter-store-config/src/main/java/io/awspring/cloud/paramstore/AwsParamStoreProperties.java
@@ -43,7 +43,7 @@ public class AwsParamStoreProperties implements Validator {
 	/**
 	 * Pattern used for prefix validation.
 	 */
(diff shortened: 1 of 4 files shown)


maciejwalkowiak (approved):
Thanks a lot @MatejNedic!

maciejwalkowiak:
@MatejNedic can you fix the build?

PR #152, opened 2021-06-17
gh-151 parameter store fix

### LINKED ISSUE #151: Spring Boot 2.4.7 issue when spring.config.import is used in application-{profile}.yaml
**Type**: Bug

**Component**:
Parameter Store

**Describe the bug**
After the update from spring boot 2.4.6 to 2.4.7 this kind of import inside a application-{profile}.yaml does not work anymore.
```
spring:
  config:
    import: 'aws-parameterstore:/config/my-service/'
```

Spring does now only call the `resolve` method on the `AwsParamStoreConfigDataLocationResolver` and not the profile specific `resolveProfileSpecific` one.

It i working when you are doing this in the application.yaml because then the `resolveProfileSpecific` method is then called passing in the profiles.
Tried this workaround but then my local profile without aws access also tries to connect to aws and the spring boot startup will fail.
Setting the import to optional does also not help, because the aws call happens anyway.

tekener:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
With the spring boot 2.4.7 update the way the ConfigDataLocationResolver works has changed. When doing the import in application-{profile}.properties the resolve method is called instead the resolveProfileSpecific one

Fixes #151 

Via unit and used it in own project

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/spring-cloud-starter-aws-parameter-store-config/src/main/java/io/awspring/cloud/autoconfigure/paramstore/AwsParamStoreConfigDataLocationResolver.java
+++ b/spring-cloud-starter-aws-parameter-store-config/src/main/java/io/awspring/cloud/autoconfigure/paramstore/AwsParamStoreConfigDataLocationResolver.java
@@ -75,30 +75,30 @@ public class AwsParamStoreConfigDataLocationResolver
 		return resolve(resolverContext, location, profiles);
 	}
 
-	private List<AwsParamStoreConfigDataResource> resolve(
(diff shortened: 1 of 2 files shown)


maciejwalkowiak:
Thanks @tekener for PR. This issue has been fixed in Spring Boot 2.4.8 (https://github.com/spring-projects/spring-boot/issues/26960)

Issue #124, opened 2021-04-28
Only process FIFO queue messages with same MessageGroupId if previous ones were acknowledged


sebastianistoblame:
**Type**: Feature

**Is your feature request related to a problem? Please describe.**
When receiving FIFO queue messages as batches of more than 1, it can happen that a newer message with the same `MessageGroupId` is processed before older ones are acknowledged.

Lets consider this example: We have two messages: A, B
A is first in the FIFO queue, B second.

Let's set our Deletion Policy to `NEVER` to manually acknowledge our messages. Now put both message A and B into the queue and receive them with maximum number of messages set to 10. Currently, if message A is not acknowledged, message B will be processed.

**Describe the solution you'd like**
I would like, that if message A is not acknowledged, message B will not be processed. Sure, setting the max. number of messages to receive to 1 will work. But I want to avoid using a value of 1 in order to reduce the number of calls made to AWS while still retaining correct ordering.

I'd like to have a configuration flag here. By turning it on, Spring AWS Messaging will not process messages, that have an identical `MessageGroupId`, if an earlier message with this `MessageGroupId` in the same batch was not acknowledged.

**Describe alternatives you've considered**
The downside of my proposed solution is, that you are moving more FIFO logic into the consumer. Therefore I can understand if you choose not to implement this. But in this case, I'd like to have a warning if Spring Cloud AWS messaging is configured with `maxNumberOfMessages` > 1 in combination with a FIFO queue.

**Additional context**
Here is my `AmazonSQSAsync` config:
```
  @Bean
  @Primary
  public AmazonSQSAsync amazonSQSClient() {
    return AmazonSQSAsyncClientBuilder.standard().build();
  }
```

Here is my `SimpleMessageListenerContainerFactory` config:
```
  @Bean
  public SimpleMessageListenerContainerFactory messageListenerContainerFactory(AmazonSQSAsync amazonSQSClient) {
    var factory = new SimpleMessageListenerContainerFactory();
    factory.setAmazonSqs(amazonSQSClient);
    factory.setMaxNumberOfMessages(10);
    factory.setWaitTimeOut(10);
    factory.setBackOffTime(1000L);
    return factory;
  }
```

sdusza1:
I've been analyzing the same issue. I think the problem lies in `MessageGroupExecutor.run()`. 

There when an exception is caught the loop continues to process messages from the group.
And I think it shouldn't continue so that the next time messages from this group are recieved 
they should be processed in the same order.

[...]

PR #184, opened 2021-09-02
Stop processing messages from FIFO queue when exception is thrown

sebastiandusza:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

I think there is a bug in `SimpleMessageListenerContainer.MessageGroupExecutor` class. 
When multiple messages from the same message group are recieved and exception is thrown while
processing one them the others should not be processed. I think this is importand because 
the order should be maintained even in case of an error.

I think it solves this issue
https://github.com/awspring/spring-cloud-aws/issues/124

I integrated it with my app that is using SQS FIFO queue and checked if order is maintained in case of an exception.
Also I prepared a unit test.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes <- not sure about that.

### DIFF
--- a/spring-cloud-aws-messaging/src/main/java/io/awspring/cloud/messaging/listener/SimpleMessageListenerContainer.java
+++ b/spring-cloud-aws-messaging/src/main/java/io/awspring/cloud/messaging/listener/SimpleMessageListenerContainer.java
@@ -428,6 +428,7 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 				}
 				catch (MessagingException messagingException) {
 					applyDeletionPolicyOnError(receiptHandle);
+					break;
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


spring-cloud-aws-messaging/src/main/java/io/awspring/cloud/messaging/listener/SimpleMessageListenerContainer.java
```diff
@@ -428,6 +428,7 @@ public void run() {
 				}
 				catch (MessagingException messagingException) {
 					applyDeletionPolicyOnError(receiptHandle);
+					break;
```
sdusza1, line 431: I'm wondering if this break should be executed only when `this.deletionPolicy == ON_SUCCESS` ?

maciejwalkowiak:
Thanks @sebastiandusza and apologies for late review. I am closing this as #192 is already in place and it has a test.

PR #195, opened 2021-10-20
Fix context order when loading properties

### LINKED ISSUE #169: spring-cloud-aws-parameter-store-config prioritizes default-context parameters over service-specific parameters
**Type**: Bug

**Component**: Parameter Store

**Describe the bug**
Spring Cloud Version: 2.3.1
In parameter store we have two parameters with names:
1. /config/application/xxx
2. /config/my-service/xxx

In our spring application.properties we have the following configuration:
`spring.application.name=my-service`

When we start the application it is the parameter from (1.) that is set, even though we would expect (2.) to override the default-context's parameter. 
Note: If we change `aws.paramstore.default-context` to something else than _application_ it will use the value from (2.).

From the description of default-context it says:
> Shared by all services that have the Configuration support enabled. Can be overridden with a service- or profile-specific property.

So we would expect (2.) to override (1.), but this is not the case?

WtfJoke:
Fixes gh-169

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Reverse the context order when loading properties to preserve documented order (eg service specific context "wins" over more general application context)

Fixes #169

By JUnit Test(s) and in a sample project [WtfJoke/chaoskotlindemo/awsparamterstoreconfig](https://github.com/WtfJoke/chaoskotlindemo/tree/awsparamterstoreconfig)

## :pencil: Checklist

- [N/A] I updated reference documentation to reflect the change
- [x] All tests passing
- [ ] No breaking changes - well its somehow breaking as it reverse the order, but imho never actually worked like intended since the (unwanted) change

### DIFF
--- a/spring-cloud-aws-parameter-store-config/src/main/java/io/awspring/cloud/paramstore/AwsParamStorePropertySources.java
+++ b/spring-cloud-aws-parameter-store-config/src/main/java/io/awspring/cloud/paramstore/AwsParamStorePropertySources.java
@@ -25,7 +25,11 @@ import org.apache.commons.logging.Log;
 import org.springframework.util.StringUtils;
 
 /**
+ * Is responsible for creating {@link AwsParamStorePropertySource} and determining
+ * automatic contexts.
+ *
  * @author Eddú Meléndez
+ * @author Manuel Wessner
  * @since 2.3
  */
 public class AwsParamStorePropertySources {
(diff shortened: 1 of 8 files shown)


piotrkorlaga:
Hello, any plans for releasing this fix, soon? :)

maciejwalkowiak (approved):
Thanks @WtfJoke for a fix! And apologies for late merge.

WtfJoke:
Glad it got merged eventually :) Thanks for taking care!

Issue #167, opened 2021-07-21
Prefix for paramstore cannot be empty


maciejwalkowiak:
From spring-cloud-aws created by [JohannesZick](https://github.com/JohannesZick): spring-cloud/spring-cloud-aws#776  

**Type**: Bug

**Component**:
Parameter Store

**Describe the bug**
According to the documentation, the prefix can be either a valid path or empty. This is great, because I want to only have the application name, no static prefix. But in reality, there is a validation that rejects `null` and empty strings.

**Sample**
Simply use `aws.paramstore.prefix: ""` in `bootstrap.yml`. The validation is also easy to find in the `AwsParamStoreProperties.java`, [line 96](https://github.com/spring-cloud/spring-cloud-aws/blob/main/spring-cloud-aws-parameter-store-config/src/main/java/org/springframework/cloud/aws/paramstore/AwsParamStoreProperties.java#L96).

maciejwalkowiak:
Either fix docs, or allow empty (if AWS allows it too).

menneske2:
Bump

PR #216, opened 2022-01-12
167 - Prefix for paramstore cannot be empty

menneske2:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Allow empty values for aws.paramstore.prefix.

Fixes bug #167 (https://github.com/awspring/spring-cloud-aws/issues/167)
Necessary fix for certain setups to be able to update aws parameter store version.

Only local tests.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
Merge this bad boy so I can update the general aws version in my company.

### DIFF
--- a/spring-cloud-aws-parameter-store-config/src/main/java/io/awspring/cloud/paramstore/AwsParamStoreProperties.java
+++ b/spring-cloud-aws-parameter-store-config/src/main/java/io/awspring/cloud/paramstore/AwsParamStoreProperties.java
@@ -84,10 +84,6 @@ public class AwsParamStoreProperties implements InitializingBean {
 
 	public void afterPropertiesSet() throws Exception {
 
-		if (!StringUtils.hasLength(prefix)) {
-			throw new ValidationException(CONFIG_PREFIX + ".prefix", "prefix should not be empty or null.");
-		}
-
 		if (!StringUtils.hasLength(defaultContext)) {
 			throw new ValidationException(CONFIG_PREFIX + ".defaultContext",
 					"defaultContext should not be empty or null.");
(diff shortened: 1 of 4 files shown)


MatejNedic (approved):
Hey, @menneske2 tnx on contribution and sorry for the late response.  Looks good to me!

spring-cloud-aws-parameter-store-config/src/main/java/io/awspring/cloud/paramstore/AwsParamStoreProperties.java
```diff
@@ -98,7 +94,7 @@ public void afterPropertiesSet() throws Exception {
 					"profileSeparator should not be empty or null.");
 		}
 
-		if (!PREFIX_PATTERN.matcher(prefix).matches()) {
+		if (StringUtils.hasLength(prefix) && !PREFIX_PATTERN.matcher(prefix).matches()) {
```
MatejNedic, line 97: Upon further inspection please change to prefix != null &&

MatejNedic (changes requested):
This will create some issues in `AwsParamStorePropertySources.class` currently method  `	protected String getContext(String prefix, String context)` uses `StringUtils.hasLength(prefix)` for empty string this will be ignored tho (false) and context will be returned. The problem with that is that one of the lookups would then be just defaultContext without the leading slash (also defaultContext/profileActive). Please change code there to `prefix != null` and add further tests in `AwsParamStorePropertySourceLocatorTest`.

menneske2:
thanks for the fix. any idea when 2.3.4 rolls out?

maciejwalkowiak:
aiming for this week. i think we are very close to finish everything we planned.

PR #237, opened 2022-02-11
Add cloudwatch metrics auto-configuration with aws sdk v2

eddumelendez:
As part of the effort to move to aws-sdk-v2. Metrics has been added using
`micrometer-registry-cloudwatch2` as a auto-configuration.


v3/spring-cloud-aws-v3-test/src/test/java/io/awspring/cloud/test/v3/cloudwatch/CloudwatchMetricsIntegrationTest.java
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.DynamicPropertyRegistry;
+import org.springframework.test.context.DynamicPropertySource;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.given;
+
+@SpringBootTest(properties = { "management.metrics.export.cloudwatch.namespace=awspring/spring-cloud-aws" })
+@Testcontainers(disabledWithoutDocker = true)
+class CloudwatchMetricsIntegrationTest {
```
maciejwalkowiak, line 40: For Parameter Store and Secrets Manager, integration tests (those that use localstack) are just placed in the modules related to this integration. 

If we would have integration tests like before, that hit real AWS i somewhat see a sense of having a separate module, but in this case, shouldn't it just go to `metrics` module?

v3/spring-cloud-aws-v3-autoconfigure/src/main/java/io/awspring/cloud/v3/autoconfigure/metrics/CloudWatchExportAutoConfiguration.java
```diff
@@ -0,0 +1,101 @@
...
+	@Bean
+	@ConditionalOnProperty(value = "management.metrics.export.cloudwatch.enabled", matchIfMissing = true)
+	public CloudWatchMeterRegistry cloudWatchMeterRegistry(CloudWatchConfig config, Clock clock,
+			CloudWatchAsyncClient client) {
+		return new CloudWatchMeterRegistry(config, clock, client);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public CloudWatchAsyncClient amazonCloudWatchAsync(AwsCredentialsProvider credentialsProvider,
```
maciejwalkowiak, line 79: perhaps bean name should be `cloudWatchClient`? or `cloudWatchAsyncClient`? We should avoid using "amazon", if we need it should be rather `AWS`, but definitely we need consistent naming for all client beans.

v3/spring-cloud-aws-v3-test/src/test/java/io/awspring/cloud/test/v3/cloudwatch/CloudwatchMetricsIntegrationTest.java
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	@Autowired
+	private MeterRegistry meterRegistry;
+
+	@Test
+	void test() {
+		Counter testCounter = Counter.builder("test").tag("module", "cloudwatch").register(this.meterRegistry);
+		testCounter.increment();
+		given().await().atMost(Duration.ofSeconds(20))
```
maciejwalkowiak, line 62: Correct me if I am wrong but I think `testCounter.count()` does not actually go to CloudWatch to check the value of this metric? If we have an integration test like that, ideally i believe it should verify that CloudWatch received this increment.

v3/spring-cloud-starter-aws-v3-metrics/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		 xmlns="http://maven.apache.org/POM/4.0.0"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<artifactId>v3</artifactId>
+		<groupId>io.awspring.cloud</groupId>
+		<version>2.4.1-SNAPSHOT</version>
+	</parent>
+	<artifactId>spring-cloud-starter-aws-v3-metrics</artifactId>
```
maciejwalkowiak, line 27: Something for discussion - should not we call this module `cloudwatch` instead of `metrics`? Similar to how `messaging` module covered SQS and SNS because these were the only options back then for messaging in AWS, i believe we should have now SQS and SNS modules.

v3/spring-cloud-aws-v3-test/src/test/java/io/awspring/cloud/test/v3/cloudwatch/CloudwatchMetricsIntegrationTest.java
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.DynamicPropertyRegistry;
+import org.springframework.test.context.DynamicPropertySource;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.given;
+
+@SpringBootTest(properties = { "management.metrics.export.cloudwatch.namespace=awspring/spring-cloud-aws" })
+@Testcontainers(disabledWithoutDocker = true)
```
maciejwalkowiak, line 39: I would prefer this test to fail without docker than quietly pass.

v3/spring-cloud-starter-aws-v3-metrics/pom.xml
eddumelendez: Cloudwatch is a service which involves logs, metrics so in this case since the support it is specific to metrics I think make sense to keep it that way. The reasoning for spring-cloud-gcp metrics module was similar.

v3/spring-cloud-aws-v3-test/src/test/java/io/awspring/cloud/test/v3/cloudwatch/CloudwatchMetricsIntegrationTest.java
eddumelendez: I am following sqs in the `spring-cloud-aws-sqs` module. I think we should better discuss about how to handle this.

v3/spring-cloud-aws-v3-test/src/test/java/io/awspring/cloud/test/v3/cloudwatch/CloudwatchMetricsIntegrationTest.java
eddumelendez: for some reason, removing this is failing immediately. I haven't go deep into this but for now will keep it.

v3/spring-cloud-aws-v3-test/src/test/java/io/awspring/cloud/test/v3/cloudwatch/CloudwatchMetricsIntegrationTest.java
eddumelendez: that's right! I will submit changes involving request to aws cloudwatch instead.

v3/spring-cloud-aws-v3-autoconfigure/src/main/java/io/awspring/cloud/v3/autoconfigure/metrics/CloudWatchExportAutoConfiguration.java
eddumelendez: agree!

v3/spring-cloud-starter-aws-v3-metrics/pom.xml
maciejwalkowiak: I remain unconvinced but let it be ;-)

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/metrics/CloudWatchProperties.java
```diff
@@ -0,0 +1,80 @@
...
+ * {@link ConfigurationProperties} for configuring CloudWatch metrics export.
+ *
+ * @author Jon Schneider
+ * @author Dawid Kublik
+ * @author Bernardo Martins
+ * @author Eddú Meléndez
+ * @since 2.0.0
+ */
+@ConfigurationProperties(prefix = "management.metrics.export.cloudwatch")
+public class CloudWatchProperties extends StepRegistryProperties {
```
maciejwalkowiak, line 32: For the sake of consistency, shouldn't it bind to `spring.cloud.aws.cloudwatch` and extends `AwsClientProperties`? Similar to how Stackdriver is done in GCP, where `StackdriverPropertiesConfigAdapter` adapts GCP properties to Micrometer.

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/metrics/CloudWatchExportAutoConfiguration.java
```diff
@@ -0,0 +1,98 @@
...
+	@Bean
+	@ConditionalOnProperty(value = "management.metrics.export.cloudwatch.enabled", matchIfMissing = true)
+	public CloudWatchMeterRegistry cloudWatchMeterRegistry(CloudWatchConfig config, Clock clock,
+			CloudWatchAsyncClient client) {
+		return new CloudWatchMeterRegistry(config, clock, client);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public CloudWatchAsyncClient amazonCloudWatchAsync(AwsCredentialsProvider credentialsProvider,
```
maciejwalkowiak, line 76: Bean name should be `cloudWatchAsyncClient`

maciejwalkowiak:
I am not sure if this is how its done in Stackdriver integration in GCP. As far as I can see, in Spring Cloud GCP client can be configured with  `spring.cloud.gcp.metrics` (https://docs.spring.io/spring-cloud-gcp/docs/current/reference/html/index.html#spring-data-cloud-spanner) but Micrometer related properties go to `management.metrics.export.stackdriver`.

Shouldn't we have the same for Spring Cloud AWS? The benefit of configuring client through `spring.cloud.aws.cloudwatch` would be ability for `CloudWatchProperties` to extends `AwsClientProperties` which could be then configured with `AwsClientBuilderConfigurer`.

eddumelendez:
Spring boot provides auto-configuration for stackdriver and other but not for cloudwatch v1 nor v2 and `management.metrics.export.stackdriver` works for those actuator auto-configurations. In GCP case, it the auto-configuration provided by gcp executes before the one for spring-boot, providing the information needed.

This PR follows the the current implementation in previous version but I was also thinking about give the shape similar to what GCP does but then I realized that we will be providing two auto-configurations in the same place, one similar to the one that provides spring boot for stackdriver and other similar to the one that spring-cloud-gcp does. I would prefer to keep it as previous version on spring-cloud aws because everything is already powered by spring boot. Let me know your thought on this.

maciejwalkowiak:
Do you know why Spring Cloud GCP provides its own Stackdriver autoconfiguration if its already covered in Spring Boot?

I think it would be nice that `CloudWatchProperties` extend `AwsClientProperties` so that we can have same ways to configure all AWS clients properties.

eddumelendez:
spring boot stackdriver auto-configuration needs a project id and credentials to be setup manually meanwhile spring cloud gcp does that automatically taking into account global properties for the attributes mentioned at the beginning.

CloudWatchPropeties doesn't extend AwsProperties because it already extends StepRegistryProperties but region and endpoint are already provided in order to keep consistency with other autoconfigurations.

maciejwalkowiak:
Look at the commit I've added as i think its easier to discuss the actual code - when client properties are separated from metrics properties, we can use `AwsClientBuilderConfigurer` and keep all clients configuration consistent - not saying that it must be like that, but I don't see any drawbacks.

maciejwalkowiak:
Thanks @eddumelendez. Docs are pending but to avoid efforts for keeping this branch up to date we can add docs in separate PR, just has to be done before we release M2.

PR #246, opened 2022-02-16
Align Parameter Store with 2.4.x and polish

maciejwalkowiak:
- adds integration tests for Parameter Store integration
- removes an option to disable integration (it did not work anyway)
- resolves credentials from `CredentialsProperties`
- adds option to overwrite `AwsCredentialProvider` in bootstrap
- adds user friendly error message when keys are not specified (`spring.config.import=aws-parameterstore:`)

TODO:
- [x] javadocs

Issue #108, opened 2021-03-31
Application cannot register SSM client


kschlesselmann:
**Type**: Bug

**Component**:
Parameter Store

**Describe the bug**
Currently we try to migrate our services to Spring Boot 2.4.4, Spring Cloud 2020.0.2 and Spring Cloud AWS 2.3.
As far as I understand the bootstrap context has been deprecated by Spring Cloud Commons and the new way to include configuration via AWS Parameter Store would be
```
spring:
  config:
    import: "aws-parameterstore:"
```
which we have in `application-aws.yml`.

Now after deployment we currently end up with
```
java.lang.IllegalStateException: Could not register object [com.amazonaws.services.simplesystemsmanagement.AWSSimpleSystemsManagementClient@785a4557] under bean name 'configDataAWSSimpleSystemsManagement': there is already object [com.amazonaws.services.simplesystemsmanagement.AWSSimpleSystemsManagementClient@785a4557] bound
    at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.registerSingleton(DefaultSingletonBeanRegistry.java:124)
    at org.springframework.beans.factory.support.DefaultListableBeanFactory.registerSingleton(DefaultListableBeanFactory.java:1138)
    at io.awspring.cloud.autoconfigure.paramstore.AwsParamStoreConfigDataLocationResolver.lambda$registerAndPromoteBean$1(AwsParamStoreConfigDataLocationResolver.java:106)
    at org.springframework.context.event.SimpleApplicationEventMulticaster.doInvokeListener(SimpleApplicationEventMulticaster.java:176)
    at org.springframework.context.event.SimpleApplicationEventMulticaster.invokeListener(SimpleApplicationEventMulticaster.java:169)
    at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent(SimpleApplicationEventMulticaster.java:143)
    at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent(SimpleApplicationEventMulticaster.java:131)
    at org.springframework.boot.DefaultBootstrapContext.close(DefaultBootstrapContext.java:133)
    at org.springframework.boot.SpringApplication.prepareContext(SpringApplication.java:392)
    at org.springframework.boot.SpringApplication.run(SpringApplication.java:325)
    at org.springframework.boot.builder.SpringApplicationBuilder.run(SpringApplicationBuilder.java:144)
[...]

PR #249, opened 2022-02-21
Align Secrets Manager with 2.4 and polish

maciejwalkowiak:
Similar to #246.


v3/spring-cloud-aws-v3-autoconfigure/src/main/java/io/awspring/cloud/v3/autoconfigure/secretsmanager/SecretsManagerConfigDataLocationResolver.java
```diff
@@ -82,22 +74,29 @@ public boolean isResolvable(ConfigDataLocationResolverContext context, ConfigDat
...
+
 		return locations;
 	}
 
 	private List<String> getCustomContexts(String keys) {
-		if (!StringUtils.isEmpty(keys)) {
+		if (StringUtils.hasLength(keys)) {
 			return Arrays.asList(keys.split(";"));
 		}
 		return Collections.emptyList();
```
MatejNedic, line 102: 2 lines below on line 104 just add a comment. Can't pin point location since review can't be left on collapsed unmodified lines

`
/**
	 * Since hook can be activated more then one time, ApplicationContext needs to be
	 * checked if bean is already registered to prevent Exception. See issue #108 for more
	 * information.
	 */
`

MatejNedic (commented):
LGM just one small comment missing.

Issue #131, opened 2021-05-06
Use Java 1.6+ MimeTypesFiletypeMap for SimpleStorageResource content type derivation


chrisrhut:
**Type**: Feature

**Is your feature request related to a problem? Please describe.**
We would like to be able to take advantage of this library's `WritableResource` creation by just using a `s3://` URL. Unfortunately there's not quite enough flexibility in determining the content type. [The code in SimpleStorageNameUtils](https://github.com/awspring/spring-cloud-aws/blob/94ad5ac4015be5491e0dc29d551d06531d75e28a/spring-cloud-aws-core/src/main/java/io/awspring/cloud/core/io/s3/SimpleStorageNameUtils.java#L106) makes a call to `URLConnection#guessContentTypeFromName()` which is [difficult to extend](https://docs.oracle.com/javase/8/docs/api/java/net/URLConnection.html#getFileNameMap--).

As a result, we have to manually create and upload using `AmazonS3` (or `TransferManager`).

As of Java 1.6, there is a facility called [MimeTypesFiletypeMap](https://docs.oracle.com/javase/8/docs/api/javax/activation/MimetypesFileTypeMap.html) which is similar but somewhat easier to customize. Any `META-INF/mime.types` file in your app classpath can augment the default types, and provide a mapping from file extension to content type.

This proposal is to augment (if not replace) the `URLConnection` call to make it easier to use user-defined mappings.

**Describe the solution you'd like**
Something like:

```java
static String getContentTypeFromLocation(String location) {
    String objectName = getObjectNameFromLocation(location);
    if (StringUtils.hasLength(objectName)) {
      String contentType = FileTypeMap.getDefaultFileTypeMap().getContentType(objectName);
      if (contentType == null) {
        contentType = URLConnection.guessContentTypeFromName(objectName);
      }

      return contentType;
    }
    return null;
}
```

**Additional context**
The `FileTypeMap` is in javax.activation API, which does seem to be included as a dependency for this project.

Assuming you agree with the general premise, I'd be happy to put together a PR for this.

maciejwalkowiak:
Thanks @chrisrhut for suggestion and apologies for late reply. I like this idea if especially if we make it in a backward compatible way - so that we don't break existing users - and also if we make using `javax.activation` API optional. Checking for class presence can be done with Spring `ClassUtils.isPresent(..)`. You have green light for preparing a PR 👍

chrisrhut:
[...]

PR #280, opened 2022-03-26
Implement WritableResource

maciejwalkowiak:
Make `S3Resource` implement `WritableResource`.

Under the hood, uses https://github.com/CI-CMG/aws-s3-outputstream for S3 `OutputStream` implementation. 

Thanks to flexible `ContentTypeResolver`, solves https://github.com/awspring/spring-cloud-aws/issues/131 by giving an option to provide custom implementation of how content type is resolved.

Dependency to `aws-s3-outputstream` project could be dropped once https://github.com/aws/aws-sdk-java-v2/issues/3128 is implemented.

Alternatives:
- implement our custom `S3OutputStream` (I would prefer not to)
- include code from `aws-s3-outputstream` in Spring Cloud AWS
- make `aws-s3-outputstream` optional and change `S3Resource` behavior to be writable only if this dependency is on the classpath
- create an abstraction `S3OutputStream` interface and the way users can provide their own. For example they could choose if they want to upload to S3 in chunks, or first upload to a disk and then flush at once to S3.

Due to lack of support for object metadata in `aws-s3-outputstream` (https://github.com/CI-CMG/aws-s3-outputstream/issues/2), it's currently not possible to implement https://github.com/spring-attic/spring-cloud-aws/issues/104


maciejwalkowiak:
In the end I went with alternative 4: "create an abstraction S3OutputStream interface and the way users can provide their own. For example they could choose if they want to upload to S3 in chunks, or first upload to a disk and then flush at once to S3."

Currently there are two implementations:

- disk buffering - inspired by https://github.com/Alluxio/alluxio/blob/master/underfs/s3a/src/main/java/alluxio/underfs/s3a/S3AOutputStream.java
- multipart - that uses  https://github.com/CI-CMG/aws-s3-outputstream

Currently both lack support for metadata and simple way to configure content type resolution.

PR #282, opened 2022-04-01
Cleanup pom files.

maciejwalkowiak:
- remove license from all `pom.xml` files use instead `<license>` tag in main pom (similar to how Spring Cloud GCP does)
- remove `<description/>` and add `<name/>` in each `pom.xml`
- polishing

### DIFF
--- a/spring-cloud-aws-autoconfigure/pom.xml
+++ b/spring-cloud-aws-autoconfigure/pom.xml
@@ -67,11 +67,6 @@
 			<artifactId>javax.mail-api</artifactId>
 			<optional>true</optional>
 		</dependency>
-		<dependency>
-			<groupId>software.amazon.awssdk</groupId>
-			<artifactId>ssm</artifactId>
-			<optional>true</optional>
-		</dependency>
 
 		<dependency>
 			<groupId>org.testcontainers</groupId>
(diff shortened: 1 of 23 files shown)


spring-cloud-aws-autoconfigure/pom.xml
```diff
@@ -78,19 +67,12 @@
...
-			<groupId>software.amazon.awssdk</groupId>
-			<artifactId>auth</artifactId>
-		</dependency>
 		<dependency>
 			<groupId>software.amazon.awssdk</groupId>
 			<artifactId>ssm</artifactId>
-		</dependency>
-		<dependency>
-			<groupId>software.amazon.awssdk</groupId>
-			<artifactId>s3</artifactId>
```
MatejNedic, line 91: S3 should not be removed.

MatejNedic (changes requested):

spring-cloud-aws-autoconfigure/pom.xml
maciejwalkowiak: Above we are pulling:

```xml
<dependency>
			<groupId>io.awspring.cloud</groupId>
			<artifactId>spring-cloud-aws-s3</artifactId>
			<optional>true</optional>
		</dependency>
```

So this s3 dependency is redundant.

MatejNedic (approved):

PR #315, opened 2022-04-13
Provide flexible way of resolving S3 objects content type.

### LINKED ISSUE #131: Use Java 1.6+ MimeTypesFiletypeMap for SimpleStorageResource content type derivation

maciejwalkowiak:
Fixes #131


maciejwalkowiak:
Javadocs, polishing and reference docs pending.

PR #390, opened 2022-05-19
Change property name resolved for plaintext secrets in Secrets Manager integration

MatejNedic:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
Since secret names that are used in secrets manager usually follow patter `/something/enviroment/secretname` we should resolve plaintext secret to `secretname`.
In case of `/certs/prod/fn_certificate `   `fn_certificate` should be secret name that will be used with `@Value("${fn_certificate}`.
This allows more fine grained namings for example:
`/certs/dev/fn_certificate`
`/certs/prod/fn_certificate`
and so on.

To allow users to fine grain secrets names depending on evniroment.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

### DIFF
--- a/docs/src/main/asciidoc/secrets-manager.adoc
+++ b/docs/src/main/asciidoc/secrets-manager.adoc
@@ -114,8 +114,6 @@ Secret value can be retrieved by referencing secret name:
 spring.datasource.url=${jdbc-url}
 ----
 
-Secrets Manager integration cannot resolve plaintext secrets that end with `/`, for example if we had `/secrets/my-certificate/` exception would be thrown and application fails to start.
-
 === Using SecretsManagerClient
 
 The starter automatically configures and registers a `SecretsManagerClient` bean in the Spring application context. The `SecretsManagerClient` bean can be used to create or retrieve secrets imperatively.
--- a/docs/src/main/asciidoc/secrets-manager.adoc
+++ b/docs/src/main/asciidoc/secrets-manager.adoc
@@ -114,6 +114,8 @@ Secret value can be retrieved by referencing secret name:
 spring.datasource.url=${jdbc-url}
 ----
 
+Secrets Manager integration cannot resolve plaintext secrets that end with `/`, for example if we had `/secrets/my-certificate/` exception would be thrown and application fails to start.
+
 === Using SecretsManagerClient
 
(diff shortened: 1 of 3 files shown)


spring-cloud-aws-secrets-manager/src/main/java/io/awspring/cloud/secretsmanager/SecretsManagerPropertySource.java
```diff
@@ -89,8 +89,11 @@ private void readSecretValue(GetSecretValueRequest secretValueRequest) {
 		}
 		catch (JsonParseException e) {
 			// If the secret is not a JSON string, then it is a simple "plain text" string
-			LOG.debug("Populating property retrieved from AWS Secrets Manager: " + secretValueResponse.name());
-			properties.put(secretValueResponse.name(), secretValueResponse.secretString());
+			String secretName = secretValueResponse.name().lastIndexOf("/") != 0
```
maciejwalkowiak, line 92: Can secrets end with "/"? If so, i think `/secret/foobar/` would end with empty `secretName`. Maybe we can verify that in tests?

MatejNedic: They can. But now thinking should we even support that, not sure if it makes sense to have a `secretName` ending with `/`:
`/something/prod/certificate/` not sure about ending `/` as valid end of a `secretName`.
Even in AWS samples they always have `/something/secretName`. 
Wdyt?

maciejwalkowiak: It's rather the question if its allowed on the AWS side to end with `/`. If it is either we have to support it or throw meaningful error message

docs/src/main/asciidoc/secrets-manager.adoc
```diff
@@ -114,6 +114,8 @@ Secret value can be retrieved by referencing secret name:
 spring.datasource.url=${jdbc-url}
 ----
 
+Secrets Manager integration cannot resolve plaintext secrets that end with `/`, for example if we had `/secrets/my-certificate/` exception would be thrown and application fails to start.
```
maciejwalkowiak, line 117: Just curious - can't we make it resolve with the name preceeding the last `/`?

Issue #393, opened 2022-05-21
Fail on startup if sqs queue does not exist


boolivar:
**Type**: Feature

**The problem**
If queue specified for `SqsListener` is not exists at the time of application startup, `SimpleMessageListenerContainer` simply ignores that queue and continue without subscribing, it just puts WARN message in the logs. 

**Possible solution**
It would be nice to have a possibility to fail on startup if queue does not exist.

**Alternative**
Another alternative is to wait for queue and start listening when queue appears in sqs.

**Current workaround**
Workaround is to throw exception from `destinationResolver` that will not be catched by `SimpleMessageListenerContainer`:

```java
public class FailingDestinationResolver<T> implements DestinationResolver<T> {
    
    private final DestinationResolver<T> destinationResolver;

    public FailingDestinationResolver(DestinationResolver<T> destinationResolver) {
        this.destinationResolver = destinationResolver;
    }

    @Override
    public T resolveDestination(String name) {
        try {
            return destinationResolver.resolveDestination(name);
        } catch (DestinationResolutionException e) {
            throw new RuntimeException("Fail to resolve '" + name + "' destination", e);
        }
    }
}
```
Configuration `SimpleMessageListenerContainerFactory` bean with destination resolver:
```java
@Bean
public SimpleMessageListenerContainerFactory sqsListenerContainerFactory(DestinationResolver<String> queueDestinationResolver) {
    SimpleMessageListenerContainerFactory factory = new SimpleMessageListenerContainerFactory();
    factory.setDestinationResolver(queueDestinationResolver);
    return factory;
}
    
@Bean
public DestinationResolver<String> queueDestinationResolver(AmazonSQSAsync sqs) {
    return new CachingDestinationResolverProxy<>(new FailingDestinationResolver<>(new DynamicQueueUrlDestinationResolver(sqs, null)));
}
```
**Note** that spring `cloud.aws.sqs.listener.*` properties ignored for this configuration, additional code is required to pick them up from `SqsProperties`.

maciejwalkowiak:
Fixed in https://github.com/awspring/spring-cloud-aws/pull/394

PR #394, opened 2022-05-21
Fail on startup if sqs queue does not exist

boolivar:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
This PR adds `failOnMissingQueue` property for `SimpleMessageListenerContainer` to fail on application startup if sqs queue is not found.

This PR is my proposal for (my) issue #393 

Unit tests and local application run that imports local version `2.4.2-SNAPSHOT` of the library:

```java
@SpringBootApplication
public class App {

    public static void main(String[] args) {
        SpringApplication.run(App.class, args);
    }
    
    @SqsListener("nonexistentqueue")
    public void queueListener(String message) {
    }
}
```
```
cloud.aws.sqs.listener.fail-on-missing-queue = true
```

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
Yes, I'm aware of `spring-cloud-aws v3`, but `spring-integration-aws` uses `SimpleMessageListenerContainer` so this feature should be exposed there as well.

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -22,7 +22,7 @@
 	<parent>
 		<groupId>io.awspring.cloud</groupId>
 		<artifactId>spring-cloud-aws</artifactId>
-		<version>2.4.1</version>
+		<version>2.4.2-SNAPSHOT</version>
 	</parent>
 	<artifactId>spring-cloud-aws-docs</artifactId>
 	<packaging>pom</packaging>
--- a/docs/src/main/asciidoc/_configprops.adoc
+++ b/docs/src/main/asciidoc/_configprops.adoc
@@ -50,12 +50,12 @@
 |cloud.aws.sqs.handler.default-deletion-policy |  | Configures global deletion policy used if deletion policy is not explicitly set on {@link SqsListener}.
 |cloud.aws.sqs.listener.auto-startup | `true` | Configures if this container should be automatically started.
 |cloud.aws.sqs.listener.back-off-time |  | The number of milliseconds the polling thread must wait before trying to recover when an error occurs (e.g. connection timeout).
+|cloud.aws.sqs.listener.fail-on-missing-queue | `false` | Configures if this container should fail on initialization if queue not found.
 |cloud.aws.sqs.listener.max-number-of-messages | `10` | The maximum number of messages that should be retrieved during one poll to the Amazon SQS system. This number must be a positive, non-zero number that has a maximum number of 10. Values higher then 10 are currently not supported by the queueing system.
 |cloud.aws.sqs.listener.queue-stop-timeout |  | The queue stop timeout that waits for a queue to stop before interrupting the running thread.
 |cloud.aws.sqs.listener.visibility-timeout |  | The duration (in seconds) that the received messages are hidden from subsequent poll requests after being retrieved from the system.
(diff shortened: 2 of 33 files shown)


maciejwalkowiak:
Thanks for the PR!

FYI `spring-integration-aws` will be eventually migrated to 3.0 as well, but I think we should include this change in 2.x and also make sure that it's gets propagated to 3.0

maciejwalkowiak (approved):
Perhaps we could also add a failure analyzer for that, but even without it, green light from my side.

@MatejNedic @eddumelendez do you want to take a look?

MatejNedic (approved):
Tnx on PR!

This looks good to me!

maciejwalkowiak:
Thanks for docs @MatejNedic but you flipped the default. By default in `2.4.x` it will be false as it has to stay backward compatible.

MatejNedic:
@maciejwalkowiak oh did not notice that tnx! Will update!

jamfor352:
Hi @maciejwalkowiak - can you confirm if this is in Spring Cloud AWS 3.0.0 under a different property name or function?
I don't seem to be able to find this in `io.awspring.cloud.autoconfigure.sqs.SqsProperties`

tomazfernandes:
Hi @jamfor352, there’s no autoconfiguration property for this behavior in 3.0 - not sure it should but we can look into that.

The default behavior is to create any missing queues.

You can change this behavior in the ContainerOptions: https://docs.awspring.io/spring-cloud-aws/docs/3.0.0-M3/reference/html/index.html#containeroptions-descriptions

Let us know if that works for you, thanks.

jamfor352:
> Hi @jamfor352, there’s no autoconfiguration property for this behavior in 3.0 - not sure it should but we can look into that.
> 
> The default behavior is to create any missing queues.
> 
> You can change this behavior in the ContainerOptions: https://docs.awspring.io/spring-cloud-aws/docs/3.0.0-M3/reference/html/index.html#containeroptions-descriptions
> 
> Let us know if that works for you, thanks.

Thanks for the reply, I'll look into that - having this in Autoconfiguration would be extremely useful, any reason it was removed when it was functioning in 2.4? I couldn't find any documentation explicitly stating its removal or any reasons why, which threw me for a loop.

(For context: A service creating infrastructure is quite often considered a no-no)

tomazfernandes:
The 3.0 version is a major rewrite of the SQS integration - there’s not much code in common with 2.x. So it’s not that it was removed, it was simply not added to this new version. We chose to do this rewrite because there were many design issues with SQS 2.x, and also to take full advantage of the non-blocking capabilities of AWS SDK 2.0.

It’s not uncommon for frameworks to have resource creation by default - I know Kafka does for example. This helps setting up development and test environments.

For a production environment, it’s expected that the application won’t have permission to create resources, so it should fail if a queue is not found regardless of the configuration.

Maybe we can add this disclaimer about the rewrite to the docs, or maybe in migration guide - sorry the lack of this information threw you in for loop.

jamfor352:
Thanks for the explanation @tomazfernandes . 
In our use-case it wouldn't make sense for the autocreation as we rely on SNS -> SQS fanout (which is a common pattern) and presumably there's no auto-configurable way to attach SQS queues to SNS topics? (If there is that would be great!)

MatejNedic:
Hey @jamfor352 , lets move this to [discussions](https://github.com/awspring/spring-cloud-aws/discussions)  on github. There might be other people wanting to see this and it is better to keep this kind of communication there. I am afraid it will be lost here.
We can then open issues depending on inputs from community and so on.

Just to note SNS has a similar implementation as well regarding resolving topicName to arn. Lets move discussion and see what conclusions we come up with! :)

PR #397, opened 2022-05-24
Avoid using ReflectionTestUtils

eddumelendez:


spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/s3/S3TransferManagerAutoConfigurationTests.java
```diff
@@ -97,9 +102,14 @@ void withCustomGlobalEndpointAndS3Endpoint() {
 			contextRunner.withPropertyValues("spring.cloud.aws.endpoint:http://localhost:8090",
 					"spring.cloud.aws.s3.endpoint:http://localhost:9999").run(context -> {
 						S3TransferManager transferManager = context.getBean(S3TransferManager.class);
-						ConfiguredAwsClient client = new ConfiguredAwsClient(resolveS3Client(transferManager));
-						assertThat(client.getEndpoint()).isEqualTo(URI.create("http://localhost:9999"));
-						assertThat(client.isEndpointOverridden()).isTrue();
+						assertThat(transferManager)
```
maciejwalkowiak, line 105: I do agree that we should avoid using `ReflectionTestUtils` in tests methods but I would rather follow in other files what's been done here with a method that extracts what's needed or separate class like `ConfiguredAwsClient`. AssertJ `extracting` heavily stringifies tests and doesn't make them much more readable.

maciejwalkowiak:
Fixed in other PRs.

PR #400, opened 2022-05-26
Add InMemoryBufferingS3OutputStream

### LINKED ISSUE #392: In-Memory Buffered S3 Output Stream
**Type**: Feature

The  `S3Resource` in the old spring-cloud-aws used to return an OutputStream that would buffer the content, and send the content (when it exceeded the buffer capacity), to S3 as a multipart upload. This is particularly useful for large streaming files where the user may not want to store the entire file on the filesystem. The current OutputStream(s) all write the content to an intermediate file on disk before sending it to S3.

Implement an `InMemoryBufferedS3OutputStream` which uses a ByteBuffer to store written bytes. Flush the buffer (starting a multipart upload, if not yet started) and send the content of the buffer as each chunk of the multipart upload. Complete the multipart upload when `close` is called.

There isn't really one in the current implementation. If `AbstractTempFileS3OutputStream` used `Path` instead of `File`, there might be a way to do it with a custom filesystem implementation.

sgarfinkel:
- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
Resolves #392 by adding a new InMemoryBufferingS3OutputStream which uses MultiPartUploads to stream large files to S3 without using the Filesystem.

As discussed in the issue, this feature existed in the old version of spring-cloud-aws. It has been brought over in this new implementation, which although inspired by it, is entirely new.

There's a pretty extensive set of tests in the new `InMemoryBufferingS3OutputStreamTests` file. Test coverage is quite good, at about 75% of lines. That 25% represents conditionals that log warnings and thrown exceptions and the like. All the business logic is fully covered.

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
Consider making this the default implementation used by the AutoConfiguration to backport the old functionality that used this same mechanism by default.

### DIFF
--- a/spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/test/java/io/awspring/cloud/s3/InMemoryBufferingS3OutputStreamTests.java
+++ b/spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/test/java/io/awspring/cloud/s3/InMemoryBufferingS3OutputStreamTests.java
@@ -40,7 +40,7 @@ import software.amazon.awssdk.services.s3.model.*;
  * @author Sam Garfinkel
  */
 @ExtendWith(MockitoExtension.class)
-public class InMemoryBufferingS3OutputStreamTests {
+class InMemoryBufferingS3OutputStreamTests {
 
 	@Mock
 	private S3Client s3Client;


sgarfinkel:
The failed security gate is complaining about the MD5 algorithm being used to generate a hash of the content :|

spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/InMemoryBufferingS3OutputStream.java
```diff
@@ -0,0 +1,228 @@
...
+
+	private final S3Client s3Client;
+
+	@Nullable
+	private final ObjectMetadata objectMetadata;
+
+	@Nullable
+	private final S3ObjectContentTypeResolver contentTypeResolver;
+
+	private final int bufferSize;
```
maciejwalkowiak, line 62: Perhaps we could either use `org.springframework.util.unit.DataSize` type or make it more explicit `bufferSizeInBytes`. I guess `DataSize` is a better option. Same applies to all the fields or parameters where we referred to bytes implicitly like `DEFAULT_BUFFER_CAPACITY`.

maciejwalkowiak (changes requested):
Thanks @sgarfinkel for a great contribution! I agree that we should consider it making a default one.

Some things to polish:

`S3ResourceIntegrationTests` should be updated to also use `InMemoryBufferingS3OutputStream` among other output stream providers (as far as i can see one tests for `InMemoryBufferingS3OutputStream` is going to fail).

maciejwalkowiak:
I did some polishing. Docs update still pending. @MatejNedic take a look please.

sgarfinkel:
> I did some polishing. Docs update still pending. @MatejNedic take a look please.

Thank you! I’ve had some personal things come up and have had very little time to do OSS work the past week.

maciejwalkowiak:
@sgarfinkel no worries, you've done great job already!

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/s3/S3AutoConfiguration.java
```diff
@@ -130,9 +131,9 @@ S3ObjectConverter s3ObjectConverter(Optional<ObjectMapper> objectMapper) {
 
 	@Bean
 	@ConditionalOnMissingBean
-	S3OutputStreamProvider diskBufferingS3StreamProvider(S3Client s3Client,
+	S3OutputStreamProvider inMemoryBufferingS3StreamProvider(S3Client s3Client,
 			Optional<S3ObjectContentTypeResolver> contentTypeResolver) {
-		return new DiskBufferingS3OutputStreamProvider(s3Client,
+		return new InMemoryBufferingS3OutputStreamProvider(s3Client,
```
MatejNedic, line 136: We should provide a way that `bufferSize` can be set in properties.

MatejNedic (commented):
Hey @sgarfinkel ,
Tnx on contribution. Looks good to me, only 1 thing I am thinking if we should introduce as well.

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/s3/S3AutoConfiguration.java
maciejwalkowiak: it can be easily defined by overwriting a bean. i am not sure how to structure it at this stage, we can add a property later when feedback like this comes.

PR #410, opened 2022-06-08
Polish SES integration.

maciejwalkowiak:


spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/ses/SesAutoConfiguration.java
```diff
@@ -67,7 +66,7 @@ public MailSender simpleMailSender(SesClient sesClient) {
 	}
 
 	@Bean
-	@ConditionalOnClass(Session.class)
+	@ConditionalOnClass(name = "javax.mail.Session")
```
maciejwalkowiak, line 69: `ConditionalOnClass` on the method level needs a string rather than actual `Class`.

Issue #447, opened 2022-07-06
S3 User defined metadata key is case-sensitive


sis-yoshiday:
**Type**: Bug

**Component**:
"S3"

**Describe the bug**
User defined metadata key is case-sensitive. 

Below document say user-defined metadata key is stored in lower-case.

> https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#object-metadata
User-defined metadata is a set of key-value pairs. Amazon S3 stores user-defined metadata keys in lowercase.

**Sample**

```java
s3Template.upload(
        "bucket",
        "key",
        new ByteArrayInputStream(new byte[] { 0 }),
        ObjectMetadata.builder()
            .metadata("myMetadata", "foo")
            .build());

// ...

S3Client s3Client;

Map<String, String> metadata =
        s3Client.getObject(GetObjectRequest.builder().bucket("bucket").key("key").build())
            .response()
            .metadata();

metadata.get("myMetadata");// null
metadata.get("mymetadata");// "foo"
```

Can we have metadata as some case-insensitive Map ?

```java
metadata = new TreeMap<String, String>(String.CASE_INSENSITIVE_ORDER);
```

## aws-java-sdk v1

https://github.com/aws/aws-sdk-java/blob/56d8e9fa02a3bd7562e4588f4fbfd2dda545bd12/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/ObjectMetadata.java#L53

```java
/**
  * Custom user metadata, represented in responses with the x-amz-meta-
  * header prefix
  */
private Map<String, String> userMetadata = 
  new TreeMap<String, String>(String.CASE_INSENSITIVE_ORDER);
```

In sdk v1, user-defined-metadata key seems to be case-insensitive.

## aws-java-sdk v2 (2.17.225)

response metadata seems to be copied by `software.amazon.awssdk.services.s3.model.MetadataCopier#copy` and finally returns as immutable map and not case-insensitive.

maciejwalkowiak:
Regarding first piece of code:

```java
s3Template.upload(
        "bucket",
        "key",
        new ByteArrayInputStream(new byte[] { 0 }),
        ObjectMetadata.builder()
            .metadata("myMetadata", "foo")
            .metadata("mymetadata", "boo")
            .build());
```

we could change `ObjectMetadata` class to store in such case only `mymetadata` property with `boo` value.

[...]

PR #451, opened 2022-07-09
Add Ability to get metadata via S3Resource

sis-yoshiday:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Add Ability to get metadata via S3Resource

We can retrieve metadata from `S3Resource` without using `S3Client` directly.

Add below integration test.

`io.awspring.cloud.s3.S3TemplateIntegrationTests#retrieveMetadata`

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#447 Should or Should not treat metadata key as case-insensitive.

### DIFF
--- a/spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/S3Resource.java
+++ b/spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/S3Resource.java
@@ -40,7 +40,7 @@ import software.amazon.awssdk.services.s3.model.NoSuchKeyException;
  * @author Agim Emruli
  * @author Alain Sahli
  * @author Maciej Walkowiak
- * @author yoshida
+ * @author Yuki Yoshida
  * @since 3.0
  */
 public class S3Resource extends AbstractResource implements WritableResource {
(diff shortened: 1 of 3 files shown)


MatejNedic (approved):

maciejwalkowiak:
Could it return `ObjectMetadata` instead of map?

sis-yoshiday:
@maciejwalkowiak 
Like this ?
`HeadMetadata` class should be removed ?

```java
public String contentType() {
  if (objectMetadata == null) {
    fetchMetadata();
  }
  return objectMetadata.getContentType();
}

@Override
public long contentLength() throws IOException {
  if (objectMetadata == null) {
    fetchMetadata();
  }
  return objectMetadata.contentLength;
}

@Override
public long lastModified() throws IOException {
  if (objectMetadata == null) {
    fetchMetadata();
  }
  return objectMetadata.lastModified.toEpochMilli();
}

public ObjectMetadata objectMetadata() {
  if (objectMetadata == null) {
    fetchMetadata();
  }
  return objectMetadata;
}

private void fetchMetadata() {
  HeadObjectResponse response = s3Client.headObject(request -> 
    request.bucket(location.getBucket())
                 .key(location.getObject())
                 .versionId(location.getVersion()));

  this.objectMetadata = ObjectMetadata.of(response);// add static factory to ObjectMetadata
}
```

There is below problems.

- `objectMetadata.contentLength` does not exists.
- `objectMetadata.lastModified` does not exists.

maciejwalkowiak:
@sis-yoshiday I was wrong. `ObjectMetadata` does not fit here, it's better to stick with a map.

PR #457, opened 2022-07-15
SNS SMS support

MatejNedic:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
This is currently just a draft how I vision Spring Cloud AWS support for sending SMS messages.
Before I go further with tests @maciejwalkowiak please check design out.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/sms/attributes/ADM.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.awspring.cloud.sns.sms.attributes;
+
+public class ADM {
```
maciejwalkowiak, line 18: Have you considered making these attribute classes immutable?

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/sms/attributes/ADM.java
```diff
@@ -0,0 +1,50 @@
...
+		return new Builder();
+	}
+
+	public static final class Builder {
+		private Long ttl;
+
+		private Builder() {
+		}
+
+		public Builder withTtl(Long ttl) {
```
maciejwalkowiak, line 39: For the sake of consistency with other builder classes in the project and builder classes in AWS SDK, instead of "withers" method names should be just named with an attribute"

```
Builder ttl(Long ttl) {
...
}
```

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/sms/core/DefaultMessageAttributeConverter.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.Map;
+import software.amazon.awssdk.services.sns.model.MessageAttributeValue;
+
+/**
+ * @author Matej Nedic
+ * @since 3.0.0
+ */
+public class DefaultMessageAttributeConverter implements MessageAttributeConverter {
+
+	public Map<String, MessageAttributeValue> convert(SmsMessageAttributes attr) {
```
maciejwalkowiak, line 32: Have you considered moving this conversion to `SmsMessageAttributes` class? Does not seem like there can be any other reasonable implementation of `MessageAttributeConverter` interface

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/sms/core/SnsSmsOperations.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package io.awspring.cloud.sns.sms.core;
+
+import io.awspring.cloud.sns.sms.attributes.SmsMessageAttributes;
+
+/**
+ * @author Matej Nedic
+ * @since 3.0.0
+ */
+public interface SnsSmsOperations {
```
maciejwalkowiak, line 24: Perhaps we can think about changing this interface somehow to contains just two methods and the method parameters would make a distinction if it goes to phone number, topic or target, but I don't have now solution in mind.

maciejwalkowiak (changes requested):

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/sms/core/SnsSmsOperations.java
MatejNedic: We could make a special class that would have TopicArn, TargetArn or a Phone Number and just accept that.

MatejNedic:
Will commit sample in following days

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/sms/attributes/AttributeCodes.java
```diff
@@ -16,6 +16,7 @@
 package io.awspring.cloud.sns.sms.attributes;
 
 /**
+ * Helper class that is used to reference field of {{@link SmsMessageAttributes}} to key that will be used in @{@link software.amazon.awssdk.services.sns.model.PublishRequest}.
```
maciejwalkowiak, line 19:
```suggestion
 * Helper class that is used to reference field of {@link SmsMessageAttributes} to key that will be used in {@link software.amazon.awssdk.services.sns.model.PublishRequest}.
 
```

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/sms/attributes/SmsMessageAttributes.java
```diff
@@ -0,0 +1,338 @@
...
+ */
+public class SmsMessageAttributes {
+	private String senderID;
+	private String originationNumber;
+	private String maxPrice;
+	private SmsType smsType;
+	private String entityId;
+	private String templateId;
+	private String messageGroupId;
+	private String deduplicationId;
```
maciejwalkowiak, line 37: Deduplication id makes sense only for sending to topics, so it should not be a property on the top level attributes used also for sending to a number. https://docs.aws.amazon.com/sns/latest/dg/fifo-message-dedup.html

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/sms/attributes/SmsMessageAttributes.java
```diff
@@ -0,0 +1,338 @@
...
+	private String messageStructure;
```
maciejwalkowiak, line 38: As far as I can see, the only valid value for message structure is "json" and it is used for sending to topic when more than single protocol endpoints are subscribed to this topic. I think this is "advanced" usage that we don't need to cover with `SnsSmsTemaplte` - we can support sending only to topics with sms endpoints or if we decide to support mixed topics, this has to be redesigned.

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/sms/core/SnsSmsOperations.java
```diff
@@ -0,0 +1,36 @@
...
+	void send(String phoneNumber, String message);
+
+	void send(String phoneNumber, String message, SmsMessageAttributes attributes);
+
+	void sendToTopicArn(String topicArn, String message);
+
+	void sendToTopicArn(String topicArn, String message, SmsMessageAttributes attributes);
+
+	void sendToTargetArn(String targetArn, String message);
```
maciejwalkowiak, line 33: What is actually the use case for sending to `targetArn`? As far as I could find out there is no use case for SMS to send to target arn. Only to topic or directly to a phone number.

spring-cloud-aws-sns/src/test/java/io/awspring/cloud/sns/sms/core/SnsSmsTemplateIntegrationTest.java
```diff
@@ -0,0 +1,67 @@
...
+				.region(Region.of(localstack.getRegion()))
+				.credentialsProvider(StaticCredentialsProvider.create(AwsBasicCredentials.create("noop", "noop")))
+				.build();
+		snsSmsTemplate = new SnsSmsTemplate(snsClient);
+	}
+
+	@Test
+	void sendValidMessage_ToPhoneNumber() {
+		Assertions
+				.assertDoesNotThrow(() -> snsSmsTemplate.send("+385 00 000 0000", "Spring Cloud AWS got you covered!"));
```
maciejwalkowiak, line 58: Can we assert log message from Localstack?

https://github.com/localstack/localstack/pull/2822

maciejwalkowiak (changes requested):
Please add `@Nullable` to attribute classes (I imagine all these parameters are nullable..?).

Also perhaps we can merge packages `attributes` and `core` into single `sms` - this way we could reduce the number of public classes

spring-cloud-aws-sns/src/test/java/io/awspring/cloud/sns/sms/core/SnsSmsTemplateIntegrationTest.java
MatejNedic: At first glance, I am not seeing it in container logs but will check again later.

MatejNedic:
Tried it against real AWS integration works as expected.

maciejwalkowiak:
Great job! Please add nullability annotations (including `package-info.java`) and javadocs on public classes & interfaces.

spring-cloud-aws-sns/src/test/java/io/awspring/cloud/sns/sms/core/SnsSmsTemplateIntegrationTest.java
tomazfernandes: If I understand correctly, we're only checking the message is sent and doesn't throw any errors. Maybe we can fetch the message from local stack after sending it, and assert that?

If there are any concurrency concerns, I usually use a `CountDownLatch` to wait on the test thread until the message is consumed.

MatejNedic: Hey @tomazfernandes I tried with getting localstack logs but I am not seeing a message. We can on other hand make operations and template return messageId and check that. Localstack uses moto to mock sms sending.

tomazfernandes: Hey @MatejNedic, nice work here! Sure, if the SNS API returns an id I guess we should return that to users.

This being an SNS integration, can't we maybe subscribe to the topic and receive the sent message ourselves too? Just guessing, have no clue how this actually works 😄

eddumelendez: @MatejNedic Did you set the DEBUG env var to "1"? in LocalStackContainer should be `.withEnv("DEBUG", "1")` and change the logger for testcontainers to DEBUG too `<logger name="org.testcontainers" level="DEBUG"/>` just to make sure something useful is in the container's log. Then you can fetch the logs doing `String logs = localstack.getLogs(OutputFrame.OutputType.STDOUT, OutputFrame.OutputType.STDERR);` and assert against it.

MatejNedic: @eddumelendez I tried that as well but I am not seeing the correct message I want to see.
`Delivering SMS message to`. I am seeing `Publishing message to TopicArn: None | Message: Spring Cloud AWS got you covered` which makes sense but is still not what I want. I have added tests with at least some kind of message. I tried this against real AWS and I am getting SMS on my phone number without an issue.

@tomazfernandes we have something similar for SNS notifications where we send it to an SQS and we poll it from SQS. Sadly with SMS we can't do that.

PR #480, opened 2022-08-23
Simplify generator path calculation

fjmacagno:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description
A piece of https://github.com/fjmacagno/spring-cloud-aws/pull/1 which i thought might still be useful. Instead of having the generator take the path that the template is at, include the template as a resource so that it will always be accessible, whether the generator is run with access to the source tree or not. It also swaps the argument to be the output directory, so that the generator doesn't need to be project-aware.

Originally this made putting the generated class into the target dir possible, but now motivation is just that i already did the work and the code is potentially cleaner for it.

Running `mvn clean` and `make build`.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/spring-cloud-aws-s3-parent/spring-cloud-aws-s3-codegen/pom.xml
+++ b/spring-cloud-aws-s3-parent/spring-cloud-aws-s3-codegen/pom.xml
@@ -29,6 +29,11 @@
 	</dependencies>
 
 	<build>
+		<resources>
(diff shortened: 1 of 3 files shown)


maciejwalkowiak:
Very nice @fjmacagno! thanks!

Issue #427, opened 2022-06-16
Exclude `spring-cloud-aws-starters` from publishing


maciejwalkowiak:
`spring-cloud-aws-starters` module gets published to Maven central even though it serves just as a container for other starters.

Simply excluding it from the release plugin means that its submodules will also be excluded, so something else has to be done.

PR #530, opened 2022-10-03
Adds aws starter for RegionProviderAutoConfiguration

rajadilipkolli:
If a project is using only metrics starter then Creation of CloudWatchAsyncClient Bean couldn't be completed as it needs AwsRegionProvider Bean.

Though there is a `RegionProviderAutoConfiguration` it can't be triggered as there is a condition on 

`@ConditionalOnClass({ StaticRegionProvider.class, AwsRegionProvider.class, ProfileFile.class })`

Here **`StaticRegionProvider.class`** is present in aws-core Hence adding  it to fix the issue

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/spring-cloud-aws-starters/spring-cloud-aws-starter-metrics/pom.xml
+++ b/spring-cloud-aws-starters/spring-cloud-aws-starter-metrics/pom.xml
@@ -36,6 +36,10 @@
 			<artifactId>spring-cloud-aws-autoconfigure</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>io.awspring.cloud</groupId>
+			<artifactId>spring-cloud-aws-core</artifactId>
+		</dependency>
+		<dependency>
 			<groupId>io.micrometer</groupId>
 			<artifactId>micrometer-registry-cloudwatch2</artifactId>
 		</dependency>
--- a/spring-cloud-aws-starters/spring-cloud-aws-starter-metrics/pom.xml
+++ b/spring-cloud-aws-starters/spring-cloud-aws-starter-metrics/pom.xml
@@ -33,11 +33,7 @@
 	<dependencies>
 		<dependency>
 			<groupId>io.awspring.cloud</groupId>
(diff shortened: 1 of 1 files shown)


spring-cloud-aws-starters/spring-cloud-aws-starter-metrics/pom.xml
```diff
@@ -35,6 +35,10 @@
 			<groupId>io.awspring.cloud</groupId>
 			<artifactId>spring-cloud-aws-autoconfigure</artifactId>
 		</dependency>
+		<dependency>
```
maciejwalkowiak, line 38: Similar to other starters, `metrics` module should depend on:

```xml
<dependency>
	<groupId>io.awspring.cloud</groupId>
	<artifactId>spring-cloud-aws-starter</artifactId>
</dependency>
```

and the dependency to `spring-cloud-aws-autoconfigure` should be removed.

maciejwalkowiak (changes requested):
Good catch @rajadilipkolli!

I suggested different solution to this problem that follows convention that's used in other starter modules.

spring-cloud-aws-starters/spring-cloud-aws-starter-metrics/pom.xml
rajadilipkolli: @maciejwalkowiak , Should I raise a PR adding it but I guess adding to the starter will bring #427 back

maciejwalkowiak: Just update this PR. #427 was only about `spring-cloud-aws-starters` module (not "starter") which used to be an aggregator module for all the starters, so there's no risk of bringing it back :)

rajadilipkolli: Updated

PR #535, opened 2022-10-14
Upgrade Localstack to 1.2.0 

rajadilipkolli:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
Increments the local stack version for better integration

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


maciejwalkowiak:
Thanks.It should be also done for all the integration tests. Would you like to update it too?

rajadilipkolli:
> Thanks.It should be also done for all the integration tests. Would you like to update it too?

Sure I will

rajadilipkolli:
@maciejwalkowiak , Shouldn't we change it to` localstack:localstack` so that always the latest one is pulled instead of changing the version for every increment?

maciejwalkowiak:
@rajadilipkolli not really it's better to rely on exact versions.

maciejwalkowiak:
Tests are failing now though so likely there is something to fix.

PR #543, opened 2022-10-18
Fix injecting `SnsTemplate` and `SnsSmsTemplate` beans.

maciejwalkowiak:


maciejwalkowiak:
Before merging this fix lets wait what comes up from https://github.com/spring-projects/spring-boot/issues/32763

Issue #576, opened 2022-12-11
Start and stop the sqs listener based on user inputs


bharatTR:
Question

I am working on spring boot application, it uses docker and k8s for deployment. I need to expose endpoints that would start and stop sqs listener based on user inputs, but I am not sure how to perform this job. Is there a way to disable/enable sqs listener based on user inputs?

tomazfernandes:
Hi @bharatTR,

For `@SqsListener` containers you can inject the `MessageListenerContainerRegistry` and call `#start` and `#stop` on it to manage all containers, or on any of the containers themselves.

Please let us know it that works for you or if you have any questions.

Thanks.

Issue #573, opened 2022-12-09
Secrets and Parameters clobber each other when the resource names are the same.


thorntonrp:
**Type**: Bug

**Component**:  "Parameter Store",  "Secrets Manager"

**Describe the bug**

In the code below, I'm importing both secrets from Secrets Manager and parameters from Parameter Store. The name of my secret is "/myapp" and the base name of my parameters are also named "/myapp". When these names are the same, my application breaks because the secret property source ends up being given the same name as the parameter store property resource. Both end of getting called "/myapp" and only one of the property resources will get added to the Spring environment.

```yaml
spring:
  config:
    import:
      - "aws-secretsmanager:/myapp"
      - "aws-parameterstore:/myapp"
```

It is not an error for a parameter base name to have the same name as a secret. My recommendation is that both the secret property resource and the parameter property resource be given the full names of their respective imports to ensure the resource names are unique in the Spring environment. For example, the secret property resource should be named `"aws-secretsmanager:/myapp"` and the parameter store property resource should be named `"aws-parameterstore:/myapp"`. This fix shouldn't affect backward compatibility since the resource names are just an implementation detail.

The benefit of this fix would be twofold. First, the one won't end of clobbering the other in the Spring environment. Second, when viewing the properties on the actuator `/env` endpoint, it will be immediately clear where the properties came from. Currently, it is not clear that the properties came from either secrets manager or parameter store.

maciejwalkowiak:
Just to make sure - this is happening in 3.0, right? Would you like to contribute a fix?

thorntonrp:
Yes, I've verified that this is happening on both 3.0 and 2.x. I'll submit a PR to fix.

PR #589, opened 2022-12-16
#576 Ensure that property source names are unique

thorntonrp:
- [ x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
- Added "aws-parameterstore:" prefix to the parameter store property resource name so that it will not conflict with a secrets manager context having the same name.
- Added "aws-secretsmanager:" prefix to the secrets manager property resource name so that it will not conflict with the parameter store context of the same name.

Fixes [573](https://github.com/awspring/spring-cloud-aws/issues/573)

Each property sources should have unique names since Spring Boot uses the names as keys to an internal map. When a parameter store context is named "/myapp" and a secrets manager context is also "/myapp", one of the property sources will end up clobbering the other. By prefixing the parameter store property source name with "aws-parameterstore:" and prefixing the secrets manager property source name with "aws-secretsmanager:", the property source names become sufficiently unique to avoid unintentional clobbering.

This fix also benefits debugging and diagnostics because it allows engineers introspecting the Spring environment or viewing the `/actuator/env` endpoint to immediately identify where the property sources were loaded from. Without the prefix, it would be impossible to infer where the properties came from.

- I added test assertions to the unit tests to validate the property source names and verified that all tests passed.
- I ran a "real" Spring Boot project against my local build artifacts and checked the /actuator/env endpoint to ensure that both the secrets manager and parameter store property source names included the new prefixes.

## :pencil: Checklist
- [ x] I added tests to verify changes
- [ x] All tests passing
- [ x] No breaking changes

### DIFF
--- a/spring-cloud-aws-parameter-store/src/main/java/io/awspring/cloud/parameterstore/ParameterStorePropertySource.java
+++ b/spring-cloud-aws-parameter-store/src/main/java/io/awspring/cloud/parameterstore/ParameterStorePropertySource.java
@@ -47,7 +47,7 @@ public class ParameterStorePropertySource extends AwsPropertySource<ParameterSto
 	private final Map<String, Object> properties = new LinkedHashMap<>();
 
 	public ParameterStorePropertySource(String context, SsmClient ssmClient) {
-		super(context, ssmClient);
+		super("aws-parameterstore:" + context, ssmClient);
 		this.context = context;
 	}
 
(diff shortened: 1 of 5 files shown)


maciejwalkowiak (approved):
Looking good! Thanks a lot @thorntonrp for contribution!

PR #597, opened 2022-12-27
Add SqsTemplate

tomazfernandes:
Hey!

I think implementation-wise this is almost done, pending documentation, javadocs and unit tests.

I'd appreciate feedback, specially on the API / interfaces, if possible.

You can check [SqsTemplateIntegrationTests](https://github.com/tomazfernandes/spring-cloud-aws/blob/GH-507/spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration/SqsTemplateIntegrationTests.java) for some examples.

### Instantiation

Basically, there are two ways of instantiating the `SqsTemplate`. The first one is a simple one with default settings:

```java
SqsOperations<Object> template = SqsTemplate.newTemplate(this.asyncClient);
```

and a more elaborate one with a builder:

```java
SqsTemplate<SampleRecord> template = SqsTemplate.<SampleRecord>builder()
			.sqsAsyncClient(this.asyncClient)
			.configure(options -> options
                                .acknowledgementMode(TemplateAcknowledgementMode.DO_NOT_ACKNOWLEDGE)
				.endpointName(SENDS_AND_RECEIVES_MANUAL_ACK_QUEUE_NAME))
			.build();
```

I chose to go with the builder because `Templates` are usually shared between many threads, so immutability will ensure it's thread safe.

### Sending and Receiving

For sending and receiving messages there are `blocking` and `async` interfaces. The most familiar ones would be [MessagingOperations](https://github.com/tomazfernandes/spring-cloud-aws/blob/GH-507/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/MessagingOperations.java) and [AsyncMessagingOperations](https://github.com/tomazfernandes/spring-cloud-aws/blob/GH-507/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/AsyncMessagingOperations.java).

But the ones I think will be most useful have a more fluent API: [SqsOperations](https://github.com/tomazfernandes/spring-cloud-aws/blob/GH-507/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsOperations.java) and [SqsAsyncOperations](https://github.com/tomazfernandes/spring-cloud-aws/blob/GH-507/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsAsyncOperations.java)

```java
template.send(to -> to
	  .queue(SENDS_AND_RECEIVES_WITH_HEADERS_QUEUE_NAME)
	  .payload(testRecord)
	  .header(myCustomHeader, myCustomValue)
	  .headers(Map.of(myCustomHeader2, myCustomValue2))
);


Optional<Message<SampleRecord>> receivedMessage2 = template.receive(from -> from
	.queue(SENDS_AND_RECEIVES_RECORD_QUEUE_NAME)
        .pollTimeout(Duration.ofSeconds(1)));
```

### Batch

There are variants for sending and receiving batches of messages:

```java
List<Message<SampleRecord>> messagesToSend = IntStream
			.range(0, 5)
			.mapToObj(index -> new SampleRecord("Hello world - " + index, "From SQS!"))
			.map(record -> MessageBuilder.withPayload(record).build())
			.toList();
template.send(SENDS_AND_RECEIVES_BATCH_QUEUE_NAME, messagesToSend);


Collection<Message<SampleRecord>> receivedMessages = template
			.receiveMany(from -> from
				.queue(SENDS_AND_RECEIVES_BATCH_QUEUE_NAME)
				.pollTimeout(Duration.ofSeconds(10))
				.maxNumberOfMessages(10));
```

### FIFO

I've also added specific `FIFO` methods that allow users to specify `messageDeduplicationId` and `messageGroupId` for sending, and `receiveRequestAttemptId` for receiving, adding random UUIDs if none is provided. I'm not sure if maybe it'd be better to simplify this and remove these FIFO methods, leaving all options in the main methods, but seems like this might help users be explicit in this choice.

```java
template.sendFifo(to -> to
	.queue(SENDS_AND_RECEIVES_MESSAGE_FIFO_QUEUE_NAME)
	.messageGroupId(UUID.randomUUID())
	.payload(testBody));

Optional<Message<Object>> receivedMessage = template
	.receiveFifo(from -> from.queue(SENDS_AND_RECEIVES_MESSAGE_FIFO_QUEUE_NAME)
		.receiveRequestAttemptId(UUID.randomUUID()));
```

The `SqsTemplate` API got a bit crowded with all blocking and async methods, so it's probably better to inject `SqsOperations` or `SqsAsyncOperations` instead.

Please let me know if you have any questions or suggestions.

Thanks!

### DIFF
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/AsyncMessagingOperations.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/AsyncMessagingOperations.java
@@ -1,7 +1,19 @@
+/*
+ * Copyright 2013-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package io.awspring.cloud.sqs.operations;
-
-import org.springframework.lang.Nullable;
-import org.springframework.messaging.Message;
 
 import java.time.Duration;
 import java.util.Collection;
@@ -9,10 +21,12 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
+import org.springframework.lang.Nullable;
+import org.springframework.messaging.Message;
 
 /**
- * Asynchronous messaging operations. Implementations should have defaults for {@link Nullable} fields,
- * and can provide chained methods interfaces for a more fluent API.
+ * Asynchronous messaging operations. Implementations should have defaults for {@link Nullable} fields, and can provide
+ * chained methods interfaces for a more fluent API.
  *
  * @author Tomaz Fernandes
  * @since 3.0
@@ -20,8 +34,8 @@ import java.util.concurrent.CompletableFuture;
 public interface AsyncMessagingOperations<T> {
 
 	/**
-	 * Send a {@link Message} to the default endpoint with the provided payload.
-	 * The payload will be serialized if necessary.
+	 * Send a {@link Message} to the default endpoint with the provided payload. The payload will be serialized if
+	 * necessary.
 	 *
 	 * @param payload the payload to send.
 	 * @return a {@link CompletableFuture} to be completed with the message's {@link UUID}.
@@ -29,8 +43,7 @@ public interface AsyncMessagingOperations<T> {
 	CompletableFuture<SendResult<T>> sendAsync(T payload);
 
 	/**
-	 * Send a message to the provided endpoint with the provided payload.
-	 * The payload will be serialized if necessary.
+	 * Send a message to the provided endpoint with the provided payload. The payload will be serialized if necessary.
 	 *
 	 * @param endpointName the endpoint to send the message to.
 	 * @param payload the payload to send.
@@ -39,9 +52,8 @@ public interface AsyncMessagingOperations<T> {
 	CompletableFuture<SendResult<T>> sendAsync(@Nullable String endpointName, T payload);
 
 	/**
-	 * Send the provided message along with its headers to the provided endpoint.
-	 * The payload will be serialized if necessary, and headers will be converted
-	 * to the specific messaging system metadata types.
+	 * Send the provided message along with its headers to the provided endpoint. The payload will be serialized if
+	 * necessary, and headers will be converted to the specific messaging system metadata types.
 	 *
 	 * @param endpointName the endpoint to send the message to.
 	 * @param message the message to be sent.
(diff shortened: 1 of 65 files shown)


maciejwalkowiak:
After quick look it looks great! Go on with docs & javadocs as it always makes it easier to follow the code.

tomazfernandes:
Almost done with the tests, will definitely wrap them and the docs up tomorrow.

maciejwalkowiak:
Perfect! Thanks a lot @tomazfernandes

maciejwalkowiak:
Please do not add `@since` tags in Javadocs - they should be there on production classes but they don't bring any value in test classes.

maciejwalkowiak:
I meant test classes Javadocs in case it wasn't clear 😅. Don't focus too much on that I'll polish javadocs before we make final release anyways

tomazfernandes:
> I meant test classes Javadocs in case it wasn't clear 😅. Don't focus too much on that I'll polish javadocs before we make final release anyways

It was clear! I saw you removed from the other tests, I removed from the new ones in this PR.

Should all tests pass, this is ready for review! 🥳

maciejwalkowiak:
I got this error when running build locally:
```
[ERROR]   SqsFifoIntegrationTests.stopsProcessingAfterAckException:275 
Actual and expected have the same elements but not in the same order, at index 0 actual element was:
  "4"
whereas expected element was:
  "3"
```

When run once again it succeed. I made an attempt to add `@RepeatedTest(20)` on the top of this test method but seems that tests are not prepared for running it multiple times. Perhaps it can be improved.

tomazfernandes:
> I got this error when running build locally:
> 
> ```
> [ERROR]   SqsFifoIntegrationTests.stopsProcessingAfterAckException:275 
> Actual and expected have the same elements but not in the same order, at index 0 actual element was:
>   "4"
> whereas expected element was:
>   "3"
> ```
> 
> When run once again it succeed. 

This means that the `4` message was consumed before `3`. As discussed in [this issue](https://github.com/localstack/localstack/issues/6766), the `LocalStack` implementation of `FIFO` queues don't respect message visibility within the same messageGroupId, so it's not hard to imagine some racing conditions happening. I've pinged them again.

Nevertheless, I've ran this test now a few hundred times against localstack and AWS and it passed all of them.

Let me know if you have any suggestions.

> I made an attempt to add `@RepeatedTest(20)` on the top of this test method but seems that tests are not prepared for running it multiple times. Perhaps it can be improved.

Good call. The tests rely on state that is not necessarily reset between them. I've fixed this particular test and will keep in mind to fix the others.

tomazfernandes:
I've increased the `messageVisibility` for messages in this test, maybe that'll help.

tomazfernandes:
...and now in the right branch 😄

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
```diff
@@ -0,0 +1,866 @@
...
+import software.amazon.awssdk.services.sqs.model.QueueAttributeName;
+import software.amazon.awssdk.services.sqs.model.ReceiveMessageRequest;
+import software.amazon.awssdk.services.sqs.model.ReceiveMessageResponse;
+import software.amazon.awssdk.services.sqs.model.SendMessageBatchRequest;
+import software.amazon.awssdk.services.sqs.model.SendMessageBatchRequestEntry;
+import software.amazon.awssdk.services.sqs.model.SendMessageBatchResponse;
+import software.amazon.awssdk.services.sqs.model.SendMessageBatchResultEntry;
+import software.amazon.awssdk.services.sqs.model.SendMessageRequest;
+
+public class SqsTemplate<T> extends AbstractMessagingTemplate<T, Message>
```
maciejwalkowiak, line 67: I am not convinced with the generic in this class (and same for `SqsOperations` interface). It feels a bit awkward to use autoconfigured `SqsTemplate<Object>`. 

```java
private final SqsTemplate<Object> sqs;
...
SendResult<Object> hello = sqs.send("my-queue", "hello");
Message<Object> result = hello.message(); // returns Message<Object> even though we have passed string
```

Could we change in a way that class does not have generics but its methods still do? So that it works in following way:

```java
private final SqsTemplate sqs;
...
SendResult<String> result = sqs.send("my-queue", "hello");
Message<String> message = hello.message();

SendResult<Person> result2 = sqs.send("my-queue", new Person("john"));
Message<Person> message2 = hello.message();
```

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/AsyncMessagingOperations.java
```diff
@@ -0,0 +1,135 @@
...
+	CompletableFuture<SendResult<T>> sendAsync(T payload);
+
+	/**
+	 * Send a message to the provided endpoint with the provided payload. The payload will be serialized if necessary.
+	 *
+	 * @param endpointName the endpoint to send the message to.
+	 * @param payload the payload to send.
+	 * @return a {@link CompletableFuture} to be completed with the message's {@link UUID}.
+	 */
+	CompletableFuture<SendResult<T>> sendAsync(@Nullable String endpointName, T payload);
```
maciejwalkowiak, line 52: `endpointName` - this essentially means queue name right? Did you give it such a generic name to support queue urls/ARNs or ..?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
tomazfernandes: Great call again, thanks!

> I am not convinced with the generic in this class (and same for `SqsOperations` interface). It feels a bit awkward to use autoconfigured `SqsTemplate<Object>`.

Yeah, I had some doubts about generics throughout the project and removed a lot of what `Spring Kafka` has. Specially since we can't lookup the generics at runtime to e.g. figure out to which class to convert to.

But then I figured out that this generics is important for type safety at compile time. Let's say you have a queue that requires special configuration, and hence you have a different `Template` with maybe even a different `SqsAsyncClient`. You'd probably want to ensure at compile time that the correct `Template` is being used, and for that you can type this generics accordingly. Makes sense to you?

Good thing is I think we can do both the generics and your suggestion!

> ```java
> private final SqsTemplate sqs;
> ...
> SendResult<String> result = sqs.send("my-queue", "hello");
> Message<String> message = hello.message();
> 
> SendResult<Person> result2 = sqs.send("my-queue", new Person("john"));
> Message<Person> message2 = hello.message();
> ```

I'm thinking maybe we can do something similar as I did for the `receive` methods, and type it as `<P extends T>`. This way for the `<Object>` template we can use any class as you suggested, and for specifically-typed templates we can use any subclass.

How does that sound?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/AsyncMessagingOperations.java
tomazfernandes: Yeah, well, I still have an idea that this is a meta-framework and hence I'd rather not tie up the generic messaging interfaces specifically to queues.

But I believe the `send(to -> to.queue("myQueue", myPayload))` and `receive(from -> from.queue("myQueue"))` should be the most used methods, and there the implementation is specific to sqs and queues.

What do you think?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
tomazfernandes: Just to be clear, I meant the `payloadClass` parameter in the `SqsReceiveOptions` interface.

maciejwalkowiak: > Let's say you have a queue that requires special configuration, and hence you have a different Template with maybe even a different SqsAsyncClient. You'd probably want to ensure at compile time that the correct Template is being used, and for that you can type this generics accordingly. Makes sense to you?

This would mean that a type determines a queue which may not be true. Same type can be sent to multiple queues. Using different `SqsTemplate` beans can be achieved with qualifiers - this does not give the compile time guarantees but let's face it many things in Spring don't give you compile time guarantees. 

> private final SqsTemplate sqs;

This gives a warning about raw use of parameterized type. I believe most people will just use `SqsTemplate<Object>`  and in such case I think it causes more headache than problems it solves. I am opting in for just `SqsTemplate` with parameterized methods without parameterized type.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsOperations.java
```diff
@@ -0,0 +1,90 @@
...
+	 * @return The {@link UUID} of the message.
+	 */
+	SendResult<T> send(Consumer<SqsSendOptions.Standard<T>> to);
+
+	/**
+	 * Send a message to a Fifo SQS queue using the {@link SqsSendOptions.Fifo} options.
+	 * @param to a {@link SqsSendOptions.Fifo} consumer.
+	 * @return The {@link UUID} of the message.
+	 */
+	SendResult<T> sendFifo(Consumer<SqsSendOptions.Fifo<T>> to);
```
maciejwalkowiak, line 52: Could we merge it into single method? By having this separating to `sendFifo` and `send` it gets unclear how other  methods can be used. For example: `SendResult<T> send(T payload);` - is it only for standard queue, fifo or both? It's quite impossible to tell without getting into details.

maciejwalkowiak:
If some code is reused from `2.4.x` it would be nice to include its authors in `@author` Javadocs.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/AsyncMessagingOperations.java
maciejwalkowiak: I think we need to balance it out. If we use `queue` in `to.queue()` it is better to change `endpointName` variable names to `queue`. Right now it is not straightforward what the endpoint name is. Without going into the source code as a user I would assume I must set there queue url.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsOperations.java
tomazfernandes: Yeah, I was in doubt about this. The main point of having them separate would be users explicitly selecting the FIFO methods and allowing them to add FIFO-specific properties, while cleaning up the regular `send` API. Also, we add the `messageGroupId` and `messageDeduplicationId` on the user's behalf if it's not specified.

But perhaps it's doing more harm than good.

I'll merge the FIFO-specific properties to the send method, and perhaps we can detect it's a `fifo` queue and add the missing properties if none is specified.

tomazfernandes:
Thanks for the suggestions @maciejwalkowiak! It'll take maybe a couple of days to update the code and docs.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/AsyncMessagingOperations.java
maciejwalkowiak: I would stick with `queue` as the word queue does not determine if it is a queue name, ARN, or URL so it leaves us high level of flexibility keeping it easy to understand.

maciejwalkowiak:
Awesome! I am looking forward.

tomazfernandes:
> If some code is reused from `2.4.x` it would be nice to include its authors in `@author` Javadocs.

Please see [this commit](https://github.com/awspring/spring-cloud-aws/pull/597/commits/8c5844b80e95beeeb1b011d138124faf0289688d)

I've reviewed the files in this PR and I think only these classes had logic from `2.x`, the template classes themselves don't.

If you see any other familiar logic please let me know and I'll add the respective authors.

tomazfernandes:
Almost there! I think the receive APIs are a bit off with a nullable payload type and without the generic typing. I'll try to think of a way to keep it minimalist and consistent. Let me know if you have any ideas.

Besides that, only missing changing the docs and we're good to go.

tomazfernandes:
This is what I thought for the `receive` APIs. Basically, two typed and two not typed for each variant. Let me know what you think - I'll try to update the docs today (BRT).

```java
// Use default settings provided when building the `SqsTemplate`
Optional<Message<?>> message1 = template.receive();  
Collection<Message<?>> messages1 = template.receiveMany();  

// Basic functionality - provide a queue and a type.
Optional<Message<String>> message2 = template.receive("my-queue", String.class);  
Collection<Message<String>> messages2 = template.receiveMany("my-queue", String.class);  

// Extended options without type
Optional<Message<?>> message3 = template.receive(from -> from
    .queue("my-queue")
    .receiveRequestAttemptId(UUID.randomUUID())
);
Collection<Message<?>> messages3 = template.receiveMany(from -> from.queue("my-queue"));  

// Extended options with type
Optional<Message<String>> message4 = template.receive(from -> from  
      .queue("my-queue")  
      .pollTimeout(Duration.ofSeconds(5)),  
   String.class);  

Collection<Message<String>> messages4 = template.receiveMany(from -> from
   .queue("my-queue")  
   .receiveRequestAttemptId(UUID.randomUUID()),
String.class);
```

tomazfernandes:
I think this is good to go! If there's anything else to change please let me know.

maciejwalkowiak:
Fantastic! Will do my best to review it today. Thanks @tomazfernandes !

maciejwalkowiak (approved):
LGTM ;-)

Issue #598, opened 2022-12-28
Abstract Container Options


tomazfernandes:
`ContainerOptions` wasn't on par with the rest of the integration in terms of architecture.

Change `ContainerOptions` to an `interface` and create an `AbstractContainerOptions` and `SqsContainerOptions`

PR #599, opened 2022-12-28
Abstract ContainerOptions (#598)

tomazfernandes:
Most changes are related to typing, actual changes are in the `ContainerOptions`, `AbstractContainerOptions` and `SqsContainerOptions` classes.

### DIFF
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -1,4 +1,4 @@
-== SQS Support
+== SQS Integration
 :source-highlighter: highlight.js
 :highlightjs-theme: googlecode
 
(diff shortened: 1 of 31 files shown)


tomazfernandes:
I'll add the missing javadocs in a future commit.

tomazfernandes:
About the intermittent error we're getting from local stack:
https://github.com/localstack/localstack/issues/7148

tomazfernandes:
I've added the missing javadocs in the `ContainerOptions`, but will do a last review before merging.

tomazfernandes:
I think this is good enough for now, and we can always edit javadocs later.

tomazfernandes:
If anyone would like to review, I'll keep it open. Thanks.

maciejwalkowiak:
I will!

tomazfernandes:
> I will!

Cool, thanks!

maciejwalkowiak (approved):
It looks good but why do we actually need this separation to `AbstractContainerOptions` and `SqsContainerOptions`?

tomazfernandes:
This is to keep it in sync with the rest of the architecture.

The way it was, the abstract classes for the listener would depend on ContainerOptions, which in turn had hard dependencies on SQS SDK classes.

This way, ContainerOptions and AbstractContainerOptions have no dependencies to SQS, like the rest of the architecture.

PR #605, opened 2023-01-03
Normalize Maven Variables

AlekSimpson:
Many POMs use variables like ${basedir}. That is the short form of ${project.basedir}. The form without prefix is deprecated: https://maven.apache.org/guides/introduction/introduction-to-the-pom.html#Available_Variables


maciejwalkowiak (approved):
Thanks @AlekSimpson!

Issue #578, opened 2022-12-14
Request to introduce SecretBinary available with AWS secretsmanager  


kishan-Dhrangadhariya:
**Type**: Feature

**Is your feature request related to a problem? Please describe.**
I have stored binary value on AWS secretsmanager which I need to fetch on Application startup 

**Describe the solution you'd like**
We can introduce optional parameter with every secret to identify whether it is binary or string and load them accordingly

**Describe alternatives you've considered**
Other way to go about this would be store secret in base64 encoded form, which would allow us to store binary data in string format, but this comes with its own tech debt

maciejwalkowiak:
Would you like to contribute a PR?

kishan-Dhrangadhariya:
Yes, I would like to contribute, if we could have a discussion on how best to implement this, it would be helpful.

maciejwalkowiak:
Figuring out how to implement it is probably 80% of work. I currently don't have time for that, but if you research this topic please come back with your ideas I am happy to review them

maciejwalkowiak:
Done in #613

PR #613, opened 2023-01-04
Introduce SecretBinary type for SecretsManager

MatejNedic:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
#578 

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/config/secretsmanager/SecretsManagerConfigDataLoaderIntegrationTests.java
+++ b/spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/config/secretsmanager/SecretsManagerConfigDataLoaderIntegrationTests.java
@@ -26,6 +26,8 @@ import static org.testcontainers.shaded.org.awaitility.Awaitility.await;
 
 import io.awspring.cloud.autoconfigure.ConfiguredAwsClient;
 import java.io.IOException;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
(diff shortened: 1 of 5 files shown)


spring-cloud-aws-secrets-manager/src/main/java/io/awspring/cloud/secretsmanager/SecretsManagerPropertySource.java
```diff
@@ -94,7 +101,8 @@ private void readSecretValue(GetSecretValueRequest secretValueRequest) {
 			String[] parts = secretValueResponse.name().split("/");
 			String secretName = parts[parts.length - 1];
 			LOG.debug("Populating property retrieved from AWS Secrets Manager: " + secretName);
-			properties.put(secretName, secretValueResponse.secretString());
+			properties.put(secretName, secretValueResponse.secretString() != null ? secretValueResponse.secretString()
+					: secretValueResponse.secretBinary().asUtf8String());
```
maciejwalkowiak, line 105: if i get it right we assume here that the secret is a string. what if it is really a binary secret like a certificate that can't be represented as a string? I guess this won't work then, right?

MatejNedic: Hmmm interesting will try that out tomorrow.

spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/config/secretsmanager/SecretsManagerConfigDataLoaderIntegrationTests.java
```diff
@@ -117,6 +144,17 @@ void resolvesPropertyFromSecretsManager_PlainTextSecret_endingWithSlash() {
...
 	}
 
+	@Test
+	void resolvesPropertyFromSecretsManager_PlainTextSecret_SecretBinary() {
+		SpringApplication application = new SpringApplication(App.class);
+		application.setWebApplicationType(WebApplicationType.NONE);
+
+		try (ConfigurableApplicationContext context = runApplication(application,
+				"aws-secretsmanager:/blob/byte_certificate")) {
+			assertThat(context.getEnvironment().getProperty("byte_certificate")).isEqualTo("77,121,32,99,101,114,116,105,102,105,99,97,116,101");
```
maciejwalkowiak, line 154: we should rather take the same string as where secrets is created and call `getBytes` on it rather than comparing strings in such a weird way

maciejwalkowiak (changes requested):
There are conflicts to resolve

Issue #318, opened 2022-04-14
Simplify signing urls for putting and getting S3 objects


maciejwalkowiak:
Add 2 methods (likely to `S3Operations`), one for getting signed url for fetching S3 object, second for getting signed url for putting an S3 object.

Work can start once https://github.com/awspring/spring-cloud-aws/pull/314 is merged.

amit-github-personal:
Hi @maciejwalkowiak , I would love to work on this

amit-github-personal:
The pull request 314 has now been merged so can i start working on this?

maciejwalkowiak:
@amit-github-personal apologies I missed this notification. Go ahead if you still feel like doing so!

whisper-bye:
Is there any progress?

maciejwalkowiak:
@whisper-bye no, but if you would like to pick it up go ahead!

SaiUpadhyayula:
@whisper-bye Do you want to work on this still? If not I can work on this.

whisper-bye:
@SaiUpadhyayula please go ahead. I just made a quick fix.

placed my code here to assist anyone who may need it before a new release is available
```java
    @Override
    public String download(String key) {
        var getObjectRequest = GetObjectRequest.builder()
                .bucket(OssBucketNameEnum.PRIVATE.getValue())
                .key(key)
                .build();

        var getObjectPresignRequest = GetObjectPresignRequest.builder()
                .getObjectRequest(getObjectRequest)
                .signatureDuration(Duration.ofHours(1))
                .build();

        try (var s3Presigner = S3Presigner.builder()
                .region(Region.of(Objects.requireNonNull(s3Properties.getRegion())))
                .credentialsProvider(getCredentialsProvider())
                .endpointOverride(s3Properties.getEndpoint())
                .build()) {
            var url = s3Presigner.presignGetObject(getObjectPresignRequest).url();
            return url.getPath() + "?" + url.getQuery();
        }
    }

    private ProfileCredentialsProvider getCredentialsProvider() {
        InputStream content;
        try {
            content = CharSource.wrap("[default]\n" + "aws_access_key_id=" + accessKey + "\n" + "aws_secret_access_key="
                            + secretKey + "\n")
                    .asByteSource(StandardCharsets.UTF_8)
                    .openStream();
        } catch (IOException e) {
            throw new IllegalStateException(e);
        }

        return ProfileCredentialsProvider.builder()
                .profileFile(ProfileFile.builder()
[...]

PR #620, opened 2023-01-11
Basic s3 signed URL implementation

Belair34:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Added two methods to S3Operations and thus S3Template that simplify creating signed urls for getting/putting objects in S3.

It simplifies the creation of signed URLs for S3 into one function call each.

#318 

So far I've manually tested it by doing ./mvnw install and then importing 3.0.0-SNAPSHOT from my local repository in a fresh spring project. 

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
I'm not 100% sure this is what the issue was requesting, so please let me know! I also don't think the way I have the S3 presigner is ideal, but I'm unsure how I should set it up. I imagine it should be configured like the S3Client, but I was unable to figure out how that works exactly and could use some advice. If any of this seems promising, I will go ahead and add tests and what not.

### DIFF
--- a/spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/S3Operations.java
+++ b/spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/S3Operations.java
@@ -16,6 +16,8 @@
 package io.awspring.cloud.s3;
 
 import java.io.InputStream;
+import java.net.URL;
+
 import org.springframework.lang.Nullable;
 import software.amazon.awssdk.services.s3.model.CreateBucketResponse;
 
@@ -103,4 +105,24 @@ public interface S3Operations {
 	 * @return downloaded object represented as {@link S3Resource}
 	 */
 	S3Resource download(String bucketName, String key);
+
+	/**
+	 * Creates a signed URL for retrieving an object from S3.
+	 *
+	 * @param bucketName - the bucket name
+	 * @param key - the object key
+	 * @param durationMinutes - duration that the URL will work
+	 * @return a {@link URL} representing the signed URL
+	 */
+	URL createSignedGetURL(String bucketName, String key, int durationMinutes);
+
+	/**
+	 * Creates a signed URL for putting an object into S3.
+	 *
+	 * @param bucketName - the bucket name
+	 * @param key - the object key
+	 * @param durationMinutes - duration that the URL will work
+	 * @return a {@link URL} representing the signed URL
+	 */
+	URL createSignedPutURL(String bucketName, String key, int durationMinutes);
 }
--- a/spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/S3Template.java
+++ b/spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/S3Template.java
@@ -17,11 +17,20 @@ package io.awspring.cloud.s3;
 
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.net.URL;
+import java.time.Duration;
+
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.StreamUtils;
 import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.model.GetObjectRequest;
 import software.amazon.awssdk.services.s3.model.PutObjectRequest;
+import software.amazon.awssdk.services.s3.presigner.S3Presigner;
+import software.amazon.awssdk.services.s3.presigner.model.GetObjectPresignRequest;
+import software.amazon.awssdk.services.s3.presigner.model.PresignedGetObjectRequest;
+import software.amazon.awssdk.services.s3.presigner.model.PresignedPutObjectRequest;
+import software.amazon.awssdk.services.s3.presigner.model.PutObjectPresignRequest;
 
 /**
  * Higher level abstraction over {@link S3Client} providing methods for the most common use cases.
@@ -126,6 +135,36 @@ public class S3Template implements S3Operations {
 		Assert.notNull(key, "key is required");
 
 		return new S3Resource(bucketName, key, s3Client, s3OutputStreamProvider);
+	}
+
+	@Override
+	public URL createSignedGetURL(String bucketName, String key, int durationMinutes){
+		S3Presigner presigner = S3Presigner.builder().build();
+		GetObjectRequest getObjectRequest = GetObjectRequest.builder()
+			.bucket(bucketName)
+			.key(key).build();
+
+		GetObjectPresignRequest presignRequest = GetObjectPresignRequest.builder()
+			.getObjectRequest(getObjectRequest)
+			.signatureDuration(Duration.ofMinutes(durationMinutes)).build();
+
+		PresignedGetObjectRequest signedRequest = presigner.presignGetObject(presignRequest);
+		return signedRequest.url();
+	}
+
+	@Override
+	public URL createSignedPutURL(String bucketName, String key, int durationMinutes){
+		S3Presigner presigner = S3Presigner.builder().build();
+		PutObjectRequest putObjectRequest = PutObjectRequest.builder()
+			.bucket(bucketName)
+			.key(key).build();
+
+		PutObjectPresignRequest presignRequest = PutObjectPresignRequest.builder()
+			.putObjectRequest(putObjectRequest)
+			.signatureDuration(Duration.ofMinutes(durationMinutes)).build();
+
+		PresignedPutObjectRequest signedRequest = presigner.presignPutObject(presignRequest);
+		return signedRequest.url();
 	}
 
 }


spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/S3Operations.java
```diff
@@ -103,4 +105,24 @@ default S3Resource upload(String bucketName, String key, InputStream inputStream
...
+
+	/**
+	 * Creates a signed URL for putting an object into S3.
+	 *
+	 * @param bucketName - the bucket name
+	 * @param key - the object key
+	 * @param durationMinutes - duration that the URL will work
+	 * @return a {@link URL} representing the signed URL
+	 */
+	URL createSignedPutURL(String bucketName, String key, int durationMinutes);
```
maciejwalkowiak, line 127: There should be also a way to set object metadata and I think set the content type? I imagine the object uploaded from the signed put url does not get correct content type automatically, so I think one more method is required here:

```java
URL createSignedPutURL(String bucketName, String key, Duration duration, @Nullable ObjectMetadata metadata, @Nullable String contentType);
```

spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/S3Operations.java
```diff
@@ -103,4 +105,24 @@ default S3Resource upload(String bucketName, String key, InputStream inputStream
...
+
+	/**
+	 * Creates a signed URL for retrieving an object from S3.
+	 *
+	 * @param bucketName - the bucket name
+	 * @param key - the object key
+	 * @param durationMinutes - duration that the URL will work
+	 * @return a {@link URL} representing the signed URL
+	 */
+	URL createSignedGetURL(String bucketName, String key, int durationMinutes);
```
maciejwalkowiak, line 117: Instead of `int` for duration we should use `java.time.Duration`

maciejwalkowiak (commented):
Thanks a lot @Belair34! I left few comments. Make sure to add integration tests for that. I think Localstack supports signed urls with S3.

spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/S3Template.java
```diff
@@ -128,4 +137,34 @@ public S3Resource download(String bucketName, String key) {
 		return new S3Resource(bucketName, key, s3Client, s3OutputStreamProvider);
 	}
 
+	@Override
+	public URL createSignedGetURL(String bucketName, String key, int durationMinutes){
+		S3Presigner presigner = S3Presigner.builder().build();
```
maciejwalkowiak, line 142: `S3Presigner` should be a bean created in `S3AutoConfiguration`. It should have `@ConditionalOnMissingBean` so that users can overwrite it, and should take `S3Configuration` from `S3Properties`.

Belair34:
@maciejwalkowiak I updated everything based on your comments! I do have one concern: I can't get all of the tests to pass even when I'm on the up-to-date main branch. It always fails at `ParameterStoreConfigDataLoaderIntegrationTests.whenKeysAreNotSpecifiedFailsWithHumanReadableFailureMessage`. I'm on Windows 10 using Java 17 for the record. 
Also, about docs: do I need to add anything besides the javadocs in S3Operations.java?

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/s3/S3AutoConfiguration.java
```diff
@@ -78,8 +82,14 @@ S3ClientBuilder s3ClientBuilder(AwsClientBuilderConfigurer awsClientBuilderConfi
...
-			S3ObjectConverter s3ObjectConverter) {
-		return new S3Template(s3Client, s3OutputStreamProvider, s3ObjectConverter);
+			S3ObjectConverter s3ObjectConverter, S3Presigner s3Presigner) {
+		return new S3Template(s3Client, s3OutputStreamProvider, s3ObjectConverter, s3Presigner);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	S3Presigner s3Presigner(S3Properties properties) {
+		return S3Presigner.builder().serviceConfiguration(properties.toS3Configuration()).build();
```
maciejwalkowiak, line 92: we also must set here credentials and region provider, endpoint override, fips enabled etc. Same things as we do to all the clients in `AwsClientBuilderConfigurer#configure`.

spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/S3Operations.java
```diff
@@ -103,4 +106,25 @@ default S3Resource upload(String bucketName, String key, InputStream inputStream
...
+
+	/**
+	 * Creates a signed URL for putting an object into S3.
+	 *
+	 * @param bucketName - the bucket name
+	 * @param key - the object key
+	 * @param duration - duration that the URL will work
+	 * @return a {@link URL} representing the signed URL
+	 */
+	URL createSignedPutURL(String bucketName, String key, Duration duration, @Nullable ObjectMetadata metadata,
```
maciejwalkowiak, line 128: Please add here one more method with a default implementation on the interface level:

```java
default URL createSignedPutURL(String bucketName, String key, Duration duration) {
  return createSignedPutURL(bucketName, key, duration, null, null);
}
```

spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/S3Template.java
```diff
@@ -128,4 +141,39 @@ public S3Resource download(String bucketName, String key) {
...
+
+	@Override
+	public URL createSignedPutURL(String bucketName, String key, Duration duration, @Nullable ObjectMetadata metadata,
+			@Nullable String contentType) {
+		Assert.notNull(bucketName, "bucketName is required");
+		Assert.notNull(key, "key is required");
+		Assert.notNull(duration, "duration is required");
+
+		PutObjectRequest.Builder putObjectRequestBuilder = PutObjectRequest.builder().bucket(bucketName).key(key);
+		if(metadata != null) putObjectRequestBuilder.metadata(metadata.getMetadata());
```
maciejwalkowiak, line 167: for the sake of consistency:
```java
if(metadata != null) {
  putObjectRequestBuilder.metadata(metadata.getMetadata());
}
```
same with `contentType`

spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/test/java/io/awspring/cloud/s3/S3TemplateIntegrationTests.java
```diff
@@ -183,6 +200,45 @@ void downloadsFile() throws IOException {
...
+		HttpPut httpPut = new HttpPut(signedPutUrl.toString());
+		httpPut.setHeader("x-amz-meta-testkey", "testvalue");
+		httpPut.setHeader("Content-Type","text/plain");
+		HttpEntity body = new StringEntity("hello");
+		httpPut.setEntity(body);
+
+		HttpResponse response = httpClient.execute(httpPut);
+		httpClient.close();
+
+		HeadObjectResponse headObjectResponse = client.headObject(HeadObjectRequest.builder()
```
maciejwalkowiak, line 233: Genuine question - shouldn't we upload something in this test?

maciejwalkowiak (changes requested):
Thanks for an update. I fixed the compilation issue. There are few more things needed - if you find it unclear please let me know!

maciejwalkowiak:
Regarding `ParameterStoreConfigDataLoaderIntegrationTests` - make sure you are up to date with main branch and just in case run `./mvnw clean` - this test used to be flaky but should not be any longer.

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/s3/S3AutoConfiguration.java
Belair34: Let me know if I did this right, I was a bit unsure.

spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/test/java/io/awspring/cloud/s3/S3TemplateIntegrationTests.java
Belair34: I guess it technically isn't uploading a local file, but it is adding a new file "file.txt" to the bucket with the request body as its contents and then checking to make sure it is there at the end. Personally, I think knowing this works is enough to be confident that the signed URL itself works. The difference is just about how you use it. That's just my interpretation, though. I'm open to other opinions!

Belair34:
Everything should be ready for review again! By the way, I found out the tests were failing for me because of lines like this:
https://github.com/awspring/spring-cloud-aws/blob/4d9da9a46185559e1b0a0308eae5eb0f31f140f1/spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/config/parameterstore/ParameterStoreConfigDataLoaderIntegrationTests.java#L118-L120
Since I'm on Windows all my new lines were '\r\n' instead of '\n' so the assertions were failing. I temporarily replaced these '\n' with `System.lineSeparator()` and all of my tests passed.

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/s3/S3AutoConfiguration.java
```diff
@@ -78,8 +81,20 @@ S3ClientBuilder s3ClientBuilder(AwsClientBuilderConfigurer awsClientBuilderConfi
...
+		return new S3Template(s3Client, s3OutputStreamProvider, s3ObjectConverter, s3Presigner);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	S3Presigner s3Presigner(S3Properties properties, AwsProperties awsProperties,
+			AwsCredentialsProvider credentialsProvider, AwsRegionProvider regionProvider) {
+		S3Presigner.Builder builder = S3Presigner.builder().serviceConfiguration(properties.toS3Configuration())
+				.credentialsProvider(credentialsProvider).region(regionProvider.getRegion());
+		Optional.ofNullable(awsProperties.getEndpoint()).ifPresent(builder::endpointOverride);
```
maciejwalkowiak, line 94: endpoint can be either set on global `AwsProperties` (`spring.cloud.aws.endpoint`) or specific to S3 `S3Properties` - `spring.cloud.aws.s3.endpoint`. 

if global endpoint set -> set it on builder
if s3 endpoint set -> set it on builder

S3 endpoint configuration if set has higher precedence than endpoint configured globally.

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/s3/S3AutoConfiguration.java
```diff
@@ -78,8 +81,20 @@ S3ClientBuilder s3ClientBuilder(AwsClientBuilderConfigurer awsClientBuilderConfi
...
 	S3Template s3Template(S3Client s3Client, S3OutputStreamProvider s3OutputStreamProvider,
-			S3ObjectConverter s3ObjectConverter) {
-		return new S3Template(s3Client, s3OutputStreamProvider, s3ObjectConverter);
+			S3ObjectConverter s3ObjectConverter, S3Presigner s3Presigner) {
+		return new S3Template(s3Client, s3OutputStreamProvider, s3ObjectConverter, s3Presigner);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	S3Presigner s3Presigner(S3Properties properties, AwsProperties awsProperties,
```
maciejwalkowiak, line 90: Update `S3AutoConfigurationTests` to test if this bean is created correctly with different properties.

maciejwalkowiak (changes requested):
Thanks for an update @Belair34. I left few comments we also need to update the reference docs (look `s3.adoc`). If you have any questions let me know!

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/s3/S3AutoConfiguration.java
maciejwalkowiak: You would need to assert on the actual `S3Presigner` bean - but as you have probably noticed it does not expose methods to get the endpoint and other properties. We had the same issue with other AWS clients and solved it by creating a concept of [ConfiguredAwsClient](https://github.com/awspring/spring-cloud-aws/blob/main/spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/ConfiguredAwsClient.java) that use reflection to access properties that are interesting for tests. Take a look how it is used in `S3AutoConfigurationTests`: https://github.com/awspring/spring-cloud-aws/blob/main/spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/s3/S3AutoConfigurationTests.java#L147

Of course it is not ideal since the test can start failing if internals of how AWS SDK works change, but it is definitely better than not testing at all or switching to integration tests that enable testing such properties.

Belair34: I did notice that, but I don't think I can use ConfiguredAwsClient because it takes an SdkClient whereas S3Presigner inherits from SdkPresigner. Should I make a separate version called ConfiguredAwsPresigner that takes an SdkPresigner? Alternatively, since SdkPresigner and SdkClient both inherit from SdkAutoCloseable I could try changing ConfiguredAwsClient to take an SdkAutoCloseable.

maciejwalkowiak: Whatever makes sense - I would need to go through internals of each to see if they have same structure so that reflections work on them. Gut feeling is that making `ConfiguredS3Presigner` makes the most sense.

Belair34:
@maciejwalkowiak I tried to make ConfiguredS3Presigner, and based on my evaluation of their internals it seemed like I didn't really need to change anything besides names. It worked perfectly for the first test I added regarding endpoints, but my second test gets null for the common aws properties. I'm really not sure why this is. Any ideas? I could also use some recommendations on which properties to test if this seems like the path forward.

maciejwalkowiak:
@Belair34 I fixed the test and did some final polishing.

maciejwalkowiak (approved):

maciejwalkowiak:
@MatejNedic if you have time please review.

spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/S3Operations.java
```diff
@@ -103,4 +105,29 @@ default S3Resource upload(String bucketName, String key, InputStream inputStream
...
+
+	/**
+	 * Creates a signed URL for retrieving an object from S3.
+	 *
+	 * @param bucketName - the bucket name
+	 * @param key - the object key
+	 * @param duration - duration that the URL will work
+	 * @return a {@link URL} representing the signed URL
+	 */
+	URL createSignedGetURL(String bucketName, String key, Duration duration);
```
MatejNedic, line 117: Should we return signedHeaders, SignedPayload, and isBrowserExecutable as well here?

Since in AWS docs for signedHeaders -> ```Returns the subset of headers that were signed, and MUST be included in the presigned request to prevent the request from failing.```

spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/S3Operations.java
```diff
@@ -103,4 +105,29 @@ default S3Resource upload(String bucketName, String key, InputStream inputStream
...
+
+	/**
+	 * Creates a signed URL for putting an object into S3.
+	 *
+	 * @param bucketName - the bucket name
+	 * @param key - the object key
+	 * @param duration - duration that the URL will work
+	 * @return a {@link URL} representing the signed URL
+	 */
+	URL createSignedPutURL(String bucketName, String key, Duration duration, @Nullable ObjectMetadata metadata,
```
MatejNedic, line 127: Should we return signedHeaders, SignedPayload, and isBrowserExecutable as well here?

MatejNedic (changes requested):
I have a few questions regarding what we should be returned but otherwise looks good!

spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/S3Operations.java
maciejwalkowiak: Now I am quite confident we need an end to end test for that.

PR #622, opened 2023-01-12
Add option to set a property prefix on Secrets Manager property sources

### LINKED ISSUE #621: Retrieve multiple secrets in that have same keys
### Discussed in https://github.com/awspring/spring-cloud-aws/discussions/388
To summarize - the issue is that when multiple secrets containing JSON properties are imported, and same keys exist in both secret JSONs then value from one secret overwrites the other.

For example

- `/secret/first-secret`:

```json
{ "username": "postgres" }
```

- `/secret/second-secret`:
```json
{ "username": "root" }
```

I am thinking how this can be solved. We could add a prefix option:

```yml
spring:
    config:
         import: aws-secretsmanager:/secrets/first-secret?prefix=first,aws-secretsmanager:/secret/second-secret?prefix=second
```

Then they would be imported as:

```
first.username=postgres
second.username=root
```

maciejwalkowiak:
Fixes #621

### DIFF
--- a/docs/src/main/asciidoc/secrets-manager.adoc
+++ b/docs/src/main/asciidoc/secrets-manager.adoc
@@ -114,6 +114,10 @@ Secret value can be retrieved by referencing secret name:
 spring.datasource.url=${jdbc-url}
 ----
 
+==== Handling property names collisions
+
+
+
 === Using SecretsManagerClient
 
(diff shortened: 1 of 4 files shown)


maciejwalkowiak:
Docs pending

MatejNedic (approved):
Looks good to me. Edit: nvm

Issue #631, opened 2023-01-19
Add Acknowledgement Shutdown Timeout


tomazfernandes:
When acknowledgement batching is used, message processing and acknowledgement are really two separate worlds.

Given `Spring` lifecycle interface methods doesn't accept a `Duration timeout` parameter, I think it'll be clearer for users to have separate timeouts.

After waiting for message processing to finish (or timeout), we'll start waiting for any remaining acknowledgements to finish.

So the maximum amount of time a `Container` will wait is `listenerShutdownTimeout` + `acknowledgementShutdownTimeout`.

Currently the same value is used for both, but even if we document that I think it'll be misleading.

maciejwalkowiak:
Done in #632

PR #632, opened 2023-01-19
Add Acknowledgement Shutdown Timeout (#631)

tomazfernandes:
See #631


tomazfernandes:
With this one I think I'm all set code-wise for RC.

After the SqsTemplate is merged I'll add it to the sample.

There are a few more blind spots of test coverage overall than I'd like, but I think it's good enough and I'll work on adding more tests when I have the time.

tomazfernandes:
I've added another commit with an API cleanup for the ContainerOptions. It probably should have been another PR, but I was afraid there would be too many conflicts with this one. The individual commits are preserved though.

maciejwalkowiak (approved):
Does it need an update in reference docs?

tomazfernandes:
@maciejwalkowiak with this I'm done!

Feel free to merge when you'd like, and let me know if you have any questions.

Looking forward to the RC!

PR #619, opened 2023-01-11
Add option to resolve SQS queue by ARN

techadri:
Fixes #561

## :loudspeaker: Type of change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

This PR introduces the option to use the queue ARN to resolve an SQS queue when specifying an SQSListener. This enables the user to listen to queues from other accounts than the account used by the application/sqs client.

**Limitation:** As the used AmazonSQS client is region-specific, queues from other regions cannot be resolved. However, this can still be achieved using the queue url. 

## :bulb: Motivation and Context

See issue #561 
Since it is common to access AWS resources via the ARN, this should also be possible for the SqsListener.

## :green_heart: How did you test it?
I built the project locally and ran the unit tests. In addition, I used a small sample application to check locally that the ARN is resolved correctly when specified as value in the SqsListener annotation. 

## :pencil: Checklist

- [x] I reviewed submitted code
- [x] I added tests to verify changes
- [x] I updated reference documentation to reflect the change
- [x] All tests passing
- [x] No breaking changes

PR #634, opened 2023-01-21
Add option to resolve sqs queue by ARN in SqsListener

### LINKED ISSUE #561: SQS should resolve ARN queue names
**Type**: Feature

The `DynamicQueueUrlDestinationResolver` can handle explicit SQS URLs, and it will perform a `GetQueueUrl` to fetch the URL when provided a simple queue name. When provided an ARN, however, it passes the ARN to Amazon as the queue name, which results in a queue-not-found error.

The resolver already checks to see whether the `String name` is a valid HTTP or HTTPS URL and uses that if provided. I would like it to also support queue `arn:` URIs. If it is advantageous to fetch the queue URL for some reason (to fast-fail?), then that request can be built parsing the ARN. If it is not so adventageous, the ARN can be converted into the queue URL via string operations.

The main alternative, when referring to queues potentially not owned by the running AWS account, would be to provide the URL. This is different from how most AWS resources are usually identified in configuration.

**Additional context**
This is specifically a concern with the existing 2.x series (including 2.4.2). If this functionality is worthwhile and the 2.4 branch is being maintained, I can try to put together a PR to implement it.

techadri:
Fixes #561

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

This PR adds the same functionality that has been added to version 2.4.x in PR #619 

This PR introduces the option to use the queue ARN to resolve an SQS queue when specifying an SQSListener. This enables the user to listen to queues from other accounts than the account used by the application/sqs client.

**Limitation:** As the used AmazonSQS client is region-specific, queues from other regions cannot be resolved.

See issue #561 
Since it is common to access AWS resources via the ARN, this should also be possible for the SqsListener.

I built the project locally and ran the tests. In addition, I used a small sample application to check locally that the ARN is resolved correctly when specified as value in the SqsListener annotation.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


maciejwalkowiak (approved):
Looks good to me! @tomazfernandes could you please take a look?

tomazfernandes:
> Looks good to me! @tomazfernandes could you please take a look?

Sure, I'll take a look! Thanks @techadri!

docs/src/main/asciidoc/sqs.adoc
```diff
@@ -638,11 +638,11 @@ The `MessageListenerContainer` interface extends `SmartLifecycle`, which provide
...
 
-At startup, the containers will make requests to `SQS` to retrieve the queues` urls for the provided queue names, and for retrieving `QueueAttributes` if so configured.
+At startup, the containers will make requests to `SQS` to retrieve the queues` urls for the provided queue names or ARNs, and for retrieving `QueueAttributes` if so configured.
 Providing queue urls instead of names and not requesting queue attributes can result in slightly better startup times since there's no need for such requests.
 
 NOTE: If retrieving the queue url fails due to the queue not existing, the framework can be configured to either create the queue or fail.
-If a URL is provided instead of a queue name the framework will not make this request at startup, and thus if the queue does not exist it will fail at runtime.
+If a URL or ARN is provided instead of a queue name the framework will not make this request at startup, and thus if the queue does not exist it will fail at runtime.
 This configuration is available in `SqsContainerOptions` `queueNotFoundStrategy.`
```
tomazfernandes, line 646: If I understand correctly, if an ARN is provided the framework will make the request, right?

spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration/QueueAttributesResolverIntegrationTests.java
```diff
@@ -127,6 +127,31 @@ void shouldResolveFromUri() {
 			client.deleteQueue(DeleteQueueRequest.builder().queueUrl(queueUrl).build()).join();
 		}
 	}
+
+	@Test
```
tomazfernandes, line 131: The test passes when running against local stack, but fails when I run against AWS:
`Caused by: software.amazon.awssdk.services.sqs.model.SqsException: The address https://sqs.us-east-1.amazonaws.com/ is not valid for this endpoint. (Service: Sqs, Status Code: 404, Request ID: 9d4c355f-c622-5eca-a4a3-6916d1f77ab5)`

If you can please set `useLocalStackClient` to off in the `BaseSqsIntegrationTest`class and re run it.

tomazfernandes (changes requested):
Thanks @techadri for the PR!

A couple of notes, and please add you name to the authors tags.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/QueueAttributesResolver.java
```diff
@@ -142,6 +153,16 @@ private boolean isValidQueueUrl(String name) {
 		}
 	}
 
+	private boolean isValidQueueArn(String queueArn) {
```
MatejNedic, line 156: Can we return Arn or null? So it can be reused in `doResolveQueueUrlFromArn` method

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/QueueAttributesResolver.java
```diff
@@ -91,8 +92,18 @@ private CompletableFuture<String> resolveQueueUrl() {
 	}
 
 	private CompletableFuture<String> doResolveQueueUrl() {
+		return isValidQueueArn(this.queueName)
+			? doResolveQueueUrlFromArn()
+			: CompletableFutures.exceptionallyCompose(this.sqsAsyncClient.getQueueUrl(req -> req.queueName(this.queueName))
+				.thenApply(GetQueueUrlResponse::queueUrl), this::handleException);
+	}
+
+	private CompletableFuture<String> doResolveQueueUrlFromArn() {
```
MatejNedic, line 101: We can pass Arn here and skip the second call. 
`private CompletableFuture<String> doResolveQueueUrlFromArn(Arn arn) {`

Instead of doing second time `Arn.fromString`
`Arn arn = Arn.fromString(this.queueName);`

MatejNedic (changes requested):
Looks good to me I would just suggest the following change.

docs/src/main/asciidoc/sqs.adoc
techadri: Correct, I have removed the reference to the ARN as it does not belong there.

spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration/QueueAttributesResolverIntegrationTests.java
techadri: @tomazfernandes  I have adapted the test so that it now also runs successfully against AWS. I am happy to receive feedback on whether the solution is acceptable.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/QueueAttributesResolver.java
techadri: @MatejNedic Thanks for the suggestion, I have amended that.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/QueueAttributesResolver.java
techadri: Thanks for the suggestion, I have changed that as well.

techadri:
@tomazfernandes @MatejNedic Many thanks for the feedback. I have incorporated the suggestions and am looking forward to another review.

MatejNedic (approved):

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/QueueAttributesResolver.java
```diff
@@ -91,8 +94,18 @@ private CompletableFuture<String> resolveQueueUrl() {
 	}
 
 	private CompletableFuture<String> doResolveQueueUrl() {
+		Arn arn = convertToQueueArn(this.queueName);
+		return arn != null
```
tomazfernandes, line 98: Well, since we've got this far, we could take a step further and try to handle both calls in the same method, so that we don't duplicate what comes after e.g. `handleException`.

What do you think? Definitely a nit / suggestion, not a blocker.

tomazfernandes (commented):
Thanks @techadri for the changes!

Left one comment, please let me know what you think.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/QueueAttributesResolver.java
tomazfernandes: Maybe something along these lines, perhaps better arranged:

```java
	private CompletableFuture<String> doResolveQueueUrl() {
		Arn arn = convertToQueueArn(this.queueName);
		return CompletableFutures.exceptionallyCompose((arn != null
			? this.sqsAsyncClient.getQueueUrl(req -> req.queueName(arn.resourceAsString()).queueOwnerAWSAccountId(arn.accountId().get()))
			: this.sqsAsyncClient.getQueueUrl(req -> req.queueName(this.queueName)))
				.thenApply(GetQueueUrlResponse::queueUrl), this::handleException);
	}
```

techadri: Thank you @tomazfernandes for the feedback. I also think it makes sense to combine this into one method. I have added a commit with a possible solution.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/QueueAttributesResolver.java
```diff
@@ -91,9 +95,18 @@ private CompletableFuture<String> resolveQueueUrl() {
 	}
 
 	private CompletableFuture<String> doResolveQueueUrl() {
-		return CompletableFutures
-			.exceptionallyCompose(this.sqsAsyncClient.getQueueUrl(req -> req.queueName(this.queueName))
-				.thenApply(GetQueueUrlResponse::queueUrl), this::handleException);
+		Arn arn = convertToQueueArn(this.queueName);
```
tomazfernandes, line 98: Minor nit: why pass `this.queueName` as an argument? We could simply use it in the method right?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/QueueAttributesResolver.java
tomazfernandes: Cool, thanks @techadri!

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/QueueAttributesResolver.java
techadri: Definitely, right! I have changed that as well.

tomazfernandes:
Looks great to me, thanks @techadri! If no one else has any further suggestions, I'll squash and merge it.

tomazfernandes (approved):

tomazfernandes:
Thanks @techadri for the contribution! Hopefully more to come!

techadri:
Thanks @tomazfernandes, I hope that I will be able to contribute from time to time!

PR #638, opened 2023-01-27
Load secret as string if json parse fails

snigdhasjg:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

If secret is not a JSON string, it will take it as string. 
Name of spring property will be secret name last suffix. 
eg. if secret name is `service1/secret1` the property name will be `secret1`.
Similar approach as `3.0.x`

This also inherits behaviour like adding prefix to the property name.

Fixes #636

Unit test only.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


maciejwalkowiak:
Thanks @snigdhasjg !

Issue #588, opened 2022-12-15
Unable to resolve SNS ARNs


PatrickGotthard:
**Type**: Bug

**Component**: SNS

**Describe the bug**
After upgrading to 3.0.0-M3 I'm unable to send SNS Notifications due to broken resolution of SNS ARNs:
```
software.amazon.awssdk.services.sns.model.InvalidParameterException: Invalid parameter: Topic Name (Service: Sns, Status Code: 400, Request ID: fa3d86c8-c434-59fe-baf0-daddd9b97589)
	at software.amazon.awssdk.core.internal.http.CombinedResponseHandler.handleErrorResponse(CombinedResponseHandler.java:125)
	at software.amazon.awssdk.core.internal.http.CombinedResponseHandler.handleResponse(CombinedResponseHandler.java:82)
	at software.amazon.awssdk.core.internal.http.CombinedResponseHandler.handle(CombinedResponseHandler.java:60)
	at software.amazon.awssdk.core.internal.http.CombinedResponseHandler.handle(CombinedResponseHandler.java:41)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.HandleResponseStage.execute(HandleResponseStage.java:40)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.HandleResponseStage.execute(HandleResponseStage.java:30)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallAttemptTimeoutTrackingStage.execute(ApiCallAttemptTimeoutTrackingStage.java:73)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallAttemptTimeoutTrackingStage.execute(ApiCallAttemptTimeoutTrackingStage.java:42)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.TimeoutExceptionHandlingStage.execute(TimeoutExceptionHandlingStage.java:78)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.TimeoutExceptionHandlingStage.execute(TimeoutExceptionHandlingStage.java:40)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallAttemptMetricCollectionStage.execute(ApiCallAttemptMetricCollectionStage.java:50)
[...]

PR #647, opened 2023-01-30
Introduce SNS resolver by list

MatejNedic:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
#588 

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sns/SnsAutoConfiguration.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sns/SnsAutoConfiguration.java
@@ -22,7 +22,6 @@ import io.awspring.cloud.autoconfigure.core.AwsClientBuilderConfigurer;
 import io.awspring.cloud.autoconfigure.core.AwsClientCustomizer;
 import io.awspring.cloud.autoconfigure.core.CredentialsProviderAutoConfiguration;
 import io.awspring.cloud.autoconfigure.core.RegionProviderAutoConfiguration;
-import io.awspring.cloud.sns.core.ListTopicArnResolverCached;
 import io.awspring.cloud.sns.core.SnsOperations;
 import io.awspring.cloud.sns.core.SnsTemplate;
 import io.awspring.cloud.sns.core.TopicArnResolver;
@@ -69,12 +68,6 @@ public class SnsAutoConfiguration {
 			ObjectProvider<AwsClientCustomizer<SnsClientBuilder>> configurer) {
 		return awsClientBuilderConfigurer.configure(SnsClient.builder(), properties, configurer.getIfAvailable())
 				.build();
-	}
-
-	@Bean
-	@ConditionalOnProperty(name = SnsProperties.PREFIX + ".resolve_by_list", havingValue = "true")
-	public TopicArnResolver topicArnResolver(SnsClient snsClient) {
-		return new ListTopicArnResolverCached(snsClient);
 	}
 
 	@ConditionalOnMissingBean(SnsOperations.class)
--- a/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sns/SnsAutoConfiguration.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sns/SnsAutoConfiguration.java
@@ -22,6 +22,7 @@ import io.awspring.cloud.autoconfigure.core.AwsClientBuilderConfigurer;
 import io.awspring.cloud.autoconfigure.core.AwsClientCustomizer;
 import io.awspring.cloud.autoconfigure.core.CredentialsProviderAutoConfiguration;
 import io.awspring.cloud.autoconfigure.core.RegionProviderAutoConfiguration;
+import io.awspring.cloud.sns.core.ListTopicArnResolverCached;
 import io.awspring.cloud.sns.core.SnsOperations;
 import io.awspring.cloud.sns.core.SnsTemplate;
 import io.awspring.cloud.sns.core.TopicArnResolver;
@@ -68,6 +69,12 @@ public class SnsAutoConfiguration {
 			ObjectProvider<AwsClientCustomizer<SnsClientBuilder>> configurer) {
 		return awsClientBuilderConfigurer.configure(SnsClient.builder(), properties, configurer.getIfAvailable())
 				.build();
+	}
+
+	@Bean
+	@ConditionalOnProperty(name = SnsProperties.PREFIX + ".resolve_by_list", havingValue = "true")
+	public TopicArnResolver topicArnResolver(SnsClient snsClient) {
+		return new ListTopicArnResolverCached(snsClient);
 	}
 
 	@ConditionalOnMissingBean(SnsOperations.class)
--- a/spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/sns/SnsAutoConfigurationTest.java
+++ b/spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/sns/SnsAutoConfigurationTest.java
@@ -41,7 +41,6 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.lang.Nullable;
 import org.springframework.test.util.ReflectionTestUtils;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
-import org.testcontainers.junit.jupiter.Testcontainers;
 import software.amazon.awssdk.arns.Arn;
 import software.amazon.awssdk.core.client.config.ClientOverrideConfiguration;
 import software.amazon.awssdk.core.client.config.SdkClientOption;
@@ -55,7 +54,6 @@ import software.amazon.awssdk.services.sns.SnsClientBuilder;
  *
  * @author Matej Nedic
  */
-@Testcontainers
 class SnsAutoConfigurationTest {
 
 	private final ApplicationContextRunner contextRunner = new ApplicationContextRunner()
--- a/spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/sns/SnsAutoConfigurationTest.java
+++ b/spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/sns/SnsAutoConfigurationTest.java
@@ -17,14 +17,12 @@ package io.awspring.cloud.autoconfigure.sns;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.mock;
-import static org.testcontainers.containers.localstack.LocalStackContainer.Service.SNS;
 
 import io.awspring.cloud.autoconfigure.ConfiguredAwsClient;
 import io.awspring.cloud.autoconfigure.core.AwsAutoConfiguration;
 import io.awspring.cloud.autoconfigure.core.AwsClientCustomizer;
 import io.awspring.cloud.autoconfigure.core.CredentialsProviderAutoConfiguration;
 import io.awspring.cloud.autoconfigure.core.RegionProviderAutoConfiguration;
-import io.awspring.cloud.sns.core.ListTopicArnResolverCached;
 import io.awspring.cloud.sns.core.SnsOperations;
 import io.awspring.cloud.sns.core.SnsTemplate;
 import io.awspring.cloud.sns.core.TopicArnResolver;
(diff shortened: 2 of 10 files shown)


spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/sns/SnsAutoConfigurationTest.java
```diff
@@ -41,21 +44,34 @@
...
 /**
  * Tests for class {@link io.awspring.cloud.autoconfigure.sns.SnsAutoConfiguration}.
  *
  * @author Matej Nedic
  */
+@Testcontainers
 class SnsAutoConfigurationTest {
 
+	@Container
+	static LocalStackContainer localstack = new LocalStackContainer(
```
MatejNedic, line 72: We need this since when creating  `ListTopicArnResolverCached` we make caching call on initialization.

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/core/ListTopicArnResolverCached.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+public class ListTopicArnResolverCached implements TopicArnResolver {
+
+	private final SnsClient snsClient;
+
+	private final ConcurrentHashMap<String, Arn> cache = new ConcurrentHashMap<>();
+
+	public ListTopicArnResolverCached(SnsClient snsClient) {
+		this.snsClient = snsClient;
+		resolveByCallingPaginator();
```
MatejNedic, line 38: Should we initialize it at initialization?
Can this map become 2 large? Also, if it is 2 large should the application have permission to see other topics that are not in the scope of the application?
cc @maciejwalkowiak

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/core/ListTopicArnResolverCached.java
```diff
@@ -0,0 +1,70 @@
...
+	public Arn resolveTopicArn(String topicName) {
+		Assert.notNull(topicName, "topicName must not be null");
+		if (topicName.toLowerCase().startsWith("arn:")) {
+			return Arn.fromString(topicName);
+		}
+		else if (cache.get(topicName) != null) {
+			return cache.get(topicName);
+		}
+		else {
+			resolveByCallingPaginator();
```
MatejNedic, line 55: I had multi-threading here with locks but maybe it is overkill? Should I return it back?

tomazfernandes: Hey @MatejNedic! I think the issue here with concurrency is if concurrent calls for non-existing keys trigger the cache refresh simultaneously, we'll have unbounded concurrent API calls. So I agree some synchronization would be called for :)

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/core/ListTopicArnResolverCached.java
tomazfernandes: Well, I'll drop my 2c, hope it's ok :)

> Should we initialize it at initialization?

IMHO we should not, as this can impact the application startup time. I'd rather keep it lazy and populate the cache on the first call.

> Can this map become 2 large? 

I think only if users have too many topics, like thousands. Doesn't seem like a problem IMO.

> Also, if it is 2 large should the application have permission to see other topics that are not in the scope of the application?

I think this is more of an IAM problem than a framework problem, so I wouldn't worry.

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/core/ListTopicArnResolverCached.java
```diff
@@ -0,0 +1,70 @@
...
+			return cache.get(topicName);
+		}
+	}
+
+	private void resolveByCallingPaginator() {
+		Map<String, Arn> nameAndArn = snsClient.listTopicsPaginator().topics().stream().collect(
+				Collectors.toMap(t -> t.topicArn().substring(t.topicArn().lastIndexOf(":") + 1), topic -> Arn.fromString(topic.topicArn())));
+		nameAndArn.forEach((key, value) -> cache.merge(key, value, (v1, v2) -> v2));
```
tomazfernandes, line 63: Why not use `putAll`?

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/core/ListTopicArnResolverCached.java
maciejwalkowiak: Even if we decide to do it on the initialization, it should not be done in constructor but rather class should implement`InitializingBean` and do it in `afterPropertiesSet()`.

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sns/SnsAutoConfiguration.java
```diff
@@ -70,6 +71,12 @@ public SnsClient snsClient(SnsProperties properties, AwsClientBuilderConfigurer
 				.build();
 	}
 
+	@Bean
+	@ConditionalOnProperty(name = SnsProperties.PREFIX + ".resolve_by_list", havingValue = "true")
```
maciejwalkowiak, line 75: Let's take it out from the auto-configuration. If users want to use it, they can declare the bean themselves.

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/core/ListTopicArnResolverCached.java
```diff
@@ -0,0 +1,70 @@
...
+		}
+		else {
+			resolveByCallingPaginator();
+			return cache.get(topicName);
+		}
+	}
+
+	private void resolveByCallingPaginator() {
+		Map<String, Arn> nameAndArn = snsClient.listTopicsPaginator().topics().stream().collect(
+				Collectors.toMap(t -> t.topicArn().substring(t.topicArn().lastIndexOf(":") + 1), topic -> Arn.fromString(topic.topicArn())));
```
maciejwalkowiak, line 62: I think this deserves some refactoring because having `lastIndexOf` and `+1` looks to me too low level. I don't understand what is happening here.

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/core/ListTopicArnResolverCached.java
```diff
@@ -0,0 +1,70 @@
...
+import org.springframework.util.Assert;
+import software.amazon.awssdk.arns.Arn;
+import software.amazon.awssdk.services.sns.SnsClient;
+import software.amazon.awssdk.services.sns.model.Topic;
+
+/**
+ * @author Matej Nedić
+ * @since 3.0.0
+ */
+public class ListTopicArnResolverCached implements TopicArnResolver {
```
maciejwalkowiak, line 30: Cache gets never refreshed which for some use cases can be an issue.

tomazfernandes: Maybe instead of refreshing the cache on misses it could have a TTL and refresh when called after a configurable time.

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/core/ListTopicArnResolverCached.java
MatejNedic: Last part of ARN is name, so I am substringing last : to get it. No other way...

maciejwalkowiak (changes requested):
No caching
No extra configuration property

Users can define it as bean and autoconfiguration will pick it up.

Lets keep it simple, if anyone needs something more sophisticated they can create their own implementation.

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/core/ListTopicArnResolverCached.java
MatejNedic: How often arn is changed is questiom than as well. Never except recreating topic. Maybe we should just drop caching.

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/core/NotExistentTopicException.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package io.awspring.cloud.sns.core;
+
+/**
+ * Thrown when {@link ListTopicArnResolver} cannot determine ARN based on topicName.
+ *
+ * @author Matej Nedic
+ * @since 3.0.0
+ */
+public class NotExistentTopicException extends RuntimeException {
```
maciejwalkowiak, line 24: Somehow this name does not feel right. Maybe `TopicNotFoundException`?

spring-cloud-aws-sns/src/test/java/io/awspring/cloud/sns/integration/SnsTemplateIntegrationTest.java
```diff
@@ -106,4 +116,26 @@ void send_validPersonObject_usesTopicChannel_send_arn_read_sqs() {
 		});
 	}
 
+	@Test
+	void send_test_message_for_existing_topic_name_trigger_trigger_iteration() {
```
maciejwalkowiak, line 120: Lets wrap these two tests into @Nested and the rest also in @Nested to not have setup in @BeforeAll that is not relevant for all the test.

spring-cloud-aws-sns/src/test/java/io/awspring/cloud/sns/integration/SnsTemplateIntegrationTest.java
```diff
@@ -106,4 +116,26 @@ void send_validPersonObject_usesTopicChannel_send_arn_read_sqs() {
...
+	void send_test_message_for_not_existing_topic_name() {
+		ListTopicArnResolver listTopicArnResolver = new ListTopicArnResolver(snsClient);
+		SnsTemplate snsTemplateTestCache = new SnsTemplate(snsClient, listTopicArnResolver, null);
+		assertThatThrownBy(
+				() -> snsTemplateTestCache.sendNotification("Some_random_topic", "message content", "subject"))
+						.isInstanceOf(NotExistentTopicException.class);
+	}
+
+	private static void createTopics() {
+		for (int i = 0; i < 300; i++) {
```
maciejwalkowiak, line 136: Lets keep this number as low as possible - higher than the single response page size, but not so large as it only extends time needed to run tests.

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/core/ListTopicArnResolver.java
```diff
@@ -0,0 +1,81 @@
...
+import org.springframework.lang.Nullable;
+import org.springframework.util.Assert;
+import software.amazon.awssdk.arns.Arn;
+import software.amazon.awssdk.services.sns.SnsClient;
+import software.amazon.awssdk.services.sns.model.ListTopicsRequest;
+import software.amazon.awssdk.services.sns.model.ListTopicsResponse;
+import software.amazon.awssdk.services.sns.model.Topic;
+
+/**
+ * Basic implementation for resolving ARN from topicName. It is using {@link SnsClient#listTopics()} to determine topicArn for topicName.
```
maciejwalkowiak, line 28: Lets make it explicit in Javadocs that it iterates over all topics for each topic resolution

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/core/ListTopicArnResolver.java
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * @author Matej Nedić
+ * @since 3.0.0
+ */
+public class ListTopicArnResolver implements TopicArnResolver {
```
maciejwalkowiak, line 33: Also something does not feel right about this name. Maybe `TopicsListingTopicArnResolver`?

spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/core/ListTopicArnResolver.java
```diff
@@ -0,0 +1,81 @@
...
+
+	private final SnsClient snsClient;
+
+	public ListTopicArnResolver(SnsClient snsClient) {
+		this.snsClient = snsClient;
```
maciejwalkowiak, line 38: `Assert.notNull`

maciejwalkowiak (changes requested):

MatejNedic:
@maciejwalkowiak can we merge this?

Issue #666, opened 2023-02-09
Add Support for StsAssumeRoleCredentialsProvider


kamaydeo:
CredentialsProviderAutoConfiguration currently supports
1. StaticCredentialsProvider
2. ProfileCredentialsProvider
3. InstanceProfileCredentialsProvider

But it does not support StsAssumeRoleCredentialsProvider. I would like it to be supported or at least provide a way to override CredentialsProviderAutoConfiguration so I can use StsAssumeRoleCredentialsProvider.

neiljbrown:
Hi @kamaydeo, 

Given your reference to CredentialsProviderAutoConfiguration, I assume you're using Spring Cloud AWS 3.x.  If so, you can override the auto-configured default by declaring your own bean of type AwsCredentialsProvider. For more details see https://docs.awspring.io/spring-cloud-aws/docs/3.0.0-RC1/reference/html/index.html#credentials. 

If you're using 2.x then I believe you can also override the default by declaring your own bean of type AwsCredentialsProvider, but it additionally has to be named credentialsProvider.

Regards, Neil.

kamaydeo:
@neiljbrown Although the bean can be overridden as shown in documentation, unfortunately, it doesn't work for my use case. I'm loading config data from SecretsManager i.e. SecretsManagerConfigDataLoader
`spring.config.import=aws-secretsmanager:...`

``` java
@Configuration
class CustomCredentialsProviderConfiguration {

    @Bean
    public AwsCredentialsProvider customAwsCredentialsProvider() {
        return new CustomAWSCredentialsProvider();
    }
}
```
The static method in `CredentialsProviderAutoConfiguration.createCredentialsProvider(...)` is invoked by `AbstractAwsConfigDataLocationResolver. configure(..)` when I use spring.config.import=aws-secretsmanager:...

see line number 126 in `AbstractAwsConfigDataLocationResolver`. Note that at this time, beans are not even initialized so customAwsCredentialsProvider doesn't help me.

maciejwalkowiak:
@kamaydeo Secrets Manager and Parameter Store integration require slightly different approach for configuring credentials provider. Take a look at the example in docs: https://docs.awspring.io/spring-cloud-aws/docs/3.0.0-RC1/reference/html/index.html#customizing-secretsmanagerclient

maciejwalkowiak:
[...]

PR #691, opened 2023-02-25
Initial support for StsWebIdentityTokenFileCredentialsProvider

eduanb:
- [ ] Bugfix
- [ x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Adds support for StsWebIdentityTokenFileCredentialsProvider, a common use case in EKS.
Note: This is still WIP. I would first like feedback before adding tests, docs etc.

The recommended way of authenticating to AWS in EKS is using Service Accounts. See here https://docs.aws.amazon.com/eks/latest/userguide/pod-configuration.html

This PR adds native support for it. Users of spring-cloud-aws now only need to add the STS dependency and set up the required service account.  spring-cloud-aws will then automatically create a StsWebIdentityTokenFileCredentialsProvider. 

The inspiration behind this was https://github.com/awspring/spring-cloud-aws/issues/666

This still needs to be tested. I first want to hear initial feedback from the team if this will be a good feature to add. 

## :pencil: Checklist
- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/core/CredentialsProviderAutoConfiguration.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/core/CredentialsProviderAutoConfiguration.java
@@ -15,6 +15,7 @@
  */
 package io.awspring.cloud.autoconfigure.core;
 
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
@@ -24,6 +25,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
+import org.springframework.util.ClassUtils;
 import org.springframework.util.StringUtils;
 import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
@@ -33,6 +35,8 @@ import software.amazon.awssdk.auth.credentials.InstanceProfileCredentialsProvide
 import software.amazon.awssdk.auth.credentials.ProfileCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
 import software.amazon.awssdk.profiles.ProfileFile;
+import software.amazon.awssdk.services.sts.StsClient;
+import software.amazon.awssdk.services.sts.auth.StsWebIdentityTokenFileCredentialsProvider;
 
 /**
  * {@link EnableAutoConfiguration} for {@link AwsCredentialsProvider}.
@@ -45,6 +49,9 @@ import software.amazon.awssdk.profiles.ProfileFile;
 @EnableConfigurationProperties(CredentialsProperties.class)
 public class CredentialsProviderAutoConfiguration {
 
+	public static final String STS_WEB_IDENTITY_TOKEN_FILE_CREDENTIALS_PROVIDER = "software.amazon.awssdk.services.sts.auth.StsWebIdentityTokenFileCredentialsProvider";
+	public static final String AWS_ROLE_ARN_ENV_VARIABLE = "AWS_ROLE_ARN";
+	public static final String AWS_WEB_IDENTITY_TOKEN_FILE_ENV_VARIABLE = "AWS_WEB_IDENTITY_TOKEN_FILE";
 	private final CredentialsProperties properties;
 
 	public CredentialsProviderAutoConfiguration(CredentialsProperties properties) {
@@ -73,6 +80,10 @@ public class CredentialsProviderAutoConfiguration {
 			providers.add(createProfileCredentialProvider(profile));
 		}
 
+		if (shouldCreateStsIdentityTokenCredentialsProvider()) {
+			providers.add(createStsCredentialsProvider());
+		}
+
 		if (providers.isEmpty()) {
 			return DefaultCredentialsProvider.create();
 		}
@@ -99,6 +110,20 @@ public class CredentialsProviderAutoConfiguration {
 			profileFile = ProfileFile.defaultProfileFile();
 		}
 		return ProfileCredentialsProvider.builder().profileName(profile.getName()).profileFile(profileFile).build();
+	}
+
+	private static boolean shouldCreateStsIdentityTokenCredentialsProvider() {
+		return ClassUtils.isPresent(STS_WEB_IDENTITY_TOKEN_FILE_CREDENTIALS_PROVIDER, null)
+			&& StringUtils.hasText(System.getenv(AWS_ROLE_ARN_ENV_VARIABLE))
+			&& StringUtils.hasText(System.getenv(AWS_WEB_IDENTITY_TOKEN_FILE_ENV_VARIABLE));
+	}
+
+	private static StsWebIdentityTokenFileCredentialsProvider createStsCredentialsProvider() {
+		return StsWebIdentityTokenFileCredentialsProvider.builder()
+			.stsClient(StsClient.builder().build())
+			.roleArn(System.getenv(AWS_ROLE_ARN_ENV_VARIABLE))
+			.webIdentityTokenFile(Path.of(System.getenv(AWS_WEB_IDENTITY_TOKEN_FILE_ENV_VARIABLE)))
+			.build();
 	}
 
 }
--- a/spring-cloud-aws-autoconfigure/pom.xml
+++ b/spring-cloud-aws-autoconfigure/pom.xml
@@ -146,6 +146,11 @@
 			<artifactId>spring-boot-starter-aop</artifactId>
 			<optional>true</optional>
 		</dependency>
+		<dependency>
+			<groupId>software.amazon.awssdk</groupId>
+			<artifactId>sts</artifactId>
+			<optional>true</optional>
+		</dependency>
 
 		<!-- AWS SDK v1 is required by testcontainers-localstack -->
 		<dependency>


maciejwalkowiak:
Also fixes #691

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/core/CredentialsProviderAutoConfiguration.java
```diff
@@ -101,4 +112,18 @@ private static ProfileCredentialsProvider createProfileCredentialProvider(Profil
 		return ProfileCredentialsProvider.builder().profileName(profile.getName()).profileFile(profileFile).build();
 	}
 
+	private static boolean shouldCreateStsIdentityTokenCredentialsProvider() {
+		return ClassUtils.isPresent(STS_WEB_IDENTITY_TOKEN_FILE_CREDENTIALS_PROVIDER, null)
+			&& StringUtils.hasText(System.getenv(AWS_ROLE_ARN_ENV_VARIABLE))
```
maciejwalkowiak, line 117: Do these properties make sense only as env variables or we should also include them in `CredentialsProperties`?

maciejwalkowiak (changes requested):
Thanks @eduanb for a PR! Could you please also update reference documentation?

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/core/CredentialsProviderAutoConfiguration.java
```diff
@@ -45,6 +49,9 @@
 @EnableConfigurationProperties(CredentialsProperties.class)
 public class CredentialsProviderAutoConfiguration {
 
+	public static final String STS_WEB_IDENTITY_TOKEN_FILE_CREDENTIALS_PROVIDER = "software.amazon.awssdk.services.sts.auth.StsWebIdentityTokenFileCredentialsProvider";
+	public static final String AWS_ROLE_ARN_ENV_VARIABLE = "AWS_ROLE_ARN";
```
MatejNedic, line 53: We should modify `CredentialsProperties` with nested `StsProperties` and move all of this there.

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/core/CredentialsProviderAutoConfiguration.java
```diff
@@ -101,4 +112,18 @@ private static ProfileCredentialsProvider createProfileCredentialProvider(Profil
...
+			&& StringUtils.hasText(System.getenv(AWS_ROLE_ARN_ENV_VARIABLE))
+			&& StringUtils.hasText(System.getenv(AWS_WEB_IDENTITY_TOKEN_FILE_ENV_VARIABLE));
+	}
+
+	private static StsWebIdentityTokenFileCredentialsProvider createStsCredentialsProvider() {
+		return StsWebIdentityTokenFileCredentialsProvider.builder()
+			.stsClient(StsClient.builder().build())
+			.roleArn(System.getenv(AWS_ROLE_ARN_ENV_VARIABLE))
+			.webIdentityTokenFile(Path.of(System.getenv(AWS_WEB_IDENTITY_TOKEN_FILE_ENV_VARIABLE)))
+			.build();
```
MatejNedic, line 126: We should add:
`.asyncCredentialUpdateEnabled(stsProperties.isAsyncCredentialsUpdate())`

So users have options to run sync or async calls.

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/core/CredentialsProviderAutoConfiguration.java
```diff
@@ -101,4 +112,18 @@ private static ProfileCredentialsProvider createProfileCredentialProvider(Profil
...
 
+	private static boolean shouldCreateStsIdentityTokenCredentialsProvider() {
+		return ClassUtils.isPresent(STS_WEB_IDENTITY_TOKEN_FILE_CREDENTIALS_PROVIDER, null)
+			&& StringUtils.hasText(System.getenv(AWS_ROLE_ARN_ENV_VARIABLE))
+			&& StringUtils.hasText(System.getenv(AWS_WEB_IDENTITY_TOKEN_FILE_ENV_VARIABLE));
+	}
+
+	private static StsWebIdentityTokenFileCredentialsProvider createStsCredentialsProvider() {
+		return StsWebIdentityTokenFileCredentialsProvider.builder()
+			.stsClient(StsClient.builder().build())
```
MatejNedic, line 123: Lets include `StsWebIdentityTokenFileCredentialsProvider.builder().roleSessionName()`

MatejNedic (changes requested):
Hey, @eduanb thanks for the PR!
Looks good so far. Could you apply the requested changes and updated docs, please?

eduanb:
Thank you for the feedback. I'm now confident that this will be a valid feature to add. 
I will work on the feedback, tests and documentation soon.

eduanb:
Hi @marcingrzejszczak and @MatejNedic, I've completely redone almost everything in this PR. I've added a test and documentation. 

I decided the `stsProperties` are big enough to be its own class. I've also made it use the standard `AWS_ROLE_ARN` so it could be configured with env variables or Spring Boot properties. 

I've added many debug logs and documented that the user can use them. Let me know if this is a good idea or should rather be removed.

MatejNedic:
Hey @eduanb , looks good, thank you so much for PR!
I did some small polish @maciejwalkowiak could you review it, please?

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/core/StsProperties.java
```diff
@@ -0,0 +1,105 @@
...
+ * @author Eduan Bekker
+ * @since 3.0.0
+ */
+public class StsProperties {
+
+	/**
+	 * Activate {@link software.amazon.awssdk.services.sts.auth.StsWebIdentityTokenFileCredentialsProvider}
+	 * AutoConfiguration for {@link software.amazon.awssdk.crt.auth.credentials.CredentialsProvider}.
+	 */
+	private boolean enabled = false;
```
MatejNedic, line 34: I would like to keep this since I am not a fan of autoconfiguring STS based only on if the class is on the classpath. There are multiple ways you can do STS and we should not force it.
This also helps if you want locally not to use STS, but once deployed to use it.

eduanb: I agree this should be configurable. I wonder if we should make the default `true` because you already need to add the dependency and it is common for Spring Boot to auto-configure things based on dependencies. Additionally, you need to set the role and have a credentials file so I don't think it will happen accidentally locally. I also like the idea of zero-config authentication in EKS. 

That said, a false default is also fine by me.

MatejNedic: I agree with all of this, but imagine someone using STS for one client and (accessKey, secretKey) for all other clients. This way he can just disable STS autoconfiguration.
Was thinking about this as well, I know it is a weird use case but I can see it happening.

docs/src/main/asciidoc/core.adoc
```diff
@@ -61,9 +76,45 @@ If it does not serve your project needs, this behavior can be changed by setting
...
+[source,xml,indent=0]
+----
+<dependency>
+	<groupId>software.amazon.awssdk</groupId>
+	<artifactId>sts</artifactId>
+</dependency>
+----
+
+===== Configuring
+In EKS no additional configuration is required as the service account already configures the correct environment variables; however, they can be overridden.
```
maciejwalkowiak, line 102: We should have here a table with available properties and i don't think we need to document system properties and env variables used by AWS SDK. Maybe we can link to their official docs.

maciejwalkowiak (changes requested):

docs/src/main/asciidoc/core.adoc
MatejNedic: Agree

MatejNedic: I think saying `(falls back to SDK default)`. What do you think @maciejwalkowiak

maciejwalkowiak (changes requested):

spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/core/CredentialsProviderAutoConfigurationTests.java
```diff
@@ -89,6 +96,42 @@ void credentialsProvider_profileNameAndPathConfigured_configuresProfileCredentia
...
+				.run((context) -> {
+					AwsCredentialsProvider awsCredentialsProvider = context.getBean("credentialsProvider",
+							AwsCredentialsProvider.class);
+					assertThat(awsCredentialsProvider).isNotNull()
+							.isInstanceOf(StsWebIdentityTokenFileCredentialsProvider.class);
+				});
+	}
+
+	@Test
+	void credentialsProvider_stsSystemPropertiesDefault_configuresStsWebIdentityTokenFileCredentialsProvider()
```
maciejwalkowiak, line 118: This test is failing, for a good reason - `spring.cloud.aws.credentials.sts.enabled` has been removed.

Look another comment

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/core/CredentialsProviderAutoConfiguration.java
```diff
@@ -73,6 +83,11 @@ public static AwsCredentialsProvider createCredentialsProvider(CredentialsProper
 			providers.add(createProfileCredentialProvider(profile));
 		}
 
+		StsProperties sts = properties.getSts();
+		if (ClassUtils.isPresent(STS_WEB_IDENTITY_TOKEN_FILE_CREDENTIALS_PROVIDER, null) && sts != null) {
```
maciejwalkowiak, line 87: Perhaps we should not check for `sts != null` and return `StsWebIdentityTokenFileCredentialsProvider` if sts is on the classpath.

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/core/CredentialsProviderAutoConfiguration.java
```diff
@@ -101,4 +116,20 @@ private static ProfileCredentialsProvider createProfileCredentialProvider(Profil
 		return ProfileCredentialsProvider.builder().profileName(profile.getName()).profileFile(profileFile).build();
 	}
 
+	/**
+	 * Wrapper class to avoid {@link NoClassDefFoundError}.
+	 */
+	private static class StsCredentialsProviderFactory {
+		private static AwsCredentialsProvider create(StsProperties stsProperties, AwsRegionProvider regionProvider) {
```
maciejwalkowiak, line 123: If the comment above makes sense, `stsProperties` here would be `@Nullable`

maciejwalkowiak:
@MatejNedic please take a look at comments above. If this make sense, i have a commit ready locally

MatejNedic:
@maciejwalkowiak I agree with all the above.

maciejwalkowiak:
@MatejNedic please take another look at the recent changes. If it matches what you had in mind we can merge

MatejNedic:
@maciejwalkowiak looks good to me

PR #726, opened 2023-03-22
Add option to set DynamoDB table prefix.

arunkpatra:
Add option to set DynamoDB table prefix.

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
A table name prefix can be configured to be used by the default `DynamoDbTableNameResolver`. If specified via the `spring.cloud.aws.dynamodb.table-prefix` property, the table prefix will be prepended to the resolved table name.

For example, if the table prefix is specified as `foo_` and the entity class is named as `Person`, then the resolved table name will be `foo_person`. Since this is an `opt-in` feature, not specifying any table prefix, will resolve the table name as `person` in this example.

- Since all DynamoDB tables live in a single namespace, there should be an option to set a prefix to tables.
- This is useful in teh scenario where multiple applications want to use the same _logical_ table name in DynamoDB hence needing a discriminator.

- Added unit tests and integration tests.
- Un-prefixed (no table prefix specified) configuration works 
- Prefixed (with table prefix) configuration works

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/spring-cloud-aws-dynamodb/src/main/java/io/awspring/cloud/dynamodb/DefaultDynamoDbTableNameResolver.java
+++ b/spring-cloud-aws-dynamodb/src/main/java/io/awspring/cloud/dynamodb/DefaultDynamoDbTableNameResolver.java
@@ -36,6 +36,10 @@ public class DefaultDynamoDbTableNameResolver implements DynamoDbTableNameResolv
 		this.tablePrefix = tablePrefix;
 	}
 
+	public DefaultDynamoDbTableNameResolver() {
+		this(null);
+	}
+
 	@Override
 	public String resolve(Class clazz) {
 		Assert.notNull(clazz, "clazz is required");
--- a/spring-cloud-aws-dynamodb/src/main/java/io/awspring/cloud/dynamodb/DefaultDynamoDbTableNameResolver.java
+++ b/spring-cloud-aws-dynamodb/src/main/java/io/awspring/cloud/dynamodb/DefaultDynamoDbTableNameResolver.java
@@ -16,7 +16,7 @@
 package io.awspring.cloud.dynamodb;
 
 import java.util.Locale;
-import org.springframework.core.env.Environment;
+import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
@@ -29,19 +29,18 @@ import org.springframework.util.StringUtils;
  */
 public class DefaultDynamoDbTableNameResolver implements DynamoDbTableNameResolver {
 
-	private final Environment environment;
+	@Nullable
+	private final String tablePrefix;
 
-	public DefaultDynamoDbTableNameResolver(Environment environment) {
-		this.environment = environment;
+	public DefaultDynamoDbTableNameResolver(@Nullable String tablePrefix) {
+		this.tablePrefix = tablePrefix;
 	}
 
 	@Override
 	public String resolve(Class clazz) {
 		Assert.notNull(clazz, "clazz is required");
 
-		String prefix = this.environment.getProperty("spring.cloud.aws.dynamodb.table-prefix");
-		String tablePrefix = StringUtils.hasText(prefix) ? prefix : "";
-
-		return tablePrefix.concat(clazz.getSimpleName().replaceAll("(.)(\\p{Lu})", "$1_$2").toLowerCase(Locale.ROOT));
+		String prefix = StringUtils.hasText(tablePrefix) ? tablePrefix : "";
+		return prefix.concat(clazz.getSimpleName().replaceAll("(.)(\\p{Lu})", "$1_$2").toLowerCase(Locale.ROOT));
 	}
 }
--- a/spring-cloud-aws-dynamodb/src/main/java/io/awspring/cloud/dynamodb/DefaultDynamoDbTableNameResolver.java
+++ b/spring-cloud-aws-dynamodb/src/main/java/io/awspring/cloud/dynamodb/DefaultDynamoDbTableNameResolver.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2022 the original author or authors.
+ * Copyright 2013-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,19 +16,32 @@
 package io.awspring.cloud.dynamodb;
 
 import java.util.Locale;
+import org.springframework.core.env.Environment;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 
 /**
  * Simple implementation of {@link DynamoDbTableNameResolver} that resolves class simple name to table name.
  *
  * @author Matej Nedic
+ * @author Arun Patra
  * @since 3.0
  */
 public class DefaultDynamoDbTableNameResolver implements DynamoDbTableNameResolver {
 
+	private final Environment environment;
+
+	public DefaultDynamoDbTableNameResolver(Environment environment) {
+		this.environment = environment;
+	}
+
 	@Override
 	public String resolve(Class clazz) {
 		Assert.notNull(clazz, "clazz is required");
-		return clazz.getSimpleName().replaceAll("(.)(\\p{Lu})", "$1_$2").toLowerCase(Locale.ROOT);
+
+		String prefix = this.environment.getProperty("spring.cloud.aws.dynamodb.table-prefix");
+		String tablePrefix = StringUtils.hasText(prefix) ? prefix : "";
+
+		return tablePrefix.concat(clazz.getSimpleName().replaceAll("(.)(\\p{Lu})", "$1_$2").toLowerCase(Locale.ROOT));
 	}
 }
--- a/spring-cloud-aws-dynamodb/src/main/java/io/awspring/cloud/dynamodb/DynamoDbTemplate.java
+++ b/spring-cloud-aws-dynamodb/src/main/java/io/awspring/cloud/dynamodb/DynamoDbTemplate.java
@@ -38,8 +38,8 @@ public class DynamoDbTemplate implements DynamoDbOperations {
 	private final DynamoDbTableNameResolver dynamoDbTableNameResolver;
 
 	public DynamoDbTemplate(Environment environment, DynamoDbEnhancedClient dynamoDbEnhancedClient) {
-		this(dynamoDbEnhancedClient, new DefaultDynamoDbTableSchemaResolver(),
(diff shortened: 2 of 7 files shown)


spring-cloud-aws-dynamodb/src/main/java/io/awspring/cloud/dynamodb/DefaultDynamoDbTableNameResolver.java
```diff
@@ -16,19 +16,32 @@
...
  *
  * @author Matej Nedic
+ * @author Arun Patra
  * @since 3.0
  */
 public class DefaultDynamoDbTableNameResolver implements DynamoDbTableNameResolver {
 
+	private final Environment environment;
+
+	public DefaultDynamoDbTableNameResolver(Environment environment) {
```
maciejwalkowiak, line 34: Shouldn't the `DynamoDbProperties` be injected here rather than `Environment`?

spring-cloud-aws-dynamodb/src/test/java/io/awspring/cloud/dynamodb/DynamoDbTemplateIntegrationTest.java
```diff
@@ -261,11 +458,36 @@ void dynamoDbTemplate_saveAndScanForParticularIndex_entitySuccessful() {
 		cleanUp(personEntity2.getUuid());
 	}
 
+	@Test
+	void prefixedDynamoDbTemplate_saveAndScanForParticularIndex_entitySuccessful() {
```
maciejwalkowiak, line 462: I am not sure if we need all these integration tests but assuming we do, lets create a nested class with JUnit 5 `@Nested` and move them all there

maciejwalkowiak (changes requested):
Thanks @arunkpatra! Some small changes are requested

spring-cloud-aws-dynamodb/src/main/java/io/awspring/cloud/dynamodb/DefaultDynamoDbTableNameResolver.java
arunkpatra: @maciejwalkowiak The `DynamoDbProperties` are in the `spring-cloud-aws-autoconfigure` module, and the dependency graph is pointing from `spring-cloud-aws-dynamodb` towards that. 
Actually if memory serves right, some time back we had been talking on the dependency graph where the motivation was.

1. Keep the lower level modules fairly de-coupled from Spring Boot per se, so that thay can be used by users who do not use Spring Boot and/or do not want to use auto-configuration.
2. We ended up following a uniform patterm where the properties are in the `spring-cloud-aws-autoconfigure` module with the intent that the properties are primarily meant to manage/customize auto-configuration.

So, I had to infer the properties via the `Environment`,

spring-cloud-aws-dynamodb/src/test/java/io/awspring/cloud/dynamodb/DynamoDbTemplateIntegrationTest.java
arunkpatra: @maciejwalkowiak Yeah, I kind of agree that having  duplicated test cases - one for the prefixed secenario and another for the un-prefixed scenario - is just leading us to un-maintainable test suites. I will wither `nest` them or do some parametrized test to ease the pain.

Thanks for bring this up.

spring-cloud-aws-dynamodb/src/main/java/io/awspring/cloud/dynamodb/DefaultDynamoDbTableNameResolver.java
arunkpatra: But `DynamoDefaultDbTableNameResolver` and `DynamoDbProperties` are in different modules and it would mean creating a cyclic dependency if we try injecting the properties. The only option left would be to pull out `DynamoDbProperties` from `spring-cloud-aws-autoconfigure` and move  it to `spring-cloud-aws-dynamodb`; this would be a departurture from the other modules. Secondly, relocating the `DynamoDbProperties` would be a breaking build time change for users of the RC release when we go GA.

Another thing is, `DynamoDefaultDbTableNameResolver` is a core construct necessary for the `DynamoDbTemplate` to work (whether you auto-configure or not). Hence, `DynamoDefaultDbTableNameResolver` has to be in the lower level module `spring-cloud-aws-dynamodb`.

Honestly I did not want to inject the `Environment` at all given the fact it sounds a bit off to do so, especially given the fact what we are using it for. But is there any other way in this case?

maciejwalkowiak: i mean to pass prefix as a string and use dynamodbproperties in the autoconfigure module

spring-cloud-aws-dynamodb/src/test/java/io/awspring/cloud/dynamodb/DynamoDbTemplateIntegrationTest.java
arunkpatra: @maciejwalkowiak This is done. I parametrized tests, so we have now test that cover both scenarios without any extra code; in a lighter note - "buy one get one free" :-)

spring-cloud-aws-dynamodb/src/main/java/io/awspring/cloud/dynamodb/DefaultDynamoDbTableNameResolver.java
arunkpatra: Aaah, I see what you mean. Done. I refactored `DynamoDefaultDbTableNameResolver` so that it does not need an `Environment` injection, instead it takes a `tablePrefix`. 

There's one special case though, which we will have to accomodate. A user wanting a  _raw_ `DynamoDbTemplate` by supplying their own `DynamoDbEnhancedClient` needs a `DefaultDynamoDbTableNameResolver` too. In that case, we would need to inject the `Environment` into the `DynamoDbTemplate` if we were to offer the benefits of the `tablePrefix`. But I guess, that should be acceptable. If we don't do this, then we will have to give the user a plain and un-decorated `DefaultDynamoDbTableNameResolver` which is not aware of the `tablePrefix`. This is just a fallout of our design and dependency graph, nothing else. IMO, this should be ok. What do you think?

spring-cloud-aws-dynamodb/src/main/java/io/awspring/cloud/dynamodb/DynamoDbTemplate.java
```diff
@@ -28,15 +29,17 @@
...
  * @since 3.0
  */
 public class DynamoDbTemplate implements DynamoDbOperations {
 	private final DynamoDbEnhancedClient dynamoDbEnhancedClient;
 	private final DynamoDbTableSchemaResolver dynamoDbTableSchemaResolver;
 	private final DynamoDbTableNameResolver dynamoDbTableNameResolver;
 
-	public DynamoDbTemplate(DynamoDbEnhancedClient dynamoDbEnhancedClient) {
-		this(dynamoDbEnhancedClient, new DefaultDynamoDbTableSchemaResolver(), new DefaultDynamoDbTableNameResolver());
+	public DynamoDbTemplate(Environment environment, DynamoDbEnhancedClient dynamoDbEnhancedClient) {
```
MatejNedic, line 40: Let's remove Environment from the constructor and add a String prefix which is nullable.
We should not be injecting Environment.

spring-cloud-aws-dynamodb/src/main/java/io/awspring/cloud/dynamodb/DynamoDbTemplate.java
```diff
@@ -28,15 +29,17 @@
...
+		this(dynamoDbEnhancedClient, new DefaultDynamoDbTableSchemaResolver(), new DefaultDynamoDbTableNameResolver(
+				environment.getProperty("spring.cloud.aws.dynamodb.table-prefix")));
```
MatejNedic, line 42: Should be taken from `DynamoDbProperties` in the best case or from the String prefix.

spring-cloud-aws-dynamodb/src/test/java/io/awspring/cloud/dynamodb/DynamoDbTemplateIntegrationTest.java
```diff
@@ -96,23 +123,27 @@ void dynamoDbTemplate_saveUpdateAndRead_entitySuccessful() {
 		assertThat(personEntity1).isEqualTo(personEntity);
 
 		// clean up
-		cleanUp(personEntity.getUuid());
+		cleanUp(dynamoDbTable, personEntity.getUuid());
 	}
 
-	@Test
-	void dynamoDbTemplate_saveAndDelete_entitySuccessful() {
+	@ParameterizedTest
```
MatejNedic, line 129: I really dislike `@ParameterizedTest`.
Would rather prefer `@Nested`
But I see code deduplication will be there so it is fine.

MatejNedic (changes requested):
Hey, @arunkpatra Thanks so much for MR.
I have left few comments that we should tackle

spring-cloud-aws-dynamodb/src/main/java/io/awspring/cloud/dynamodb/DynamoDbTemplate.java
arunkpatra: Thanks @MatejNedic. Since `DynamoDbProperties` is not accessible in this module because of teh dependency graph we have, so I have used `String` prefix

spring-cloud-aws-dynamodb/src/main/java/io/awspring/cloud/dynamodb/DynamoDbTemplate.java
arunkpatra: done

maciejwalkowiak (approved):

maciejwalkowiak:
Thanks once again @arunkpatra!

acollins4-chwy:
Thanks for preparing this @arunkpatra. However, your solution, while very thorough, is very narrowly focused on only making the `DynamoDbTemplate` more convenient to use.

Did you consider using something like this in the `DynamoDbAutoConfiguration`?
```
@ConditionalOnMissingBean
@Bean
public DynamoDbTableNameResolver dynamoDbTableNameResolver(DynamoDbProperties properties) {
  return new DefaultDynamoDbTableNameResolver(properties.getTablePrefix());
}
```

I feel like utilizing this would have optimized the number of auto-configuration opportunities.

Issue #658, opened 2023-02-03
Change SQS Integration Tests to use SqsTemplate


tomazfernandes:
When the integration tests for the `SQS` integration were created, `SqsTemplate` didn't exist.

Now it's necessary to change them so they use the template instead of the raw `SqsAsyncClient`.

rajadilipkolli:
I will pick this up, please assign to me

tomazfernandes:
Cool, thanks!

rajadilipkolli:
@tomazfernandes , Should I keep current tests as is and add newer ones with SqsTemplate or replace SqsAsyncClient?

tomazfernandes:
For this issue the idea is to stop using the raw client at all - if there’s any use case the Template doesn’t cover please let me know.

If there’s anything you’d like to write a new test for that should be ok too, but perhaps in a separate PR.

tomazfernandes:
@rajadilipkolli, just curious about how this is going. Any issues so far?

PR #727, opened 2023-03-22
Use SqsTemplate instead of raw SqsAsyncClient in SQS integration tests

MikhailStrokov:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description
For integration tests in SQS module, raw SqsAsyncClient client was substituted by SqsTemplate in methods for sending messages to queues.

Resolves awspring/spring-cloud-aws#658

When the integration tests for the SQS integration were created, SqsTemplate didn't exist.
Now it's necessary to change them so they use the template instead of the raw SqsAsyncClient.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration/SqsFifoIntegrationTests.java
+++ b/spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration/SqsFifoIntegrationTests.java
@@ -37,6 +37,8 @@ import io.awspring.cloud.sqs.listener.acknowledgement.AcknowledgementOrdering;
 import io.awspring.cloud.sqs.listener.acknowledgement.AcknowledgementResultCallback;
 import io.awspring.cloud.sqs.listener.acknowledgement.handler.AcknowledgementHandler;
 import io.awspring.cloud.sqs.listener.acknowledgement.handler.OnSuccessAcknowledgementHandler;
+import io.awspring.cloud.sqs.operations.SendResult;
+import io.awspring.cloud.sqs.operations.SqsTemplate;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -48,7 +50,6 @@ import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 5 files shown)


tomazfernandes (changes requested):
Thanks a lot for the PR @MikhailStrokov. Those `sendMessageTo` and `sendMessageBatch` were created to add functionality to the raw `SqsClient`, but shouldn't be necessary for `SqsTemplate`. Think you can replace those with equivalent from `SqsTemplate` and see if all tests remain green?

tomazfernandes:
> Hi @tomazfernandes, I have doubts about `io.awspring.cloud.sqs.integration.SqsFifoIntegrationTests#sendMessageTo`, so I want to clarify first. This method has logic based on `useLocalStackClient` flag. 

Yeah, this is because `LocalStack` implementation for `FIFO` queues is not reliable, so it needs to be sync to keep ordering. On the other hand some tests are also used to stress test against AWS, and for those cases async is better.

I think we can move the `if` statement on the `useLocalStackClient` flag to the `receivesMessagesInOrderFromManyMessageGroups` test, which is the one that cares about this the most, and just use the Template's blocking send methods on the other ones to help them pass on LocalStack.

We'd also need to have a `high throughput template` for that test in case we're targeting AWS.

How does that sound?

> Also, `sendMessageTo` and `sendMessageBatch` methods were removed from the others test classes.

Cool, thanks!

MikhailStrokov:
Hi @tomazfernandes, I've changed sending type to sync except `receivesMessagesInOrderFromManyMessageGroups`. I tried to get rid of `io.awspring.cloud.sqs.integration.SqsFifoIntegrationTests#doSendMessageTo` but it looked like a mess when I merged it into `io.awspring.cloud.sqs.integration.SqsFifoIntegrationTests#sendManyAsyncTo` method. So I left it as it is. Let me know what I can change to make it more clear or improve readability.

tomazfernandes:
> Hi @tomazfernandes, I've changed sending type to sync except `receivesMessagesInOrderFromManyMessageGroups`. I tried to get rid of `io.awspring.cloud.sqs.integration.SqsFifoIntegrationTests#doSendMessageTo` but it looked like a mess when I merged it into `io.awspring.cloud.sqs.integration.SqsFifoIntegrationTests#sendManyAsyncTo` method. So I left it as it is. Let me know what I can change to make it more clear or improve readability.

I think we should not need any of these `sendMessageTo` auxiliary methods, and should be able to rely on the `SqsTemplate` API directly instead in the test methods.

The one thing I'm aware that the API doesn't offer much help yet is to create multiple messages for `sendMany` methods. For these we should be able to maybe just use a `stream#map` to create the messages in the test method itself, or maybe have a method that takes the parameters and return `List<Message<>>`.

How does that sound?

Thanks

tomazfernandes:
@MikhailStrokov, thanks a lot for the PR!

We're releasing 3.0 GA soon, so I've added one commit with what was missing and since I couldn't push to your branch I opened a new PR.

I added your name to the test classes you changed, hope it's ok.

Thanks again!

Closed in favor of https://github.com/awspring/spring-cloud-aws/pull/770

Issue #670, opened 2023-02-12
Add More Samples to SQS Integration


tomazfernandes:
We currently have only one [simple sample](https://github.com/awspring/spring-cloud-aws/blob/main/spring-cloud-aws-samples/spring-cloud-aws-sqs-sample/src/main/java/io/awspring/cloud/sqs/sample/SpringSqsSample.java) for the `SQS Integration`.

It would be nice to have more sample examples for more complex use-cases, such as:
- Manual container instantiation and message consumption
- Batch sending and receiving messages
- Manual acknowledgement
- Receive from multiple queues

And a lot more. Not all use cases must be contemplated in a single PR, a PR with one use case is already great.

The project's [Integration Tests](https://github.com/awspring/spring-cloud-aws/tree/main/spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration) can be used as inspiration, and also the [integration's documentation](https://docs.awspring.io/spring-cloud-aws/docs/3.0.0-RC1/reference/html/index.html#sqs-integration).

Messages should be sent with `SqsTemplate`.

Please let us know of any issues or difficulties.

mcruzdev:
No @tomazfernandes, can you assign for me?

tomazfernandes:
Cool @mcruzdev, let us know if you run into any issues! Thanks

mcruzdev:
Hi @tomazfernandes, sorry for delay :) 

I added a sample where a listen receives messages from multiple queues.

mcruzdev:
The pull request for manua ack is here #774

dottnul:
Hi @tomazfernandes 

I add a sample for task executor https://github.com/awspring/spring-cloud-aws/pull/839 

thanks :)

tomazfernandes:
I think we're good for now, we may reopen the issue if necessary.

Thank you all for the contributions!

PR #742, opened 2023-03-28
Docs: Add new SQS sample (listen from multiple queues)

mcruzdev:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description

This pull request adds a sample for SQS where is shows a listener receiving messages from multiple queues.

The motivation is described in issue #670.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/spring-cloud-aws-samples/spring-cloud-aws-sqs-sample/src/main/java/io/awspring/cloud/sqs/sample/SpringSqsListenMultipleQueues.java
+++ b/spring-cloud-aws-samples/spring-cloud-aws-sqs-sample/src/main/java/io/awspring/cloud/sqs/sample/SpringSqsListenMultipleQueues.java
@@ -20,7 +20,6 @@ import io.awspring.cloud.sqs.operations.SqsTemplate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.boot.ApplicationRunner;
-import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.context.annotation.Bean;
 import software.amazon.awssdk.services.sqs.model.Message;
@@ -34,10 +33,6 @@ public class SpringSqsListenMultipleQueues {
 
 	private static final String ORDER_QUEUE = "order-queue";
 	private static final String WITHDRAWAL_QUEUE = "withdrawal-queue";
-
-	public static void main(String[] args) {
-		SpringApplication.run(SpringSqsListenMultipleQueues.class, args);
-	}
 
 	@SqsListener(queueNames = { ORDER_QUEUE, WITHDRAWAL_QUEUE })
 	void listen(Message message) {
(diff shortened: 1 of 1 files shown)


spring-cloud-aws-samples/spring-cloud-aws-sqs-sample/src/main/java/io/awspring/cloud/sqs/sample/SpringSqsListenMultipleQueues.java
```diff
@@ -0,0 +1,63 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.boot.ApplicationRunner;
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.annotation.Bean;
+import software.amazon.awssdk.services.sqs.model.Message;
+
+import java.util.UUID;
+
+@SpringBootApplication
```
maciejwalkowiak, line 30: Having multiple main classes breaks the build. Can we have this sample working as a part of the other main class?

maciejwalkowiak (changes requested):

spring-cloud-aws-samples/spring-cloud-aws-sqs-sample/src/main/java/io/awspring/cloud/sqs/sample/SpringSqsListenMultipleQueues.java
mcruzdev: Hi @maciejwalkowiak, thank you for you review.

I think that is not necessary to add `@Order` for ApplicationRunner bean.

> Interface used to indicate that a bean should run when it is contained within a [SpringApplication](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/SpringApplication.html). Multiple [ApplicationRunner](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/ApplicationRunner.html) beans can be defined within the same application context and can be ordered using the [Ordered](https://docs.spring.io/spring-framework/docs/6.0.7/javadoc-api/org/springframework/core/Ordered.html) interface or [@Order](https://docs.spring.io/spring-framework/docs/6.0.7/javadoc-api/org/springframework/core/annotation/Order.html) annotation.

What do you think?

maciejwalkowiak: Only if its relevant for the sample. If sample can work without `@Order` then its better without.

Once you removed `@SpringBootApplication` i think this class is not picked at all right? It must have `@Configuration`

tomazfernandes:
@mcruzdev sorry for the delay and thanks for the PR!

Looks good to me, if you can just add yourself as the author.

@maciejwalkowiak other than the author tag I've nothing else to add, feel free to merge when you're ok with it.

Issue #745, opened 2023-03-30
SQS message attributes mapping fails when using custom data type


npoirier:
**Type**: Bug

**Component**:
SQS message attribute mapping

**Describe the bug**

Mapping of SQS message attributes fails when using a  type other than Number with a custom data type.  ex: `String.Json` or `Binary.Protobuf`.

In `SqsHeaderMapper.getMessageAttributesAsHeaders()` it defaults to Number and then fails to parse the value:

```
	private Object getValue(Map.Entry<String, MessageAttributeValue> entry) {
		MessageAttributeValue value = entry.getValue();
		String dataType = value.dataType();
		Assert.notNull(dataType, "dataType must not be null");
		return MessageAttributeDataTypes.STRING.equals(dataType)
			? value.stringValue()
			: MessageAttributeDataTypes.BINARY.equals(dataType)
				? value.binaryValue()
				: getNumberValue(value);
	}
```

Custom data types : https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-metadata.html#sqs-message-attributes

tomazfernandes:
Fixed by #746

PR #746, opened 2023-03-30
Fix: mapping SQS message attributes with custom type

npoirier:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Fixes an error trying to parse a message attribute value as a Number when its data type is String or Binary with a custom data type.

Changes:
* Ignore custom type (`.custom-data-type`) when determining data type.
* Default to String for unknown data types instead of Number.

Fixes issue #745

Unit test

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SqsHeaderMapper.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SqsHeaderMapper.java
@@ -181,11 +181,16 @@ public class SqsHeaderMapper implements ContextAwareHeaderMapper<Message> {
 		MessageAttributeValue value = entry.getValue();
 		String dataType = value.dataType();
 		Assert.notNull(dataType, "dataType must not be null");
-		return MessageAttributeDataTypes.STRING.equals(dataType)
-			? value.stringValue()
-			: MessageAttributeDataTypes.BINARY.equals(dataType)
-				? value.binaryValue()
-				: getNumberValue(value);
+
+		if (dataType.contains(".")) {
+			dataType = dataType.substring(0, dataType.indexOf('.'));
+		}
+
+		return switch (dataType) {
+			case MessageAttributeDataTypes.NUMBER -> getNumberValue(value);
+			case MessageAttributeDataTypes.BINARY -> value.binaryValue();
+			default -> value.stringValue();
+		};
 	}
 
 	private Map<String, String> getMessageSystemAttributesAsHeaders(Message source) {
--- a/spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/support/converter/SqsHeaderMapperTests.java
+++ b/spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/support/converter/SqsHeaderMapperTests.java
@@ -58,9 +58,39 @@ class SqsHeaderMapperTests {
 		String headerName = "stringAttribute";
 		String headerValue = "myString";
 		Message message = Message.builder().body("payload")
+			.messageAttributes(
+				Map.of(headerName,
+					MessageAttributeValue.builder().dataType(MessageAttributeDataTypes.STRING)
+						.stringValue(headerValue).build()))
+			.messageId(UUID.randomUUID().toString()).build();
+		MessageHeaders headers = mapper.toHeaders(message);
+		assertThat(headers.get(headerName)).isEqualTo(headerValue);
+	}
+
+	@Test
+	void shouldAddStringCustomMessageAttributes() {
+		SqsHeaderMapper mapper = new SqsHeaderMapper();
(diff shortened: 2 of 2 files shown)


spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SqsHeaderMapper.java
```diff
@@ -181,11 +181,21 @@ private Object getValue(Map.Entry<String, MessageAttributeValue> entry) {
...
-			? value.stringValue()
-			: MessageAttributeDataTypes.BINARY.equals(dataType)
-				? value.binaryValue()
-				: getNumberValue(value);
+
+		if (dataType.contains(".")) {
+			dataType = dataType.substring(0, dataType.indexOf('.'));
+		}
+
+		switch (dataType) {
```
tomazfernandes, line 189: Given we're in JDK 17, IntelliJ suggested this instead:
```java
		return switch (dataType) {
			case MessageAttributeDataTypes.NUMBER -> getNumberValue(value);
			case MessageAttributeDataTypes.BINARY -> value.binaryValue();
			default -> value.stringValue();
		};
```

Had no idea this was possible, but sure looks good. WDYT?

tomazfernandes (commented):
Thanks a lot @npoirier for reporting and for the fix!

Do you think it might be worth it storing the custom type information (after the `.`) somehow? Maybe not for this PR, but we could open an issue.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SqsHeaderMapper.java
npoirier: Updated to use the new switch statements.

npoirier:
> Thanks a lot @npoirier for reporting and for the fix!
> 
> Do you think it might be worth it storing the custom type information (after the `.`) somehow? Maybe not for this PR, but we could open an issue.

That's an interesting idea, the custom type information sure could be useful.  But I'm not sure how it could be stored, as spring-messaging headers are key-value pairs.

tomazfernandes:
> That's an interesting idea, the custom type information sure could be useful. But I'm not sure how it could be stored, as spring-messaging headers are key-value pairs.

I'm thinking since the `Header` value contract is `Object`, we could create something like a `SqsCustomDataType` record that we could return with all necessary information along with the header value. Since users already need to `cast` the header value to use it and in a way be aware of the expected type, I think casting should not be a problem.

A more elaborate idea would be to allow a configurable `mapper` function which users could set to map the `message attribute` name and value to any specific type they prefer.

We could also combine both approaches, and have the the `mapper` function default to populate and return a `SqsCustomDataType`.

Makes sense?

npoirier:
> > That's an interesting idea, the custom type information sure could be useful. But I'm not sure how it could be stored, as spring-messaging headers are key-value pairs.
> 
> I'm thinking since the `Header` value contract is `Object`, we could create something like a `SqsCustomDataType` record that we could return with all necessary information along with the header value. Since users already need to `cast` the header value to use it and in a way be aware of the expected type, I think casting should not be a problem.
> 
> A more elaborate idea would be to allow a configurable `mapper` function which users could set to map the `message attribute` name and value to any specific type they prefer.
> 
> We could also combine both approaches, and have the the `mapper` function default to populate and return a `SqsCustomDataType`.
> 
> Makes sense?

I'm not sold on returning a `SqsCustomDataType` encapsulating the value and the data type.  I think the header value should be just that.

I like the idea of registering mappers for custom datatypes.  If no mapper is registered for the custom data type, then the value is returned as String, Binary or Number as specified by its "primary" data type.  If a mapper is registered for the custom data type, use that to map the value.

Some mappers could be provided for types like `String.Array`

tomazfernandes:
> Is anything missing in this PR to get it approved?

Sorry @npoirier, busy week here.

I'm wondering if we should try to get this mapping enhancement in time for 3.0, or if it's something we could add later for 3.1. If it's for 3.1 we can merge this PR and open an issue for the enhancement.

> I like the idea of registering mappers for custom datatypes. If no mapper is registered for the custom data type, then the value is returned as String, Binary or Number as specified by its "primary" data type. If a mapper is registered for the custom data type, use that to map the value.

I think I'd rather have a simple function taking a `MessageAttributeName` and `MessageAttributeValue` and outputting an `Object`, than having many mappings. Users could set the mapping with any logic they want, and If it returns `null` we could apply the default mapping not considering custom types.

But perhaps a `Map<dataType, mapperFunction>` should work too.

Let me know your thoughts. Thanks.

tomazfernandes:
@npoirier, thanks for the PR!

Would you like to create an issue for the mapping enhancement, and if so, would you also like to contribute a PR for that?

PR #759, opened 2023-04-06
Move testcontainers dependency

eddumelendez:
Currently, `org.testcontainers:testcontainers-bom` is part of the
spring-cloud-aws-dependencies which is imported in a consumer
project when the dependency is declared. The version provided by that
module is forced and can not be overriden. For that reason, this
PR suggest to move the dependency from that module.

Probably, another dependencies should be revisited.


maciejwalkowiak:
Good point @eddumelendez! Thanks for the PR!

maciejwalkowiak (approved):

Issue #693, opened 2023-02-27
Allow to set the phase for MessageListenerContainer related to the SmartLifecycle interface on 3.0.0+


estigma88:
**Type**: Feature

**Is your feature request related to a problem? Please describe.**
With previous versions of Spring Could AWS (No Spring Boot 3), we had troubles with the graceful shutdown process in Spring Boot. We have an application with the following components:
- SqsListener
- Quartz Scheduler

Seems like when the application is shutting down, SQS components and Quartz Scheduler components, have the same `SmartLifecycle.phase`, and they collide, causing Quartz to not shutting down completely, and letting some crons in the database corrupted.

The workaround we found was the following:

```
    @Bean
    fun beanXXXX(
       simpleMessageListenerContainer: SimpleMessageListenerContainer
    ): String {
        /**
         * SQS and Quartz seem having the same shutdown priority,
         * which causes problems with Quartz and cannot shutdown fine.
         *
         * phase = Integer.MAX_VALUE - 1 tells SQS to shutdown before Quartz.
         */
        simpleMessageListenerContainer.phase = Integer.MAX_VALUE - 1
        return "ok"
    }
```
Moving SQS components one phase before, guarantee SQS shutdowns first, and later, Quartz.

**Describe the solution you'd like**
A way to set the phase for the SmartLifecycle interface for 3.0.0+

**Describe alternatives you've considered**
- Testing again the collision with the latest version and see if the cron problem still appears.
- Going the rabbit hole and trying to find what's happening under the hood with this collision.

**Additional context**
I will try to create a minimal project to reproduce the issue, but it is environment dependent, not sure I can make it.

tomazfernandes:
Hi @estigma88, sorry for the delay - I wanted 3.0.0 to be out before making many more modifications to the SQS code.

Would you like to contribute a PR for this feature?

estigma88:
Sure, I can give it a try. Let me come back with one MR for this.

PR #821, opened 2023-05-18
Ability to set SmartLifecycle.phase to SqsMessageListenerContainer/DefaultListenerContainerRegistry

estigma88:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

- Adding the ability to set the `SmartLifecycle.phase` on `SqsMessageListenerContainer` using the builder
- Adding the ability to set the `SmartLifecycle.phase` on `AbstractMessageListenerContainer` using the `setPhase` method
- Adding the ability to set the `SmartLifecycle.phase` on `DefaultListenerContainerRegistry` using the `setPhase` method
- Update unit tests on `SqsMessageListenerContainerTests`,  `AbstractMessageListenerContainerTests` and `DefaultListenerContainerRegistryTests`

Details on https://github.com/awspring/spring-cloud-aws/issues/693

- Update unit tests on `SqsMessageListenerContainerTests` and `AbstractMessageListenerContainerTests`

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -955,6 +955,8 @@ MessageListenerContainer<Object> listenerContainer(SqsAsyncClient sqsAsyncClient
 }
 ----
 
+NOTE: You can also specify the `SmartLifecycle.phase`, using `SqsMessageListenerContainer.builder()`, if you are interested on overriding the default value defined in `MessageListenerContainer.DEFAULT_PHASE`
+
 ===== Retrieving Containers from the Registry
 
 Containers can be retrieved by fetching the `MessageListenerContainer` bean from the container and using the `getListenerContainers` and `getContainerById` methods.
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -927,7 +927,7 @@ The `MessageListenerContainer` interface extends `SmartLifecycle`, which provide
 Containers created from `@SqsListener` annotations are registered in a `MessageListenerContainerRegistry` bean that is registered by the framework.
 The containers themselves are not Spring-managed beans, and the registry is responsible for managing these containers` lifecycle in application startup and shutdown.
 
-NOTE: The framework offers the `DefaultListenerContainerRegistry` implementation, where you can override the `SmartLifecycle.phase`. By default, the `phase` value is `MessageListenerContainer.DEFAULT_PHASE`
+NOTE: The `DefaultListenerContainerRegistry ` implementation provided by the framework allows the phase value to be set through the `setPhase` method. The default value is `MessageListenerContainer.DEFAULT_PHASE`.
 
 At startup, the containers will make requests to `SQS` to retrieve the queues` urls for the provided queue names or ARNs, and for retrieving `QueueAttributes` if so configured.
 Providing queue urls instead of names and not requesting queue attributes can result in slightly better startup times since there's no need for such requests.
@@ -957,7 +957,7 @@ MessageListenerContainer<Object> listenerContainer(SqsAsyncClient sqsAsyncClient
 }
 ----
 
-NOTE: You can also specify the `SmartLifecycle.phase`, using `SqsMessageListenerContainer.builder()`, if you are interested on overriding the default value defined in `MessageListenerContainer.DEFAULT_PHASE`
+NOTE: The `SqsMessageListenerContainer.builder()` allows to specify the `SmartLifecycle.phase`, to override the default value defined in `MessageListenerContainer.DEFAULT_PHASE`
 
 ===== Retrieving Containers from the Registry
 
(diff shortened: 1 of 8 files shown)


spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/AbstractMessageListenerContainer.java
```diff
@@ -69,6 +69,8 @@
 	private AsyncAcknowledgementResultCallback<T> acknowledgementResultCallback = new AsyncAcknowledgementResultCallback<T>() {
 	};
 
+	private int phase = DEFAULT_PHASE;
```
tomazfernandes, line 72: Looking at Spring Kafka I see they have this:
`public static final int DEFAULT_PHASE = Integer.MAX_VALUE - 100; // late phase`

Likely to leave room for users' custom phased components to be phased before and after the containers, so I think we should use it here as well?

tomazfernandes (changes requested):
Looks great @estigma88 , thanks! Just one minor comment.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/AbstractMessageListenerContainer.java
estigma88: Makes sense, but this might be a breaking change? People might depend on the current phase of their application components start/stop lifecyle, it might be uncommon though

tomazfernandes: Yup, you're right, it'd be a breaking change.

I'm ok with it as is then, but maybe @maciejwalkowiak would like to weigh in?

tomazfernandes (approved):
Thanks a lot for the PR @estigma88!

Looking forward to more, there are quite a few open SQS issues, if you see any that you'd like to contribute just let us know.

maciejwalkowiak:
Perhaps we should update reference docs with an explanation?

tomazfernandes:
> Perhaps we should update reference docs with an explanation?

Good call @maciejwalkowiak.

@estigma88, would you mind adding a few lines to the doc regarding phase? Let me know if you need any help with that.

Thanks!

estigma88:
Sure @maciejwalkowiak @tomazfernandes , I will add some docs.

Btw, I was checking a little deep over the sqs framework, and I found that `DefaultListenerContainerRegistry` also implements `SmartLifeCycle`, seems like there are different ways to initialize listeners in the framework, should I add the ability to set the phase there also? I think thr PR is incomplete

tomazfernandes:
@estigma88 sorry, I’m in the middle of lots of things here and couldn’t reply before.

It makes sense to update the DefaultListenerContainerRegistry, good catch. Please add it to the PR if you can.

Thanks.

estigma88:
I added some docs, let me know if they make sense @tomazfernandes, also, I did a minor refactoring to extract the DEFAULT_PHASE somewhere else, so, we can modify the default in the future easily.

Regarding `DefaultListenerContainerRegistry`, I will create a new PR adding the phase to it, and we can discuss whether it makes sense.

estigma88:
@tomazfernandes don't worry.

I updated the docs, let me know if that makes sense, also, did a little refactor to reuse the `DEFAULT_PHASE` in the `DefaultListenerContainerRegistry` and being able to change it easily in the future.

docs/src/main/asciidoc/sqs.adoc
```diff
@@ -955,6 +957,8 @@ MessageListenerContainer<Object> listenerContainer(SqsAsyncClient sqsAsyncClient
 }
 ----
 
+NOTE: You can also specify the `SmartLifecycle.phase`, using `SqsMessageListenerContainer.builder()`, if you are interested on overriding the default value defined in `MessageListenerContainer.DEFAULT_PHASE`
```
tomazfernandes, line 960: Same comment here about using `you`. Please try to rewrite using the passive voice.

tomazfernandes (changes requested):
@estigma88 , just two comments about the documentation changes, please let me know if you have any questions.

Thanks.

docs/src/main/asciidoc/sqs.adoc
```diff
@@ -927,6 +927,8 @@ The `MessageListenerContainer` interface extends `SmartLifecycle`, which provide
 Containers created from `@SqsListener` annotations are registered in a `MessageListenerContainerRegistry` bean that is registered by the framework.
 The containers themselves are not Spring-managed beans, and the registry is responsible for managing these containers` lifecycle in application startup and shutdown.
 
+NOTE: The framework offers the `DefaultListenerContainerRegistry` implementation, where you can override the `SmartLifecycle.phase`. By default, the `phase` value is `MessageListenerContainer.DEFAULT_PHASE`
```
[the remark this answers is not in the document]
estigma88: Done

docs/src/main/asciidoc/sqs.adoc
estigma88: Done

tomazfernandes (approved):
Looks great, thanks @estigma88! Looking forward to more!

Issue #761, opened 2023-04-07
Add Acknowledgement Mode to @SqsListener Annotation


oleksandr-skoryi:
**Type**: Feature

**Is your feature request related to a problem? Please describe.**
I am using spring boot 3, the V2 SqsListener simply doesn't work, I have to migrate to V3 SqsListener
The V2 lib has the following declaration of the SqsListener
```
  @SqsListener(
      value = "${queue.name}",
      deletionPolicy = SqsMessageDeletionPolicy.ON_SUCCESS)
```
The same annotation in V3 library doesn't have deletionPolicy
**Describe the solution you'd like**
I would like to have it there, or having alternative.

**Describe alternatives you've considered**
Currently do @Scheduled annotation and manual polling

maciejwalkowiak:
Message deletion in v3 is done through "acknowledging": https://docs.awspring.io/spring-cloud-aws/docs/3.0.0-SNAPSHOT/reference/html/index.html#acknowledging-messages

Perhaps we can bring the acknowledge option to the annotation itself? @tomazfernandes

oleksandr-skoryi:
> https://docs.awspring.io/spring-cloud-aws/docs/3.0.0-SNAPSHOT/reference/html/index.html#acknowledging-messages

But it is a global configuration, what about making it more granular, per SqsListener?

tomazfernandes:
> But it is a global configuration, isn't it? What about making it more granular, per SqsListener?

You can configure different acknowledgement strategies for different factories, and specify the factory you want to use in the `@SqsListener` annotation.

> Perhaps we can bring the acknowledge option to the annotation itself? @tomazfernandes

Yeah, we can look into that. I may be wrong, but I believe the main reason this was in the annotation in 2.x is that given you could not configure separate factories for each annotation any customization had to be in the annotation itself.

tomazfernandes:
I guess you edited your comment about the same time I was writing mine.

> And for the 99% of cases, dev doesn't care about other options except acknowledgment type

3.0 has a lot of other configurations available that might be interesting. For example, for acknowledgement you can define whether to batch acks or not, the thresholds to acknowledge a batch, etc. But yeah, maybe that's a more advanced topic that most users might not look into.

> I think the V2 lib users are also got used to it.

I think this is a good point.

iparadiso:
[...]

PR #870, opened 2023-08-23
Add acknowledgement mode on @SqsListener annotation

jvcalassio:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
This PR adds a parameter on the `@SqsListener` annotation that configures the acknowledgement mode for the created container. The default behavior is the same as it's now: uses the acknowledgement mode defined on the container factory. However, once the parameter is set, the factory value is overwritten in favor of the value defined on the annotation.

This feature was requested on issue #761. By having this feature, users coming from the 2.x.x versions will have a easier time migrating to 3.x.x, given that this parameter was used to configure the acknowledgement mode on the previous major version of the library.

1. Integration tests
2. Sample application with new version of the library

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

### DIFF
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -636,6 +636,8 @@ Note that if there are messages available the call may return earlier than this 
 - `messageVisibilitySeconds` - Set the minimum visibility for the messages retrieved in a poll.
 Note that for `FIFO` single message listener methods, this visibility is applied to the whole batch before each message is sent to the listener.
 See <<FIFO Support>> for more information.
+- `acknowledgementMode` - Set the acknowledgement mode for the container. Default value is `SqsListenerAcknowledgementMode.INHERIT`,
+meaning that it will inherit the acknowledgement mode of the container factory. See <<Acknowledgement Mode>> for more information.
 
 
 ===== Listener Method Arguments
@@ -1325,6 +1327,8 @@ NOTE: All options are available for both `single message` and `batch` message li
 - `ON_SUCCESS` - Acknowledges a message or batch of messages after successful processing.
 - `ALWAYS` - Acknowledges a message or batch of messages after processing returns success or error.
 - `MANUAL` - The framework won't acknowledge messages automatically and `Acknowledgement` objects can be received in the listener method.
+
+The `Acknowledgement` strategy can be configured in the `SqsContainerOptions` or in the `@SqsListener` annotation.
 
 ==== Acknowledgement Batching
 
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -636,8 +636,8 @@ Note that if there are messages available the call may return earlier than this 
 - `messageVisibilitySeconds` - Set the minimum visibility for the messages retrieved in a poll.
 Note that for `FIFO` single message listener methods, this visibility is applied to the whole batch before each message is sent to the listener.
 See <<FIFO Support>> for more information.
-- `acknowledgementMode` - Set the acknowledgement mode for the container. Default value is `SqsListenerAcknowledgementMode.INHERIT`,
-meaning that it will inherit the acknowledgement mode of the container factory. See <<Acknowledgement Mode>> for more information.
+- `acknowledgementMode` - Set the acknowledgement mode for the container. Default value is `SqsListenerAcknowledgementMode.DEFAULT`,
+meaning that it will use the acknowledgement mode defined for the container factory. See <<Acknowledgement Mode>> for more information.
 
 
 ===== Listener Method Arguments
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -636,9 +636,9 @@ Note that if there are messages available the call may return earlier than this 
 - `messageVisibilitySeconds` - Set the minimum visibility for the messages retrieved in a poll.
 Note that for `FIFO` single message listener methods, this visibility is applied to the whole batch before each message is sent to the listener.
 See <<FIFO Support>> for more information.
-- `acknowledgementMode` - Set the acknowledgement mode for the container. Default value is `SqsListenerAcknowledgementMode.DEFAULT`,
-meaning that it will use the acknowledgement mode defined for the container factory. See <<Acknowledgement Mode>> for more information.
-
+- `acknowledgementMode` - Set the acknowledgement mode for the container.
+If any value is set, it will take precedence over the acknowledgement mode defined for the container factory options.
+See <<Acknowledgement Mode>> for more information.
 
 ===== Listener Method Arguments
 
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/SqsListener.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/SqsListener.java
@@ -138,8 +138,8 @@ public @interface SqsListener {
 	String messageVisibilitySeconds() default "";
 
 	/**
-	 * The acknowledgement mode to be used for the provided queues. If not specified, the acknowledgement mode defined
-	 * for the container factory will be used.
+	 * The acknowledgement mode to be used for the provided queues. If not specified, the
+	 * acknowledgement mode defined for the container factory will be used.
 	 */
 	SqsListenerAcknowledgmentMode acknowledgmentMode() default SqsListenerAcknowledgmentMode.INHERIT;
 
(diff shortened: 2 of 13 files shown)


spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/SqsListenerAcknowledgementMode.java
```diff
@@ -0,0 +1,57 @@
...
+ * @see NeverAcknowledgementHandler
+ * @see io.awspring.cloud.sqs.listener.ContainerOptions
+ * @see SqsListener
+ */
+public enum SqsListenerAcknowledgementMode {
+
+	/**
+	 * Use acknowledge mode defined by the container.
+	 */
+	INHERIT,
```
tomazfernandes, line 38: Not sure about `INHERIT`, how about we call it `DEFAULT`?

Another thing is - enums don't really play nice with annotations because we can't use SPeL for them.

I'm thinking if it would make sense to have the `AcknowledgementMode` as a `String` and provide users with a Constant class.

WDYT?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/SqsContainerOptionsBuilder.java
```diff
@@ -65,4 +66,11 @@ SqsContainerOptionsBuilder messageSystemAttributeNames(
 	 */
 	SqsContainerOptionsBuilder queueNotFoundStrategy(QueueNotFoundStrategy queueNotFoundStrategy);
 
+	/**
+	 * Set the {@link AcknowledgementMode} for the container.
+	 * @param acknowledgementMode the acknowledgement mode.
+	 * @return this instance.
+	 */
+	SqsContainerOptionsBuilder acknowledgementMode(AcknowledgementMode acknowledgementMode);
```
tomazfernandes, line 74: Not sure why we need this - we already have `acknowledgeMode` in the `AbstractContainerOptions` class.

spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration/SqsIntegrationTests.java
```diff
@@ -231,6 +251,52 @@ void doesNotAckOnErrorBatchAsync() throws Exception {
 		assertThat(latchContainer.doesNotAckBatchAsyncLatch.await(10, TimeUnit.SECONDS)).isTrue();
 	}
 
+	@Test
```
tomazfernandes, line 254: I think this test suite is already too big for us to introduce this many tests.

Would you mind moving these tests to their own suite?

tomazfernandes (commented):
Thanks for the PR @jvcalassio, looks good.

Left a couple of comments, please let me know if they make sense to you.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/SqsContainerOptionsBuilder.java
jvcalassio: you're right. thanks!

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/SqsListenerAcknowledgementMode.java
jvcalassio: yeah, "default" is better. I was thinking something like "inherit from the container (factory)".

I agree with using the const strings instead of enums -- I didn't really think about SpEL, thanks!!

I'll change it

jvcalassio:
Thanks for the comments @tomazfernandes. They all made a lot of sense, so I pushed a new version with the suggested changes. Let me know if you have any other suggestions

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/SqsListener.java
```diff
@@ -137,4 +137,10 @@
 	 */
 	String messageVisibilitySeconds() default "";
 
+	/**
+	 * The acknowledgement mode to be used for the provided queues. If not specified, the acknowledgement mode defined
+	 * for the container factory will be used.
+	 */
+	String acknowledgementMode() default SqsListenerAcknowledgementMode.DEFAULT;
```
tomazfernandes, line 144: Now that we've settled with String, we don't need this DEFAULT anymore, we can't simply default to `""`.

Then if the string has text we resolve the value, otherwise we leave it as null.

How does that sound?

docs/src/main/asciidoc/sqs.adoc
```diff
@@ -636,6 +636,8 @@ Note that if there are messages available the call may return earlier than this
 - `messageVisibilitySeconds` - Set the minimum visibility for the messages retrieved in a poll.
 Note that for `FIFO` single message listener methods, this visibility is applied to the whole batch before each message is sent to the listener.
 See <<FIFO Support>> for more information.
+- `acknowledgementMode` - Set the acknowledgement mode for the container. Default value is `SqsListenerAcknowledgementMode.DEFAULT`,
```
tomazfernandes, line 639: I think the behavior we're after is - if any `AcknowledgementMode` is set in the annotation, it takes precedence over the one set in the factory / `ContainerOptions`

tomazfernandes (commented):
Hey @jvcalassio, thanks for the changes!

Left a couple other comments, let me know if they make sense to you.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/SqsListener.java
jvcalassio: It does make sense haha. Thanks!

docs/src/main/asciidoc/sqs.adoc
jvcalassio: well, it's kind of stated above (line 628).

I'm down to change it anyway. What do you think of

```suggestion
- `acknowledgementMode` - Set the acknowledgement mode for the container. If any value is set, it will take precedence over the acknowledgement mode defined for the container factory options. See <<Acknowledgement Mode>> for more information.
```

?

tomazfernandes: Looks great, thanks!

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/SqsListener.java
tomazfernandes: Ok, still missing this change.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/SqsListenerAcknowledgementMode.java
```diff
@@ -0,0 +1,56 @@
...
+ * @see NeverAcknowledgementHandler
+ * @see io.awspring.cloud.sqs.listener.ContainerOptions
+ * @see SqsListener
+ */
+public class SqsListenerAcknowledgementMode {
+
+	/**
+	 * Use acknowledge mode defined by the container.
+	 */
+	public static final String DEFAULT = "DEFAULT";
```
tomazfernandes, line 38: No need for this now.

tomazfernandes (changes requested):
@jvcalassio not sure if you were done with the changes, just looking at them again.

Also please add your name as an author to the classes you changed.

Thanks!

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/SqsListener.java
jvcalassio: I was waiting for the feedback on the other comment to push the changes.. Sorry! It's now up to date

jvcalassio:
> @jvcalassio not sure if you were done with the changes, just looking at them again.
> 
> Also please add your name as an author to the classes you changed.
> 
> Thanks!

Did it! I also put `@since 3.1` on the new classes, considering that the original issue had 3.1 as milestone. Is that correct?

tomazfernandes:
@jvcalassio seems there are some failing tests - would you mind taking a look?

Thanks.

jvcalassio:
> @jvcalassio seems there are some failing tests - would you mind taking a look?
> 
> Thanks.

Sorry! The condition for resolving the value was inverted (even my own tests were failing 🤦). I ran all the SQS tests locally and it seems to be fine now.

tomazfernandes (approved):
Thanks for the PR @jvcalassio, looking forward for more!

Issue #881, opened 2023-09-12
spring.config.import order is not honored with third-party plugins


tomek82:
**Type**: Bug

**Component**: Secrets Manager

**Describe the bug**
Continuing from https://github.com/spring-projects/spring-boot/issues/37273

While trying to use a third party Spring Cloud for AWS to load configuration from Secrets Manager and also a local Properties resource I'm observing this:

With _application.properties_ having:
`spring.config.import=configtree:config/,classpath:config.properties`

I get this order:
```
OriginTrackedMapPropertySource {name='Config resource 'class path resource [config.properties]' via location 'classpath:config.properties''}
ConfigTreePropertySource {name='Config tree 'xxx\config''}
OriginTrackedMapPropertySource {name='Config resource 'class path resource [application.properties]' via location 'optional:classpath:/''}
```

Trying:
`spring.config.import=classpath:config.properties,configtree:config/`

Gives this order:
```
ConfigTreePropertySource {name='Config tree 'xxx\config''}
OriginTrackedMapPropertySource {name='Config resource 'class path resource [config.properties]' via location 'classpath:config.properties''}
OriginTrackedMapPropertySource {name='Config resource 'class path resource [application.properties]' via location 'optional:classpath:/''}
```

So far so good and consistent with documentation which says "Several locations can be specified under a single spring.config.import key. Locations will be processed in the order that they are defined, with later imports taking precedence."

However setting:
`spring.config.import=aws-secretsmanager:sample/config,classpath:config.properties`

Gives me this:

```
AwsSecretsManagerPropertySource {name='sample/config'}
OriginTrackedMapPropertySource {name='Config resource 'class path resource [config.properties]' via location 'classpath:config.properties''}
OriginTrackedMapPropertySource {name='Config resource 'class path resource [application.properties]' via location 'optional:classpath:/''}
```

which makes the Plugin take precedence even if declared before the additional file.

Running with Spring Boot 3.1.3 and spring-cloud-starter-aws-secrets-manager-config v2.4.4.

maciejwalkowiak:
I can't reproduce this issue in Spring Cloud AWS 3.0.2. See https://github.com/awspring/spring-cloud-aws/pull/882#issuecomment-1784549918

MatejNedic:
@maciejwalkowiak can we close this?

maciejwalkowiak:
Yup

PR #882, opened 2023-09-12
Fixed the order of importing properties

tomek82:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Change the Secrets Manager resolver to use resolve instead of resolveProfileSpecific to fix the import order bug.

https://github.com/awspring/spring-cloud-aws/issues/881

Build locally and run application with 
`spring.config.import=aws-secretsmanager:${ENV}/config,optional:classpath:config.properties`
and verified that the properties are overridden correctly. Also tested with profile-specific application-local.properties.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/config/secretsmanager/SecretsManagerConfigDataLocationResolver.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/config/secretsmanager/SecretsManagerConfigDataLocationResolver.java
@@ -24,10 +24,7 @@ import java.util.ArrayList;
 import java.util.List;
 import org.apache.commons.logging.Log;
 import org.springframework.boot.BootstrapContext;
-import org.springframework.boot.context.config.ConfigDataLocation;
-import org.springframework.boot.context.config.ConfigDataLocationNotFoundException;
-import org.springframework.boot.context.config.ConfigDataLocationResolverContext;
-import org.springframework.boot.context.config.Profiles;
+import org.springframework.boot.context.config.*;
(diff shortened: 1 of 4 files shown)


maciejwalkowiak (changes requested):
Thanks @tomek82. Good catch. Can you add a test for this change?

tomek82:
What is a good way to test it? My attempts with adding a classpath source to command line in io.awspring.cloud.autoconfigure.config.secretsmanager.SecretsManagerConfigDataLoaderIntegrationTests fail as it seems to be working without this change and I can't get a separate @PropertySource to work either (it does load the file but doesn't process the secrets import).

maciejwalkowiak:
You can test it like this.

Add test to `SecretsManagerConfigDataLoaderIntegrationTests`:

```
@Test
void respectsImportOrder() {
	SpringApplication application = new SpringApplication(App.class);
	application.setWebApplicationType(WebApplicationType.NONE);

	try (ConfigurableApplicationContext context = runApplication(application,
		"aws-secretsmanager:/config/spring,classpath:config.properties")) {
		assertThat(context.getEnvironment().getProperty("another-parameter")).isEqualTo("from properties file");
	}
}
```

and `config.properties` file in `src/test/resources`:

```properties
another-parameter=from properties file
```

The test passes without your changes too, so I am a bit puzzled. Perhaps this bug exists only in 2.x?

tomek82:
I've added a test with one more level of indirection, without my fix this fails to load the properties in the right order.

maciejwalkowiak (approved):
Thanks! Because it may have surprising side effects we will release it with 3.1.0

Issue #852, opened 2023-07-24
Work done in PR https://github.com/awspring/spring-cloud-aws/pull/248 for 2.4.1 missing from 3.x


zampettim:
**Type**: Bug

**Component**:"Parameter Store"

**Describe the bug**
The work done in PR https://github.com/awspring/spring-cloud-aws/pull/248 and released in tagged version 2.4.1 seems to have been lost or removed from 3.x. As far as I can tell, this was not intentional and should be easy to re-implement.

MatejNedic:
Thanks on pointing this out @zampettim ,
Are you willing to make PR :) ?

PR #248, opened 2022-02-19
Adds array properties support to parameter store. (#245)

rmpestano:
## :loudspeaker: Type of change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

This is a draft PR, the approach still needs to be discussed. Right now it's replacing the property names that follow the format `_INDEX_`  by `[INDEX]`

## :bulb: Motivation and Context

The main idea is to be able to represent array properties in aws parameter store params names, details in #245 

## :pencil: Checklist

- [x] I reviewed submitted code
- [x] I added tests to verify changes
- [x] I updated reference documentation to reflect the change
- [x] All tests passing
- [x] No breaking changes - ~It can break applications that have `_ANY INTEGER_` pattern in their parameter store param names~ It shouldn't break any existing apps as we changed only the internal representation of how we store list properties.

Issue #245, opened 2022-02-16
Question: How to represent array properties in parameter store?


rmpestano:
**Type**: Bug

**Component**: 
"Parameter Store"

**Describe the bug**
I need to represent [list properties](https://github.com/spring-projects/spring-boot/wiki/Relaxed-Binding-2.0#list-types-1) in parameter store, is it possible?

The main issue is that list indexes are represented by square brackets in Spring which are not supported as parameter names in [ssm API](https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_PutParameter.html), here's the pattern: `a-zA-Z0-9_.-`

**Sample**
Here's a sample spring config I want to represent on ssm:

```
config.users:
  - name: 'user1'
    password: `pass1`
  - name: 'user2'
    password: `pass2`
```

in parameter store I'd need the following parameters: 

/config/app/config.users[0].name
/config/app/config.users[0].password
/config/app/config.users[1].name
/config/app/config.users[1].password

But I cannot create them because of the brackets. 

Note that the [env var representation](https://github.com/spring-projects/spring-boot/wiki/Relaxed-Binding-2.0#lists-1) would work as it uses `_` for the array index.

rmpestano:
Hi again,

We fixed this in our project by representing array indexes with `_INDEX_`. The example on the description of this issue would be represented as:

```
/config/app/config.users_0_.name
/config/app/config.users_0_.password
/config/app/config.users_1_.name
/config/app/config.users_1_.password
```

Not sure that would work as a general solution without breaking existing applications.


If someone else needs this now, in our project we've created a property source that just gets the parameters resolved by AwsParamStorePropertySource (they'll have the underscore array index notation) and transform them into the system properties array notation (with brackets):

```
public class ArraySupportParamStorePropertySource implements PropertySourceLocator {

    private final Map<String, Object> configProperties;

    public ArraySupportParamStorePropertySource(final AwsParamStorePropertySource awsParamStorePropertySource) {
        var parameterStoreParams = Arrays.stream(awsParamStorePropertySource.getPropertyNames()).toList();
        configProperties = parameterStoreParams.stream()
                .filter(param -> param.startsWith("my-config"))
                .map(param -> Map.entry(param.replaceAll("_(\\d)_", "[$1]"), awsParamStorePropertySource.getProperty(param)))
                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
    }

    @Override
[...]

PR #894, opened 2023-09-21
Adds array properties support to parameter store

deki:
see #245 and #852, this PR just merges the changes from #248


maciejwalkowiak (approved):
Thank you @deki!

mikelhamer:
Love seeing these improvements! I recently starting using the param store feature and it's amazing. In my early testing I think I've tested using a StringList value type on 3.0.2 and had no issues with it resolving to an array though. Is this PR doing something other than that?

Issue #887, opened 2023-09-14
Map SNS notification message as object in SqsListener


lenlaurito:
**Type**: Feature

**Is your feature request related to a problem? Please describe.**
Is there a way to automatically map SNS notification messages in `@SqsListener`? If the SQS queue is subscribed to a topic, the listener for the queue would process the SNS Notification. How can we automatically map the SNS Notification body as an object in the Sqs Listener?

**Describe the solution you'd like**
Ideally we can use the `@NotificationSubject` and `@NotificationMessage` in the method for the `SqsListener`. This worked for earlier versions of spring cloud (2.x). But I can't use it for spring cloud 3. Is this possible?

```java
  @SqsListener(value = QUEUE_NAME)
  public void consumeMessage(
      @NotificationSubject String subject,
      @NotificationMessage SomeObject messageAsObject) {
    log.info(subject);
    log.info(messageAsObject)
  }
```

dbouclier:
@lenlaurito did you tried with this attribute => https://docs.aws.amazon.com/sns/latest/dg/sns-large-payload-raw-message-delivery.html

more explanation here: https://github.com/awspring/spring-cloud-aws/issues/886

lenlaurito:
@dbouclier Yes, we are also doing mapping of the "Message" in the SNS notification to an object-- similar to your comment. We're looking for an easier way of doing this if possible. Like having the "Message" mapped to an object automatically.

In previous versions of spring cloud (2.x), we were able to achieve this using `@NotificationMessage` annotation. But it doesn't seem to be working with spring cloud 3?

dbouclier:
it's working actually, I was able to automatically de-serialize object in my @SqsListener, like that: 

```
@SqsListener(queueNames = "some_queue.fifo")
    public void onEvent(final Message<MyCustomeType> event) {
      .....
    }

OR 

@SqsListener(queueNames = "some_queue.fifo")
    public void onEvent(final MyCustomeType event) {
      .....
    }
```

but in the subscription you have to enable "RawMessageDelivery" attribute

![image](https://github.com/awspring/spring-cloud-aws/assets/98832464/540464e4-6d0c-411f-88b6-3e3181e46e37)

lenlaurito:
@dbouclier We'd still like to be able to access the other fields of the notification though, like the subject.

tomazfernandes:
Hi @lenlaurito and @dbouclier.

[...]

PR #898, opened 2023-09-26
Add `SnsMessageConverter` and `NotificationMessageArgumentResolver` to allow reading SNS messages on SQS

msosa:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

#887 

It allows for reading an SNS Message on an SQS Queue by using `@NotficationMessage`

Manually(will try to add tests when I have time)

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

@tomazfernandes This is more of a draft for now, this works locally for me but was unsure where these files truly belong. I have put then in the sns project for now, since they use `@NotificationMessage` which lives there, and didn't want SQS project to have a dependency on the SNS one. 

Let me know if you agree, or we should create a new annotation to be used.

Also in the previous implementation of [NotificationRequestConverter](https://github.com/spring-attic/spring-cloud-aws/blob/2.3.x/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/support/converter/NotificationRequestConverter.java#L53-L79) it added `MessageAttributes` to the message. For my use-case I was able to remove that completely and things worked fine. Let me know if that is actually needed, but I did see `SqsHeaderMapper` has a lot of the same logic for what was being done for `MessageAttributes`

### DIFF
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -611,6 +611,19 @@ Any number of `@SqsListener` annotations can be used in a bean class, and each a
 
 NOTE: Queues declared in the same annotation will share the container, though each will have separate throughput and acknowledgement controls.
 
+===== Annotation-driven SqsListener
+
+The default payload of `@SqsListener` includes all the attributes within the SQS Object; However if you only need the `Message` part of the payload, you can utilize the `@SnsNotificationMessage` annotation.
+[source, java]
+----
+@SqsListener("my-queue")
+public void listen(@SnsNotificationMessage Pojo pojo) {
+	System.out.println(pojo.field);
+}
+----
+
+NOTE: You will need to configure an `ObjectMapper` with `SqsListenerConfigurer` in order to utilize this annotation.
+
 ===== Specifying a MessageListenerContainerFactory
 A `MessageListenerContainerFactory` can be specified through the `factory` property.
 Such factory will then be used to create the container for the annotated method.
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -611,9 +611,9 @@ Any number of `@SqsListener` annotations can be used in a bean class, and each a
 
 NOTE: Queues declared in the same annotation will share the container, though each will have separate throughput and acknowledgement controls.
 
-===== Annotation-driven SqsListener
+===== SNS Messages
 
-The default payload of `@SqsListener` includes all the attributes within the SQS Object; However if you only need the `Message` part of the payload, you can utilize the `@SnsNotificationMessage` annotation.
+When receiving SNS messages through the `@SqsListener`, the message includes all attributes of the `SnsNotification`. To only receive need the `Message` part of the payload, you can utilize the `@SnsNotificationMessage` annotation.
 [source, java]
 ----
 @SqsListener("my-queue")
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -613,7 +613,7 @@ NOTE: Queues declared in the same annotation will share the container, though ea
 
 ===== SNS Messages
 
-When receiving SNS messages through the `@SqsListener`, the message includes all attributes of the `SnsNotification`. To only receive need the `Message` part of the payload, you can utilize the `@SnsNotificationMessage` annotation.
+Since 3.1.1, when receiving SNS messages through the `@SqsListener`, the message includes all attributes of the `SnsNotification`. To only receive need the `Message` part of the payload, you can utilize the `@SnsNotificationMessage` annotation.
 [source, java]
 ----
 @SqsListener("my-queue")
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -622,8 +622,6 @@ public void listen(@SnsNotificationMessage Pojo pojo) {
 }
 ----
 
-NOTE: You will need to configure an `ObjectMapper` with `SqsListenerConfigurer` in order to utilize this annotation.
-
 ===== Specifying a MessageListenerContainerFactory
 A `MessageListenerContainerFactory` can be specified through the `factory` property.
 Such factory will then be used to create the container for the annotated method.
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/AbstractListenerAnnotationBeanPostProcessor.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/AbstractListenerAnnotationBeanPostProcessor.java
@@ -26,6 +26,7 @@ import io.awspring.cloud.sqs.listener.acknowledgement.handler.AcknowledgementMod
 import io.awspring.cloud.sqs.support.resolver.AcknowledgmentHandlerMethodArgumentResolver;
 import io.awspring.cloud.sqs.support.resolver.BatchAcknowledgmentArgumentResolver;
 import io.awspring.cloud.sqs.support.resolver.BatchPayloadMethodArgumentResolver;
+import io.awspring.cloud.sqs.support.resolver.NotificationMessageArgumentResolver;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
@@ -303,6 +304,7 @@ public abstract class AbstractListenerAnnotationBeanPostProcessor<A extends Anno
 				new HeadersMethodArgumentResolver(),
 				new BatchPayloadMethodArgumentResolver(messageConverter, this.endpointRegistrar.getValidator()),
 				new MessageMethodArgumentResolver(messageConverter),
+				new NotificationMessageArgumentResolver(messageConverter),
 				new PayloadMethodArgumentResolver(messageConverter,  this.endpointRegistrar.getValidator()));
 	}
 	// @formatter:on
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/AbstractListenerAnnotationBeanPostProcessor.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/AbstractListenerAnnotationBeanPostProcessor.java
@@ -285,6 +285,9 @@ public abstract class AbstractListenerAnnotationBeanPostProcessor<A extends Anno
 			MessageConverter messageConverter, ObjectMapper objectMapper) {
 		return Collections.emptyList();
 	}
+	protected Collection<HandlerMethodArgumentResolver> createAdditionalArgumentResolvers() {
+		return createAdditionalArgumentResolvers(null, null);
+	}
 
 	protected CompositeMessageConverter createCompositeMessageConverter() {
 		List<MessageConverter> messageConverters = new ArrayList<>();
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/AbstractListenerAnnotationBeanPostProcessor.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/AbstractListenerAnnotationBeanPostProcessor.java
@@ -283,10 +283,10 @@ public abstract class AbstractListenerAnnotationBeanPostProcessor<A extends Anno
 
 	protected Collection<HandlerMethodArgumentResolver> createAdditionalArgumentResolvers(
 			MessageConverter messageConverter, ObjectMapper objectMapper) {
-		return Collections.emptyList();
+		return createAdditionalArgumentResolvers();
 	}
 	protected Collection<HandlerMethodArgumentResolver> createAdditionalArgumentResolvers() {
-		return createAdditionalArgumentResolvers(null, null);
+		return Collections.emptyList();
 	}
 
 	protected CompositeMessageConverter createCompositeMessageConverter() {
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/AbstractListenerAnnotationBeanPostProcessor.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/AbstractListenerAnnotationBeanPostProcessor.java
@@ -26,7 +26,6 @@ import io.awspring.cloud.sqs.listener.acknowledgement.handler.AcknowledgementMod
 import io.awspring.cloud.sqs.support.resolver.AcknowledgmentHandlerMethodArgumentResolver;
 import io.awspring.cloud.sqs.support.resolver.BatchAcknowledgmentArgumentResolver;
 import io.awspring.cloud.sqs.support.resolver.BatchPayloadMethodArgumentResolver;
-import io.awspring.cloud.sqs.support.resolver.NotificationMessageArgumentResolver;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
@@ -275,14 +274,15 @@ public abstract class AbstractListenerAnnotationBeanPostProcessor<A extends Anno
 		CompositeMessageConverter compositeMessageConverter = createCompositeMessageConverter();
 
 		List<HandlerMethodArgumentResolver> methodArgumentResolvers = new ArrayList<>(
-				createAdditionalArgumentResolvers());
+				createAdditionalArgumentResolvers(compositeMessageConverter, this.endpointRegistrar.getObjectMapper()));
 		methodArgumentResolvers.addAll(createArgumentResolvers(compositeMessageConverter));
 		this.endpointRegistrar.getMethodArgumentResolversConsumer().accept(methodArgumentResolvers);
 		handlerMethodFactory.setArgumentResolvers(methodArgumentResolvers);
 		handlerMethodFactory.afterPropertiesSet();
 	}
 
-	protected Collection<HandlerMethodArgumentResolver> createAdditionalArgumentResolvers() {
+	protected Collection<HandlerMethodArgumentResolver> createAdditionalArgumentResolvers(
+			MessageConverter messageConverter, ObjectMapper objectMapper) {
 		return Collections.emptyList();
 	}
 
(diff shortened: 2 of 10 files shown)


tomazfernandes:
Hi @msosa, thanks for the PR.

I'm leaning towards having a separate annotation, since the modules are now separated, and I believe it would make sense for a project to consume an `SNS` message from `SQS` but not need anything from the `SNS` module itself.

We'd probably want to have a different name for the annotation so that it's not confusing for the users.

@MatejNedic, do you have any thoughts on this? Anything I might be missing regarding the integration between the modules?

Thanks

msosa:
Makes sense, I have moved all the code into SQS and created a new annotation called `SnsNotificationMessage`(open to alternative names).

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/NotificationRequestConverter.java
```diff
@@ -0,0 +1,103 @@
...
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.converter.MessageConversionException;
+import org.springframework.messaging.converter.MessageConverter;
+import org.springframework.messaging.support.GenericMessage;
+import org.springframework.util.Assert;
+
+/**
+ * @author Michael Sosa
+ */
+public class NotificationRequestConverter implements MessageConverter {
```
tomazfernandes, line 31: I'm not sure `Request` makes sense in this context. It seems more related to the `Controller` methods in the `SNS` integration. WDYT?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/NotificationRequestConverter.java
```diff
@@ -0,0 +1,103 @@
...
+
+	private final ObjectMapper jsonMapper = new ObjectMapper();
```
tomazfernandes, line 33: We should use the `ObjectMapper` provided by the user (if any).

tomazfernandes (changes requested):
Hey @msosa, thanks, it's looking good.

Added a few comments, let me know if they make sense to you.

It would also be great if you can include an integration test so we can see it in action.

Thanks

msosa (commented):
I am now in the process of making the integration test. I will complete that in another commit later on

spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration/SqsMessageConversionIntegrationTests.java
```diff
@@ -135,6 +140,17 @@ void resolvesMyOtherPojoFromHeader() throws Exception {
 		assertThat(latchContainer.resolvesMyOtherPojoFromMappingLatch.await(10, TimeUnit.SECONDS)).isTrue();
 	}
 
+	@Test
+	void resolvesMyPojoFromNotificationMessage() throws Exception {
```
msosa, line 144: Here is the test @tomazfernandes. It actually initially failed in the constructor of `NotificationRequestConverter` because the mapper was null.

```java
this.jsonMapper = jsonMapper == null ? new ObjectMapper() :jsonMapper
```

not sure if the reason was because I am using `this.endpointRegistrar.getObjectMapper()` to pass in the objectMapper, and I should be using something else. Or it is just a test issue that I need to address elsewhere. Let me know the best approach

jerchende:
Hi, 
I just saw your merge request. I am currently working around this missing feature, since we want to update to spring boot 3. And I wonder if the MessageAttributes from the SNS notification don't need to be mapped into MessageHeader as well?
Best regards,
Johannes

gustavomonarin:
Hello @msosa ,

Do you have plans to further work on this MR?

I have implemented something internally very similarly and would love to drop this code in favor of the opensource implementation.

The main difference was related to the MessageConverter. In our case we do use CloudEvents and the current MessageConverter is not able to handle well subtypes likes CloudEvent<MyEvent>. It is a super tiny change but that enables a lot. I would be more than happy to create a mr to your branch if you wish.

msosa:
HI @gustavomonarin,

Yes I do plan on working on this MR, was just waiting for review(I just realized I may not have requested said review)

But yeah am open to checking out and change you mentioned and bringing into this MR

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/AbstractListenerAnnotationBeanPostProcessor.java
```diff
@@ -261,14 +261,15 @@ protected void configureDefaultHandlerMethodFactory(DefaultMessageHandlerMethodF
...
-				createAdditionalArgumentResolvers());
+				createAdditionalArgumentResolvers(compositeMessageConverter, this.endpointRegistrar.getObjectMapper()));
 		methodArgumentResolvers.addAll(createArgumentResolvers(compositeMessageConverter));
 		this.endpointRegistrar.getMethodArgumentResolversConsumer().accept(methodArgumentResolvers);
 		handlerMethodFactory.setArgumentResolvers(methodArgumentResolvers);
 		handlerMethodFactory.afterPropertiesSet();
 	}
 
-	protected Collection<HandlerMethodArgumentResolver> createAdditionalArgumentResolvers() {
+	protected Collection<HandlerMethodArgumentResolver> createAdditionalArgumentResolvers(
```
tomazfernandes, line 271: Yup, this is a breaking change. We create an overload and have the old method call the new method without parameters. Then we need to make sure we can handle not having the arguments.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/NotificationRequestConverter.java
tomazfernandes: I think `SnsMessageConverter` should do the trick, wdyt?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/NotificationRequestConverter.java
```diff
@@ -0,0 +1,104 @@
...
+	@Override
+	public Message<?> toMessage(Object payload, MessageHeaders headers) {
+		throw new UnsupportedOperationException(
+				"This converter only supports reading a SNS notification and not writing them");
+	}
+
+	/**
+	 * Notification request wrapper.
+	 */
+	public static class NotificationRequest {
```
tomazfernandes, line 83: Maybe use a record instead?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/resolver/NotificationMessageArgumentResolver.java
```diff
@@ -0,0 +1,49 @@
...
+
+	@Override
+	public boolean supportsParameter(MethodParameter parameter) {
+		return parameter.hasParameterAnnotation(SnsNotificationMessage.class);
+	}
+
+	@Override
+	public Object resolveArgument(MethodParameter par, Message<?> msg) throws Exception {
+		Object object = this.converter.fromMessage(msg, par.getParameterType());
+		NotificationRequestConverter.NotificationRequest nr = (NotificationRequestConverter.NotificationRequest) object;
```
tomazfernandes, line 45: Lets use `Assert.isInstance()` here to throw a pretty message in case the instance doesn't match the expected class?

spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration/SqsMessageConversionIntegrationTests.java
tomazfernandes: We can use a `SqsListenerConfigurer` bean to set the `ObjectMapper` instance.

tomazfernandes (commented):
Looking good, left a few comments.

tomazfernandes:
Hey @msosa, really sorry for the delay - I thought I was the one waiting for your reply. Left a few comments on the PR, please let me know if they make sense to you.

@gustavomonarin thanks for showing up. Can you share a bit more about what's your proposed change?

Thanks.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/SqsListenerAnnotationBeanPostProcessor.java
```diff
@@ -66,9 +70,15 @@ protected String getMessageListenerContainerRegistryBeanName() {
 	}
 
 	@Override
-	protected Collection<HandlerMethodArgumentResolver> createAdditionalArgumentResolvers() {
-		return Arrays.asList(new VisibilityHandlerMethodArgumentResolver(SqsHeaders.SQS_VISIBILITY_TIMEOUT_HEADER),
+	protected Collection<HandlerMethodArgumentResolver> createAdditionalArgumentResolvers(
+			MessageConverter messageConverter, ObjectMapper objectMapper) {
+		var additionalArgumentResolvers = Arrays.asList(new VisibilityHandlerMethodArgumentResolver(SqsHeaders.SQS_VISIBILITY_TIMEOUT_HEADER),
 				new SqsMessageMethodArgumentResolver(), new QueueAttributesMethodArgumentResolver());
+		if (messageConverter != null && objectMapper != null) {
```
msosa, line 77: here is where I am handling the empty arguments, though maybe there is a better approach than this? I don't like it too much

msosa (commented):
No problem @tomazfernandes, I had it in draft so it was a bit confusing.

Also in regards to @jerchende's comment, I have implemented it locally similar to how it was done [previously](https://github.com/spring-attic/spring-cloud-aws/blob/2.3.x/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/support/converter/NotificationRequestConverter.java#L53-L79). But I had to make this [method](https://github.com/awspring/spring-cloud-aws/blob/e189ea5c740c6543006aeeab95b9821e9b6634a2/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SqsHeaderMapper.java#L239) public static. Would it make sense to move `NumberParser` to its own class, so we can use the `getNumberValue`?

tomazfernandes:
> current MessageConverter is not able to handle well subtypes likes CloudEvent<MyEvent>

@gustavomonarin, can you elaborate on this please?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/SqsListenerAnnotationBeanPostProcessor.java
tomazfernandes: Maybe we can invert this - we can leave the old method as it was, and in the new method we call get the list from it and add the new resolver.

Wdyt?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SnsMessageConverter.java
```diff
@@ -0,0 +1,86 @@
...
+	@Override
+	public Message<?> toMessage(Object payload, MessageHeaders headers) {
+		throw new UnsupportedOperationException(
+				"This converter only supports reading a SNS notification and not writing them");
+	}
+
+	/**
+	 * Notification request wrapper.
+	 */
+	public record SnsMessage(String subject, Object message) {
```
tomazfernandes, line 83: Since this is a wrapper, maybe we could call it `SnsMessageWrapper`?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SnsMessageConverter.java
```diff
@@ -0,0 +1,86 @@
...
+ * @author Michael Sosa
+ */
+public class SnsMessageConverter implements MessageConverter {
+
+	private final ObjectMapper jsonMapper;
+
+	private final MessageConverter payloadConverter;
+
+	public SnsMessageConverter(MessageConverter payloadConverter, ObjectMapper jsonMapper) {
+		this.payloadConverter = payloadConverter;
```
tomazfernandes, line 37: Let's add `Assert.notNull()` to these Constructor parameters too

spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration/SqsMessageConversionIntegrationTests.java
tomazfernandes: Makes sense?

msosa: Oh yeah, I did this on line 306, is that the correct place or should I be doing it somewhere else?

gustavomonarin:
Hi @msosa and @tomazfernandes  i mad a small change in the test to cover the case i mentioned. 

Also provided two implementation solutions, a small one in the spring way of compatibility and a second one with a bit more ofbreaking change on the bean post processor. 


The PR has not much deatils in the description but the two commit messages goes  in the details, so i would suggest to review the commits individually.

spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration/SqsMessageConversionIntegrationTests.java
tomazfernandes: > Oh yeah, I did this on line 306, is that the correct place or should I be doing it somewhere else?

Yes, in L306 it's correct.

Is it working as expected? Anything else to do regarding this?

tomazfernandes:
> Hi @msosa and @tomazfernandes i mad a small change in the test to cover the case i mentioned.
> 
> Also provided two implementation solutions, a small one in the spring way of compatibility and a second one with a bit more ofbreaking change on the bean post processor.
> 
> The PR has not much deatils in the description but the two commit messages goes in the details, so i would suggest to review the commits individually.

@gustavomonarin I don't see your PR with these commits, where can I find it?

tomazfernandes:
@msosa we're pretty much there!

The only thing missing is the documentation, let me know if you need any help with that.

Thanks!

gustavomonarin:
> > Hi @msosa and @tomazfernandes i mad a small change in the test to cover the case i mentioned.
> > 
> > Also provided two implementation solutions, a small one in the spring way of compatibility and a second one with a bit more ofbreaking change on the bean post processor.
> > 
> > The PR has not much deatils in the description but the two commit messages goes in the details, so i would suggest to review the commits individually.
> 
> @gustavomonarin I don't see your PR with these commits, where can I find it?

It is on the @msosa original repo, as is based on his changes.

spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration/SqsMessageConversionIntegrationTests.java
msosa: Yes working as expected, I moved it down in the code where it made a little more sense.

I also added a null check for `objectMapper` in `createAdditionalArgumentResolvers`, I noticed that if objectMapper was null it would cause some issues

msosa:
@tomazfernandes Awesome, appreciate all the feedback!

I gave the documentation a try, let me know what you think.

docs/src/main/asciidoc/sqs.adoc
```diff
@@ -611,6 +611,19 @@ Any number of `@SqsListener` annotations can be used in a bean class, and each a
 
 NOTE: Queues declared in the same annotation will share the container, though each will have separate throughput and acknowledgement controls.
 
+===== Annotation-driven SqsListener
```
tomazfernandes, line 614: @msosa, I'm not sure I understand this section. I was thinking we could have a `SNS Messages` or similar section describing this feature.

tomazfernandes:
@gustavomonarin, I looked at your suggestion to switch from implementing the `MessageConverter`  interface to implementing the `SmartMessageConverter` interface which allows type hints and it makes sense.

I prefer the solution in [this PR](https://github.com/msosa/spring-cloud-aws/pull/1/files) as it allows more flexibility for the user and is not a breaking change.

Now, not sure how to coordinate this - we do have an eminent version release coming up and it'd be great to have this feature in it.

@msosa, how would you feel about incorporating @gustavomonarin's changes into your PR?

docs/src/main/asciidoc/sqs.adoc
msosa: Thanks, that makes more sense

msosa:
@tomazfernandes of course, I have merged your preferred solution of @gustavomonarin into this PR.

Also I have updated the the documentation section to talk about `SNS Messages`

docs/src/main/asciidoc/sqs.adoc
```diff
@@ -611,6 +611,19 @@ Any number of `@SqsListener` annotations can be used in a bean class, and each a
...
+When receiving SNS messages through the `@SqsListener`, the message includes all attributes of the `SnsNotification`. To only receive need the `Message` part of the payload, you can utilize the `@SnsNotificationMessage` annotation.
+[source, java]
+----
+@SqsListener("my-queue")
+public void listen(@SnsNotificationMessage Pojo pojo) {
+	System.out.println(pojo.field);
+}
+----
+
+NOTE: You will need to configure an `ObjectMapper` with `SqsListenerConfigurer` in order to utilize this annotation.
```
tomazfernandes, line 625: We can leave this note out - in a regular Spring Boot application an `ObjectMapper` will already be auto configured.

tomazfernandes (commented):
Almost there @msosa, just one comment on the documentation

msosa:
Removed the `Note` @tomazfernandes, let me know if there's anything else

tomazfernandes (commented):
Looking great @msosa!

Just a couple of finishing touches.

1. Please add `@since 3.1.1` to the javadocs
2. Please add @gustavomonarin's name under yours in the relevant files
3. Please add "Since 3.1.1, when receiving SNS messages..."

That, and we're good to go!

Thanks!

msosa:
@tomazfernandes Done! I just used @gustavomonarin username as I wasn't sure if exact first/last name.

tomazfernandes (approved):
Great work @msosa! And thanks @gustavomonarin for the improvement!

Looking forward for more!

coco1979ka:
Looks great for Single Message processing. What is your recommendation to receive a List of SNSNotificationMessages for batch processing?

EstelaGarcia:
We are also looking for a way to receive a List of SnsNotificationMessages and struggling with this. Any suggestions?

Issue #938, opened 2023-11-03
Cannot send message to cross account FIFO SQS queue


ZhongXiLu:
**Type**: Bug

**Component**:
SQS

**Describe the bug**
Whenever I try to send messages to a FIFO SQS queue in another AWS account, it results in following error:
`Value ContentBasedDeduplication for parameter Attribute.Name is invalid. Reason: Only owners can access specified attribute.`

It seems retrieving the `ContentBasedDeduplication` queue attribute does not work for a cross-account SQS queue, see https://github.com/awspring/spring-cloud-aws/blob/main/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java#L265.

I'm using spring cloud version 3.0.2.
The bug seems to be introduced since 3.0.1 through this PR: https://github.com/awspring/spring-cloud-aws/pull/799.

**Sample**
```java
SqsTemplate sqsTemplate = SqsTemplate
                .builder()
                .sqsAsyncClient(SqsAsyncClient.create())
                .build();
        
sqsTemplate.send("<other_aws_account_sqs_queue_url>", "test");
```

maciejwalkowiak:
@tomazfernandes perhaps we can give users option to configure the value for `ContentBasedDeduplication` - in such case SQS integration instead of checking the value from real queue attributes would just believe in what user provided?

tomazfernandes:
> @tomazfernandes perhaps we can give users option to configure the value for ContentBasedDeduplication - in such case SQS integration instead of checking the value from real queue attributes would just believe in what user provided?

Sure. From the top of my mind, we could have a `ContentDeduplication` enum, such as:

`ContentDeduplication.ENABLED`
`ContentDeduplication.DISABLED`
`ContentDeduplication.AUTO`

And then AUTO would be the default so it's not a breaking change.

Makes sense?

maciejwalkowiak:
It does!

tomazfernandes:
> It does!

Cool!

@ZhongXiLu, what do you think? Would you be interested in opening a PR with this change?

ZhongXiLu:
@tomazfernandes Sure, I'll take a look 👍

tejashwinisubbanna:
Hi All, After migrating from Spring cloud 2 to 3.0.3.I am getting below error

[...]

PR #950, opened 2023-11-09
Fix sending message to cross account FIFO SQS queue, fixes gh-938.

ZhongXiLu:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Currently, it is not possible to send a message to a cross account FIFO SQS queue due to a bug where AWS does not allow you to access the `ContentDeduplication` queue attribute value of a cross account FIFO SQS queue.

As described in [this comment](https://github.com/awspring/spring-cloud-aws/issues/938#issuecomment-1796988934) and [this comment](https://github.com/awspring/spring-cloud-aws/pull/950#issuecomment-1814353850), I've introduced a new enum `ContentDeduplication` and added this as an option in `SqsTemplateOptions`.

If the value is `ContentDeduplication.AUTO`, then it follows the current implementation and will always retrieve the ContentDeduplication value at runtime. Otherwise, we take the value from the user and will not make an extra request to AWS to resolve this value. Of course, by default, the strategy is `ContentDeduplication.AUTO`.

Fixes [this issue](https://github.com/awspring/spring-cloud-aws/issues/938).

- Added some unit tests in `SqsTemplateTests`.
- Also manually tested with a real cross account FIFO SQS queue.

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -210,6 +210,13 @@ Such attributes are available as `MessageHeaders` in received messages.
 |All
 |Set the message system attribute names that will be retrieved with messages on receive operations.
 Such attributes are available as `MessageHeaders` in received messages.
+
+|`contentBasedDeduplication`
+|ContentBasedDeduplication
+|ContentBasedDeduplication
+#AUTO
+|Set the ContentBasedDeduplication queue attribute value of the queues the template is sending messages to.
+With `ContentBasedDeduplication#AUTO`, the queue attribute value will be resolved automatically.
 |===
 
 ==== Sending Messages
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/ContentBasedDeduplication.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/ContentBasedDeduplication.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2013-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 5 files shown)


tomazfernandes:
Hey @ZhongXiLu , thanks a lot for the PR.

Considering this issue is more of an edge case scenario, I think this configuration would be a better fit for the `SqsContainerOptions` instead of the main API.

What do you think?

maciejwalkowiak:
I agree with @tomazfernandes. Ideally we do not add new methods to `SqsOperations` for such thing.

ZhongXiLu:
@tomazfernandes I'm not too familiar with the code base, but the `SqsContainerOptions` is now only used for the `SqsMessageListenerContainer`, so there is no container for sending messages currently?

tomazfernandes:
> @tomazfernandes I'm not too familiar with the code base, but the `SqsContainerOptions` is now only used for the `SqsMessageListenerContainer`, so there is no container for sending messages currently?

Oh, my bad, I meant SqsTemplateOptio s.

ZhongXiLu:
> > @tomazfernandes I'm not too familiar with the code base, but the `SqsContainerOptions` is now only used for the `SqsMessageListenerContainer`, so there is no container for sending messages currently?
> 
> Oh, my bad, I meant SqsTemplateOptions.

Ah, I thought about that, but you're maybe right that it fits better here with this use case. I'll try to change it. 👍

ZhongXiLu:
@tomazfernandes I've rewritten my implementation, looks indeed better this way.

tejashwinisubbanna:
Hi Team , Can you please let me know when this fixes available to use (in the MVN repo)

tomazfernandes:
@ZhongXiLu, while reviewing your PR I found a bug in the content deduplication logic and had to change a few things.

I used your logic and added you as an author for `SqsTemplate` and the content deduplication logic, but couldn't use your commits directly.

Hope it's ok.

Thanks for the PR and looking forward for more!

Superseded by #987

PR #958, opened 2023-11-16
Reorganize readme & contributing pages.

maciejwalkowiak:
readme.md and contributing.md will be used on awspring.io.


tomazfernandes:
One thing I was thinking - should we direct users to tag questions on Stack Overflow with let's say `spring-cloud-aws` or similar so we can receive notifications?

maciejwalkowiak:
It has been already discussed before you joined I think, and the conclusion was that we don't want to give promises that we will respond to questions on StackOverflow. I still don't think I can commit to that.

Issue #954, opened 2023-11-15
Automatically populate PropertySource with instance metadata when running within an EC2 environment.


kennyk65:
On startup within an EC2 environment, populate a PropertySource containing EC2 instance metadata (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html).  Running outside of an EC2 environment should result in no properties being set.  Feature should work on both IMDS v1 and v2.  Should allow disabling with a property.  Should work whenever instance metadata is available, i.e. EC2, Elastic Beanstalk, ECS, EKS, etc.

PR #962, opened 2023-11-16
Automatically populate PropertySource with instance metadata when running within an EC2-based environment

kennyk65:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
A new package added to the autoconfigure project (with test classes) to populate a PropertySource whenever IMDS is present, which occurs on EC2, ElasticBeanstalk, ECS, EKS, etc.  Auto configuration silently ignores the absence of IMDS.  A starter pom is added, docs are updated.

Issue addressed is:  https://github.com/awspring/spring-cloud-aws/issues/954
There are various times when an application needs to know its IP address, availability zone, region, mac address, etc.  All this is available via IMDS.  This feature simply adds these properties to the Environment when available.

Three new classes, three new JUnit tests.  Plus I created a sample application using the libraries for a sanity test.

## :pencil: Checklist

- [x ] I added tests to verify changes
- [x ] I updated reference documentation to reflect the change
- [ x] All tests passing
- [x ] No breaking changes

## :crystal_ball: Next steps
Others may wish to expand this to retrieve dynamic data, user data, set a common prefix, etc.  May want to use a ConfigProperties only to cause the one and only property to be documented.


maciejwalkowiak:
Thanks @kennyk65 for a PR! We need an integration test for that. Until Localstack supports metadata endpoint (https://github.com/localstack/localstack/issues/8911) we would need to mock it - ideally with WireMock, to make sure whole flow works as expected.

docs/src/main/asciidoc/imds.adoc
```diff
@@ -0,0 +1,51 @@
...
+
+=== Enabling
+
+To enable instance metadata, add the spring-cloud-aws-starter-imds starter.  
+
+[source,xml]
+----
+<dependency>
+    <groupId>io.awspring.cloud</groupId>
+    <artifactId>spring-cloud-aws-starter-parameter-store</artifactId>
```
maciejwalkowiak, line 23:
```suggestion
    <artifactId>spring-cloud-aws-starter-imds</artifactId>
```

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/imds/ImdsPropertySource.java
```diff
@@ -0,0 +1,66 @@
...
+	@Override
+	public ImdsPropertySource copy() {
+		return new ImdsPropertySource(context, source);
+	}
+
+	@Override
+	public void init() {
+		if (!imdsUtils.isRunningOnCloudEnvironment())
+			return;
+		properties.putAll(imdsUtils.getEc2InstanceMetadata());
```
maciejwalkowiak, line 52: I'm thinking if there should not be an option to add prefix for properties retrieved with Imds. For example, `profile` property can already come from another property source and in such case it would be overwritten.

spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/imds/ImdsPropertySourceTest.java
```diff
@@ -0,0 +1,97 @@
...
+		when(mockUtils.isRunningOnCloudEnvironment()).thenReturn(true);
+		when(mockUtils.getEc2InstanceMetadata()).thenReturn(testMap);
+	}
+
+	@Test
+	public void testInit() {
+		ImdsPropertySource propertySource = new ImdsPropertySource("test", mockUtils);
+		propertySource.init();
+
+		assertTrue("Resulting PropertySource should contain the test data",
```
maciejwalkowiak, line 60: not a dealbreaker and i can fix it myself, but we use AssertJ assertions so would be nice if you change it too for sake of consistency.

maciejwalkowiak (changes requested):
Overall - great stuff! Thanks a lot! While most comments are details, an integration test is a must to merge it.

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/imds/ImdsPropertySource.java
kennyk65: I thought about that, but wanted to limit the scope for the initial commit.  This is my first one and it was starting to get big.

MatejNedic (approved):

maciejwalkowiak:
Thanks @kennyk65 once again and apologies it took so long.

kennyk65:
Thanks!  Let me know if I can help out with anything else.

PR #974, opened 2023-12-03
Fix S3Template.createSignedPutURL not applying given ObjectMetadata

LucasJC:
…Also added contentLength property to ObjectMetadata

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

- Method S3Template#createSignedPutURL is not applying the ObjectMetadata object that it is passed in as a parameter.
- Added contentLength property to ObjectMetadata class so it can be applied to the url build request.

We are migrating our s3 integration to spring-cloud-aws and found the S3Template class very useful.
Trying to migrate some code I found the mentioned bug and also that contentLength was not yet supported.

- First of all made sure that the build and all tests finished OK locally after making changes.
- Updated an existing test for the ObjectMetadata class. There is a test for createSignedPutURL but since it uses local stack, the content length property cant be really tested since it is of no effect at all (Real S3 does validate the created signature with the content length sent in the request).
- I installed my changes locally and tested them out on a real life project generating a presigned url and using it to upload an example file.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/ObjectMetadata.java
+++ b/spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/ObjectMetadata.java
@@ -45,6 +45,9 @@ public class ObjectMetadata {
 	private final String contentLanguage;
 
 	@Nullable
+	private final Long contentLength;
+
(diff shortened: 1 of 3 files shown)


maciejwalkowiak:
Thanks @LucasJC!

PR #996, opened 2023-12-13
fix : typo

rajadilipkolli:
Typo fix in workflow

### DIFF
--- a/.github/workflows/upload-docs.yml
+++ b/.github/workflows/upload-docs.yml
@@ -22,7 +22,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - name: Set up JDK 1.8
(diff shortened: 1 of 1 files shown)


maciejwalkowiak:
Thanks @rajadilipkolli! Once you're on it, there are other places in this file that also refer incorrectly Java 8, maybe you can change it too?

rajadilipkolli:
> Thanks @rajadilipkolli! Once you're on it, there are other places in this file that also refer incorrectly Java 8, maybe you can change it too?

Hi @maciejwalkowiak , I have reverified again and don't find it. Can you please point me to place where we need change?

maciejwalkowiak:
There's a `java-version: 8` but we can live with that. Thanks!

PR #1001, opened 2023-12-18
Assert queues are all the same type in setQueueNames so applications …

internetstaff:
…fail to start rather than throwing endless Executor errors.

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

This asserts there's not a mix of FIFO and non-FIFO queues earlier in `SqsMessageListenerContainer.setQueueNames` instead of waiting for the container to fail on startup.

Migrating from 2.x, we had an `@SqsListener` with a mix of queue types. Instead of failing at startup, there's only one line logged:

```s.c.a.AnnotationConfigApplicationContext : Exception encountered during context initialization - cancelling refresh attempt: org.springframework.context.ApplicationContextException: Failed to start bean 'io.awspring.cloud.messaging.internalEndpointRegistryBeanName'```

... followed by an endless loop of AWS SDK `executor not accepting a task` errors from all queues.

This is probably not an ideal fix - it seems like any failure in `io.awspring.cloud.sqs.listener.DefaultListenerContainerRegistry#start` should be fatal rather than leaving queues in a weird half-started state. I didn't go that deeply. Feel free to chuck this for a wider fix. :)

Added a simple UT, then pulled the SNAPSHOT into the application with the issue. Verified the app failed immediately with a very clear error.

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

### DIFF
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/SqsMessageListenerContainer.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/SqsMessageListenerContainer.java
@@ -122,14 +122,15 @@ public class SqsMessageListenerContainer<T>
 
 	@Override
 	protected Collection<ContainerComponentFactory<T, SqsContainerOptions>> createDefaultComponentFactories() {
-		Assert.isTrue(allQueuesSameType(),
-				"SqsMessageListenerContainer must contain either all FIFO or all Standard queues.");
 		return Arrays.asList(new FifoSqsComponentFactory<>(), new StandardSqsComponentFactory<>());
 	}
 
-	private boolean allQueuesSameType() {
-		return getQueueNames().stream().allMatch(this::isFifoQueue)
-				|| getQueueNames().stream().noneMatch(this::isFifoQueue);
+	@Override
+	public void setQueueNames(Collection<String> queueNames) {
+		Assert.isTrue(
+			queueNames.stream().allMatch(this::isFifoQueue) || queueNames.stream().noneMatch(this::isFifoQueue),
+			"SqsMessageListenerContainer must contain either all FIFO or all Standard queues.");
+		super.setQueueNames(queueNames);
 	}
 
 	private boolean isFifoQueue(String name) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/SqsMessageListenerContainer.java [hunk 1]
```diff
@@ -118,14 +118,21 @@ public SqsMessageListenerContainer(SqsAsyncClient sqsAsyncClient) {
...
 		return Arrays.asList(new FifoSqsComponentFactory<>(), new StandardSqsComponentFactory<>());
 	}
 
-	private boolean allQueuesSameType() {
-		return getQueueNames().stream().allMatch(this::isFifoQueue)
-				|| getQueueNames().stream().noneMatch(this::isFifoQueue);
+	@Override
+	public void setQueueNames(Collection<String> queueNames) {
+		super.setQueueNames(queueNames);
+		assertAllQueuesSameType();
```
tomazfernandes, line 128: I don't think we need to check this in two places, how about we only check it here on `setQueueNames` and do that before calling `super`?

internetstaff:
> I'm not sure I follow - as far as I'm aware if a container throws at startup, the app won't start up. So with the change you're proposing I believe it should prevent the app from starting up? What am I missing?
> 

Sorry for the delay - I had to fire this back up myself. :)

What I see with this condition is that I do get "Application run failed" but the application doesn't shut down. Instead, it stays in a zombie state.

With this change, it logs the error and shuts down.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/SqsMessageListenerContainer.java
internetstaff: Done, and confirmed this still resolves the issue for me.

tomazfernandes (approved):
Thanks for the contribution @internetstaff, looking forward to more!

PR #1008, opened 2024-01-03
Add BackOff when SQS Poll throws an exception

### LINKED ISSUE #714: SQS back off time equivalent in 3.x.x
### Discussed in https://github.com/awspring/spring-cloud-aws/discussions/713

<div type='discussions-op-text'>

<sup>Originally posted by **timothysugar** March 14, 2023</sup>
While using Starter SQS from 3.0.0-RC1, at times my queue consumer suffers from network connectivity failures with AWS. I see a very large number of errors raised due to reconnection attempts. My immediate issue is that this floods the logs with errors. It appears to me that in the 2.x versions, this behaviour could be affected using the `cloud.aws.sqs.listener.back-off-time` configuration property, but I don't see any equivalent for 3.x.

Is that correct? Is there any suggested way of affecting this behaviour?

Thank you for your contributions to this project.</div>

tomazfernandes:
Fixes #714

Adds support for a Spring Retry BackOffPolicy to be executed when an exception is thrown by a polling thread.

Adds the configuration to ContainerOptions with a exponential default of 1000ms / 2.0 / 10000ms.

### DIFF
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -812,6 +812,11 @@ See AWS documentation for more information.
 After that period, the framework will try to perform a partial acquire with the available permits, resulting in a poll for less than `maxMessagesPerPoll` messages, unless otherwise configured.
 See <<Message Processing Throughput>>.
 
+|<<pollBackOffPolicy>>
+|Any valid `BackOffPolicy` implementation
+|`ExponentialBackOffPolicy`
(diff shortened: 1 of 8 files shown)


CasaSky:
@tomazfernandes  @maciejwalkowiak how to integrate exponential retry? I found it in the interface `ContainerOptionsBuilder` but it seems not to be supported yet in version 3.1.1? Or is it just becoming a default behavior for SQS Listeners?
Is there an example how to use it?

PR #1009, opened 2024-01-03
Add `enabled` flag to Secrets Manager integration.

### LINKED ISSUE #1005: Add 'enabled' Property for Secret Manager configuration
**Type**: Feature

I would like to propose the addition of an enabled property for the AWS Secret Manager configuration in Spring Cloud AWS. Currently, there seems to be no direct way to enable or disable the Secret Manager integration globally via configuration properties.

The enabled property would allow users to programmatically enable or disable the Secret Manager usage in their Spring Cloud applications. This is particularly useful for scenarios where different environments require different configurations or for testing purposes.

**Proposed Solution**
Introduce an enabled property under the spring.cloud.aws.secretsmanager namespace. The property should be a boolean with a default value of true, maintaining the current behavior as the default.

```yml
spring:
  cloud:
    aws:
      secretsmanager:
        enabled: true  # or false to disable the integration
```

Other services already support it, like S3, SES, SNS, SQS and DynamoDB.

**Workaround?**

The SecretManagerConfigDataLocationResolver is part of the internals of the Spring Framework and its lifecycle differs from the regular Spring beans. It is not a bean managed by Spring's application context. It is a component being used during the early stage of application context initialization. Therefore, it doesn't fall under Spring's bean management system and cannot be mocked using @MockBean annotation like other auto-configured beans.

The only workaround I can think of is the following:

Do not specify the spring cloud config in your `application.yml`:
```yml
spring:
#  cloud:
#    aws:
#     parameterstore:

```

But, add the configuration in a separate environment application yaml file, like for example, `application-prd.yml`:
```yml
spring:
  config:
    import: aws-secretsmanager:/ms/common?prefix=common;
  cloud:
    aws:
      secretsmanager:
        reload:
          period: 60s
          strategy: refresh

```

By doing so, your specific test config yaml, like for example:`application-integration-test.yml`, will not inherit the config, and is because of that disabled.

The downside is that, you need to duplicate the spring cloud config to all environments (application-acc.yml etc.).

Btw, also, overriding the `config.activate.on-profile` does not work.

maciejwalkowiak:
Fixes #1005

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/config/secretsmanager/SecretsManagerConfigDataLoader.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/config/secretsmanager/SecretsManagerConfigDataLoader.java
@@ -16,6 +16,7 @@
 package io.awspring.cloud.autoconfigure.config.secretsmanager;
 
 import io.awspring.cloud.autoconfigure.config.BootstrapLoggingHelper;
+import io.awspring.cloud.secretsmanager.SecretsManagerPropertySource;
 import java.util.Collections;
 import java.util.Map;
 import org.springframework.boot.context.config.ConfigData;
@@ -25,8 +26,6 @@ import org.springframework.boot.context.config.ConfigDataResourceNotFoundExcepti
 import org.springframework.boot.logging.DeferredLogFactory;
 import org.springframework.core.env.MapPropertySource;
 import org.springframework.lang.Nullable;
-
-import io.awspring.cloud.secretsmanager.SecretsManagerPropertySource;
 import software.amazon.awssdk.services.secretsmanager.SecretsManagerClient;
 
 /**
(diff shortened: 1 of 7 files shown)


maciejwalkowiak:
@MatejNedic @tomazfernandes up for a review. Same approach will be used for Parameter Store

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/config/secretsmanager/SecretsManagerConfigDataLoader.java
```diff
@@ -46,14 +48,23 @@ public SecretsManagerConfigDataLoader(DeferredLogFactory logFactory) {
...
 	public ConfigData load(ConfigDataLoaderContext context, SecretsManagerConfigDataResource resource) {
 		try {
-			SecretsManagerClient sm = context.getBootstrapContext().get(SecretsManagerClient.class);
-			SecretsManagerPropertySource propertySource = resource.getPropertySources()
-					.createPropertySource(resource.getContext(), resource.isOptional(), sm);
-			if (propertySource != null) {
-				return new ConfigData(Collections.singletonList(propertySource));
+			// resource is disabled if secrets manager integration is disabled via
+			// spring.cloud.aws.secretsmanager.enabled=false
+			if (resource.isEnabled()) {
```
MatejNedic, line 53: I would extract everything under if block in new method.

MatejNedic (approved):

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/config/secretsmanager/SecretsManagerConfigDataLoader.java
maciejwalkowiak: I am not a fan of private methods. Not critical, merging as it is.

PR #1013, opened 2024-01-08
fix issue with running Java 8

rajadilipkolli:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

After increasing the setup-java to v4, it needs distribution as a mandatory input parameter

Ran the pipeline in my local and it is able to download the version and proceed further

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/.github/workflows/upload-docs.yml
+++ b/.github/workflows/upload-docs.yml
@@ -28,11 +28,12 @@ jobs:
         with:
           candidate: java
           version: '17.0.1-tem'
-      - name: Set up 8.0.332-tem
+      - name: Set up JDK 8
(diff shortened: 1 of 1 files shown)


maciejwalkowiak:
@rajadilipkolli apologies for late reply. We should go with 17. But we also have to make a decision either we remove using SDKMan here, or we use version provided by SDKMan. I don't really have a strong opinion. @MatejNedic do you have any preference or good reason to use SDKMan in Github actions, considering all Java versions we need are available out of the box?

PR #1038, opened 2024-02-03
Add Missing Synchronization in Ack Buffer Access

### LINKED ISSUE #1029: Spring Cloud AWS SQS - some messages do not get acknowledged
**Type**: Bug

**Component**:
SQS

**Describe the bug**
JDK:17
spring-cloud-aws-dependencies : 3.1.0
org.springframework.boot: 3.2.2

When using `BatchingAcknowledgementProcessor` some messages do not get acknowledged and end up in a dead letter queue even if the processing went without any errors. Looking at the trace logs it seems that there is some race condition between purging `acksBuffer`  and adding new items to it:
`Jan 24, 2024 @ 20:09:37.170 Received message ${messageId} to ack in io.awspring.cloud.sqs.sqsListenerEndpointContainer#2-0.`
`Jan 24, 2024 @ 20:09:37.170 Removing groups [default] from buffer in io.awspring.cloud.sqs.sqsListenerEndpointContainer#2-0`

**Sample**
```
    @Bean
    fun defaultSqsListenerContainerFactory(sqsAsyncClient: SqsAsyncClient): SqsMessageListenerContainerFactory<Any> {
        return SqsMessageListenerContainerFactory.builder<Any>()
            .configure { options -> options
                .acknowledgementMode(AcknowledgementMode.MANUAL)
            }
            .sqsAsyncClient(sqsAsyncClient)
            .build()
    }

  @SqsListener(queueNames = "queue-name")
    fun process(
        sqsMessage: SqsMessage,
        acknowledgement: Acknowledgement
    ) {
        process()
       acknowledgment.acknowledge()
    }
```

tomazfernandes:
Fixes #1029

There was one piece of missing synchronization to the Acknowledgement Buffer that led to acknowledgements being lost.

This commit adds the missing synchronization to fix the issue.

### DIFF
--- a/spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration/BaseSqsIntegrationTest.java
+++ b/spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration/BaseSqsIntegrationTest.java
@@ -44,7 +44,9 @@ abstract class BaseSqsIntegrationTest {
 
 	protected static final boolean useLocalStackClient = true;
 
-	protected static final boolean purgeQueues = true;
+	protected static boolean purgeQueues = true;
+
+	protected static boolean waitForPurge = true;
 
 	private static final String LOCAL_STACK_VERSION = "localstack/localstack:2.3.2";
 
@@ -93,7 +95,14 @@ abstract class BaseSqsIntegrationTest {
 			if (purgeQueues) {
 				String queueUrl = v.queueUrl();
 				logger.debug("Purging queue {}", queueName);
-				return client.purgeQueue(req -> req.queueUrl(queueUrl).build());
+				return client.purgeQueue(req -> req.queueUrl(queueUrl).build())
+					.thenRun(() -> {
+						if (waitForPurge) {
+							logger.info("Waiting 30000 seconds to start sending.");
+							sleep(30000);
+							logger.info("Done waiting.");
+						}
+					});
 			}
 			else {
 				logger.debug("Skipping purge for queue {}", queueName);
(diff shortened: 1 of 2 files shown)


tomazfernandes:
Note: I wasn't able to reproduce acknowledgements being lost in many runs, even with a few million messages and 1000 threads, which indicates it's likely a minor racing condition.

Also, in my tests there was no noticeable difference in performance with the added synchronization, which is expected since all processing is non-blocking and the bottleneck of the overall process is I/O.

spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration/BaseSqsIntegrationTest.java
```diff
@@ -93,7 +95,14 @@ protected static CompletableFuture<?> createQueue(SqsAsyncClient client, String
 			if (purgeQueues) {
 				String queueUrl = v.queueUrl();
 				logger.debug("Purging queue {}", queueName);
-				return client.purgeQueue(req -> req.queueUrl(queueUrl).build());
+				return client.purgeQueue(req -> req.queueUrl(queueUrl).build())
+					.thenRun(() -> {
+						if (waitForPurge) {
+							logger.info("Waiting 30000 seconds to start sending.");
```
anna-kr, line 101: 30000 seconds seems like quite long ;)

tomazfernandes: Whoops... I do wish I made this mistake in the logs less often, but it just comes naturally apparently 🙂 

I've fixed this and changed a couple default test settings.

Anything else you see?

Thanks.

anna-kr: it looks good to me now 🤌

anna-kr (approved):

Issue #923, opened 2023-10-25
[Parameter Store] Strip leading dot from imported parameters


matthias-seibert-tng:
**Type**: Bug

**Component**: Parameter Store

**Describe the bug**
The library adds a leading dot (`.`) to imported variables, when using the setup as suggested in the readme (`spring.config.import=aws-parameterstore:/config/spring`)

This leads to problems when importing variables stored in a nested hierarchy. For example, if a variable stored at `/config/spring/my/message`, it is imported as `.my.message`, and cannot be accessed via `@Value("${my.message}")`.

Changing the config to  `spring.config.import=aws-parameterstore:/config/spring/` resolves this problem, but it would be a nice feature to strip the leading dot in the library.

MatejNedic:
Hey @matthias-seibert-tng , since we are loading by path it is important to end with the `/` .
It only makes sense since you want all on the path.

matthias-seibert-tng:
I can understand that, but then it would be great if you could update the docs, because in the examples there, there is no slash at the end. See this screenshot:

![Screenshot](https://github.com/awspring/spring-cloud-aws/assets/61148008/6f8dccf5-9108-473a-8ef1-6cdd28a2d453)

maciejwalkowiak:
Closed in #1086

PR #1086, opened 2024-03-10
Added trailing slash to parameter store examples

### LINKED ISSUE #1058: Parameter Store path requires trailing slash
I created some parameters in parameter store with a "/dev/appname" prefix, e.g. "/dev/appname/spring.datasource.url". I configured the `spring.config.import` property via an environment variable. As per the documentation, I did not include a trailing slash.

```
SPRING_CONFIG_IMPORT=aws-parameterstore:/dev/appname
```

Looking at the debug log, I saw that a leading period "." was added to the property name, e.g.

```
2024-02-20T09:37:54.470+11:00 DEBUG 33036 --- [           main] i.a.c.p.ParameterStorePropertySource     : Populating property retrieved from AWS Parameter Store: .spring.datasource.url
```

This meant that my app wouldn't start because it couldn't connect to the database, since the "spring.datasource.url" was not populated. This behaviour makes sense looking at the `ParameterStorePropertySource` code in the `getParameters` method, which removes the path "/dev/appname" leaving "/spring.datasource.url", then replaces all remaining slashes with periods, i.e. ".spring.datasource.url".

Changing the path to include a trailing slash, i.e. "aws-parameterstore:/dev/appname/" resolved the problem, and the correct "spring.datasource.url" property was populated. To prevent others experiencing this problem in the future, I think that either
1. the examples in the documentation should be updated to specify a trailing slash in the parameter path, or 
2. the code should be updated to strip any leading slashes after the path is removed.

I'm happy to submit a PR if that helps.

I am using spring-cloud-aws-starter-parameter-store 3.1.0.

thelateperseus:
The trailing slash is required to avoid a leading dot in the property name. For example, if the parameter is named `/config/spring/message` and the `spring.config.import` property is `aws-parameterstore:/config/spring`, then the resulting property in the Spring environment will be `.message`. Instead, the property should be `message`, so the `spring.config.import` property should be `aws-parameterstore:/config/spring/`

Fixes #1058, #923

### DIFF
--- a/docs/src/main/asciidoc/parameter-store.adoc
+++ b/docs/src/main/asciidoc/parameter-store.adoc
@@ -17,7 +17,7 @@ Maven coordinates, using <<index.adoc#bill-of-materials, Spring Cloud AWS BOM>>:
 
 To fetch parameters from Parameter Store and add them to Spring's environment properties, add `spring.config.import` property to `application.properties`:
 
-For example, assuming that the parameters in Parameter Store are stored under path `/config/spring`:
+For example, assuming that the parameters in Parameter Store are stored under path `/config/spring/`:
 
 |===
 | Parameter Name | Parameter Value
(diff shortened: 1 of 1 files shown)


maciejwalkowiak (approved):
Thanks a lot @thelateperseus!

PR #1127, opened 2024-04-03
Add SQS QueueNotFoundStrategy auto-configure support.

maxjiang153:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
By default, when using SQS, even if a queue doesn't exist or is mispelled, it will automatically create a new queue, and there are no autoconfiguration options supported for this feature. So add a new auto-configure option to control the behavior of this.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sqs/SqsProperties.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sqs/SqsProperties.java
@@ -16,6 +16,8 @@
 package io.awspring.cloud.autoconfigure.sqs;
 
 import io.awspring.cloud.autoconfigure.AwsClientProperties;
+import io.awspring.cloud.sqs.listener.QueueNotFoundStrategy;
+
 import java.time.Duration;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.lang.Nullable;
@@ -24,6 +26,7 @@ import org.springframework.lang.Nullable;
  * Properties related to AWS SQS.
  *
  * @author Tomaz Fernandes
+ * @author Wei Jiang
  * @since 3.0
  */
 @ConfigurationProperties(prefix = SqsProperties.PREFIX)
@@ -36,12 +39,22 @@ public class SqsProperties extends AwsClientProperties {
 
 	private Listener listener = new Listener();
 
+	private Options options = new Options();
+
 	public Listener getListener() {
 		return this.listener;
 	}
 
 	public void setListener(Listener listener) {
 		this.listener = listener;
+	}
+
+	public Options getOptions() {
+		return this.options;
+	}
+
+	public void setOptions(Options options) {
+		this.options = options;
 	}
 
 	public static class Listener {
@@ -91,6 +104,23 @@ public class SqsProperties extends AwsClientProperties {
 		public void setPollTimeout(Duration pollTimeout) {
 			this.pollTimeout = pollTimeout;
 		}
+	}
+
+	public static class Options {
+
+		/**
+		 * The default {@link QueueNotFoundStrategy}
+		 */
+		private QueueNotFoundStrategy queueNotFoundStrategy = QueueNotFoundStrategy.CREATE;
+
+		public QueueNotFoundStrategy getQueueNotFoundStrategy() {
+			return queueNotFoundStrategy;
+		}
+
+		public void setQueueNotFoundStrategy(QueueNotFoundStrategy queueNotFoundStrategy) {
+			this.queueNotFoundStrategy = queueNotFoundStrategy;
+		}
+
 	}
 
 }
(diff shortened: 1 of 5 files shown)


spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sqs/SqsProperties.java
```diff
@@ -93,4 +106,21 @@ public void setPollTimeout(Duration pollTimeout) {
 		}
 	}
 
+	public static class Options {
+
+		/**
+		 * The default {@link QueueNotFoundStrategy}
+		 */
+		private QueueNotFoundStrategy queueNotFoundStrategy = QueueNotFoundStrategy.CREATE;
```
tomazfernandes, line 114: Your approach is solid, but we already define the default in  `ContainerOptions`.

I think we could be more defensive in this PR and not have a default set here, and instead only set these values to the Options objects if the user adds a value in the configuration.

Wdyt?

tomazfernandes (commented):
Cool PR @wmz7year!

Left one comment for your consideration, please let me know your thoughts.

Thanks.

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sqs/SqsProperties.java
wmz7year: Yes, `ContainerOptions` is a good way to do this, but there are two methods to use SQS component: the first one is 'SqsMessageListenerContainerFactory', and the second one is `SqsTemplate`.
If we only set `QueueNotFoundStrategy` on `ContainerOptions`, the `SqsTemplate` won't work.


And also about the default value, currently both `SqsTemplate` and `SqsContainerOptions` are defined as default `QueueNotFoundStrategy` to `CREATE` value. To avoid making a change about this, I keep the default value here.


Check these out:
https://github.com/awspring/spring-cloud-aws/blob/081366151f7f3f302f25132d2665ce6ddd7d38ec/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java#L632


https://github.com/awspring/spring-cloud-aws/blob/081366151f7f3f302f25132d2665ce6ddd7d38ec/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/SqsContainerOptions.java#L128


TBH, I think `QueueNotFoundStrategy` default value should be `FAIL`. If someone first uses this component and doesn't have create queue permission, this will be more clearly stated in this context.

tomazfernandes: > And also about the default value, currently both SqsTemplate and SqsContainerOptions are defined as default QueueNotFoundStrategy to CREATE value. To avoid making a change about this, I keep the default value here.

Yup, my suggestion is we do not define a default for autoconfiguration. If the user sets a value we set it, otherwise we don't do anything. This way whatever is the default set for either the Template or Container is respected.

Makes sense?

> I think QueueNotFoundStrategy default value should be FAIL

I see, the idea is to facilitate development time which is common, so users don't need to add logic to create queues when targeting e.g. LocalStack, or even AWS itself in a dev environment.

Maybe we can improve the error thrown in this case to make it more clear - let me know if you have any suggestion.

Thanks!

wmz7year: > Yup, my suggestion is we do not define a default for autoconfiguration

Yes, I can remove the default value for autoconfiguration. If I remove the `CREATE` default strategy, the component's default behavior will still be `CREATE`. So that's the reason I keep this default value to make it clearer. So I wonder if this is necessary or if I should just leave it here.

> Maybe we can improve the error thrown in this case to make it more clear - let me know if you have any suggestion.

I agree. Maybe we can add a log when trying to create a new queue?  If the strategy is `CREATE` and the user doesn't have create queue permissions, they will know why to throw a create queue permissions error while they just want to receive or send a message to the queue.

tomazfernandes: > Yes, I can remove the default value for autoconfiguration. If I remove the CREATE default strategy, the component's default behavior will still be CREATE. So that's the reason I keep this default value to make it clearer. So I wonder if this is necessary or if I should just leave it here.

I think it's important that we leave the default in a single place - in this case in the `Template` and `ContainerOptions` themselves.

If we want to make it clearer, we can add a javadoc linking to the relevant classes and / or write a comment.

Makes sense?

> I agree. Maybe we can add a log when trying to create a new queue?  

Sounds good!

> If the strategy is CREATE and the user doesn't have create queue permissions, they will know why to throw a create queue permissions error while they just want to receive or send a message to the queue.

Yeah, and if the error is not related to the create logic, we can also add a message asking the user to check for connection errors in the stacktrace or something similar.

tomazfernandes:
@maxjiang153, I made a few changes and opened this other PR to expedite merging:
https://github.com/awspring/spring-cloud-aws/pull/1190

Thank you for the PR, looking forward to more!

Issue #1041, opened 2024-02-07
Enable receiving messages from different message groups in the same Batch for FIFO queues


tomazfernandes:
See https://github.com/awspring/spring-cloud-aws/issues/1020

For this we need to add a `ContainerOptions` to use in the `FifoSqsComponentFactory` to determine the proper `Sink` composition.

alexisgra:
Hello @tomazfernandes ,

From what I've seen and in relation to your messages, I think we should add an option in `SqsContainerOptions`  to not use the `MessageGroupingSinkAdapter` but another one that doesn't group by message group.

Is that right? If yes, I can do it ! 
Thanks !

tomazfernandes:
Hey @alexisgra, yup, that sounds about right!

Go for it, I'll be happy to look at the PR.

alexisgra:
Hi @tomazfernandes,
Thanks for your answer! 
I'm currently working on the PR. I was wondering if I should expose a property in the @SqsListener annotation? Or just in the factory?
From what I understand from #1020, the use case may seem rather edgy. Moreover, the property only makes sense for FIFO queues with a BATCH listener.

For now, I've added a boolean `messageGrouping` field to the `SQSContainerOptions`, which defaults to true and is used in the `FifoSqsComponentFactory` to choose the right `Sink` composition.

Thanks for your help !

PR #1134, opened 2024-04-09
#1041 Enable receiving messages from different message groups i…

alexisgra:
…n the same Batch for FIFO queues

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Added a new SqsContainerOptions option : `messageGrouping`. True by default. When false, a new `MessageSinkAdapter` is used. This adapter does not have grouping function.

I was wondering whether I should add the new option `messageGrouping` in the `@SqsListener` annotation or leave the configuration only in the factory. 
Although the option is only relevant for FIFO Queues with a listener in `batch` mode, I've decided to put the option in `@SqsListener`, which will make it easier for library users to configure.

I don't know if the name `messageGrouping` is meaningful enough, I can change it if necessary.

Fix the issue #1041 

I added a test in `SqsFifoIntegrationTests`. In this test, I configure a new listener with `messageGrouping` set to `false`. I send several messages from two different message groups. 
In the listener, I verify that the batch of messages received contains messages from different message groups. Then, I check that all messages from each group have been received.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/FifoSqsComponentFactory.java
```diff
@@ -70,9 +71,11 @@ public MessageSource<T> createMessageSource(SqsContainerOptions options) {
...
 	public MessageSink<T> createMessageSink(SqsContainerOptions options) {
 		MessageSink<T> deliverySink = createDeliverySink(options.getListenerMode());
-		return new MessageGroupingSinkAdapter<>(
-				maybeWrapWithVisibilityAdapter(deliverySink, options.getMessageVisibility()),
-				getMessageGroupingFunction());
+		MessageSink<T> wrappedDeliverySink = maybeWrapWithVisibilityAdapter(deliverySink,
+				options.getMessageVisibility());
+		return options.getMessageGrouping()
+				? new MessageGroupingSinkAdapter<>(wrappedDeliverySink, getMessageGroupingFunction())
+				: new MessageSinkAdapter<>(wrappedDeliverySink);
```
tomazfernandes, line 78: Why do we need this `MessageSinkAdapter`?

Can't we return the `wrappedDeliverySink` directly?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/SqsListener.java
```diff
@@ -131,6 +131,13 @@
 	 */
 	String maxMessagesPerPoll() default "";
 
+	/**
+	 * Specifies whether messages coming from FIFO queues should be grouped when retrieved by the container in listener
+	 * mode {@link io.awspring.cloud.sqs.listener.ListenerMode#BATCH}.
+	 * @return message grouping boolean
+	 */
+	boolean messageGrouping() default true;
```
tomazfernandes, line 139: While I understand the practicality of having this configuration in the annotation, this seems to be a configuration for a reasonably specific case. We need to keep the annotation lean, otherwise we may end up with every container option in it, and it might become too polluted.

So my suggestion is we leave it only in the `ContainerOptions`.

What do you think?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/config/SqsEndpoint.java
```diff
@@ -88,6 +91,15 @@ public Integer getMaxMessagesPerPoll() {
 		return this.maxMessagesPerPoll;
 	}
 
+	/**
+	 * Return whether message should be grouped or not.
+	 * @return the message grouping.
+	 */
+	@Nullable
+	public Boolean getMessageGrouping() {
```
tomazfernandes, line 99: How about instead of having a boolean for this configuration, which can be a bit cryptic sometimes and is hard to evolve, we use an enum instead?

Something like:
`FifoBatchGroupingStrategy` 
PROCESS_MESSAGE_GROUPS_IN_PARALLEL
RECEIVE_ALL_GROUPS_IN_SAME_BATCH

Not sold on the names, but something similar might give users a better sense of what the options are, and allows us to add more possibilities down the road should the need appear.

What do you think?

tomazfernandes:
Thanks for the PR @alexisgra! Left a few comments for your consideration.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/SqsListener.java
alexisgra: I agree, I asked myself the same question and then I said: "who can do more can do less" so I implemented it. I think it's great to keep the property only in `ContainerOptions` !

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/config/SqsEndpoint.java
alexisgra: I did not think about it, but I think it's a great idea, thank you!

alexisgra:
Thank you very much for the review @tomazfernandes !
I've made the changes and updated the doc. I remain available if other changes need to be made 😃

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/SqsContainerOptions.java
```diff
@@ -181,6 +196,13 @@ public SqsContainerOptionsBuilder messageVisibility(Duration messageVisibility)
 			return this;
 		}
 
+		@Override
+		public SqsContainerOptionsBuilder batchGroupingStrategy(FifoBatchGroupingStrategy batchGroupingStrategy) {
```
tomazfernandes, line 200: I think we can name this `fifoBatchGroupingStrategy`, since this only applies to FIFO queues, wdyt?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/FifoBatchGroupingStrategy.java
```diff
@@ -0,0 +1,35 @@
...
+
+	/**
+	 * Group messages by message group. Each batch contains messages from a single message group.
+	 */
+	PROCESS_MESSAGE_GROUPS_IN_PARALLEL_BATCHES,
+
+	/**
+	 * Each batch contains messages originating from all the message groups.
+	 */
+	PROCESS_ALL_GROUPS_IN_EACH_BATCH
```
tomazfernandes, line 34: How about `PROCESS_MULTIPLE_GROUPS_IN_SAME_BATCH`?

It might be helpful to add that in this case message order within each group will still be preserved, since FIFO will only serve a message group again after all messages from a prior batch for that group is acknowledged.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/FifoBatchGroupingStrategy.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @author Alexis SEGURA
+ * @since 3.1.2
+ */
+public enum FifoBatchGroupingStrategy {
+
+	/**
+	 * Group messages by message group. Each batch contains messages from a single message group.
+	 */
+	PROCESS_MESSAGE_GROUPS_IN_PARALLEL_BATCHES,
```
tomazfernandes, line 29: We could add that this is default and provides maximum throughput while keeping order of the messages.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/SqsContainerOptions.java
alexisgra: I agree!

alexisgra:
Thanks for the review @tomazfernandes !
I've updated the doc. I remain available if other changes need to be made !

tomazfernandes:
Thanks a lot for the PR @alexisgra!

I did some polishing to speed up merging and opened a new PR: https://github.com/awspring/spring-cloud-aws/pull/1192, so I'll close this one.

Looking forward to more!

PR #1216, opened 2024-09-18
Upgrade AWS SDK to 2.28.2

maciejwalkowiak:


spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration/SqsIntegrationTests.java
```diff
@@ -188,7 +189,10 @@ void receivesMessageBatch() throws Exception {
 		List<Message<String>> messages = create10Messages("receivesMessageBatch");
 		sqsTemplate.sendMany(RECEIVES_MESSAGE_BATCH_QUEUE_NAME, messages);
 		logger.debug("Sent 10 messages to queue {}", RECEIVES_MESSAGE_BATCH_QUEUE_NAME);
-		assertThat(latchContainer.receivesMessageBatchLatch.await(10, TimeUnit.SECONDS)).isTrue();
+		await().untilAsserted(() -> {
```
maciejwalkowiak, line 192: @tomazfernandes this test was flaky as it is not guaranteed that each batch will contain exactly 10 messages. In my local setup i often received 1 or 2 messages in first batch so the latch never went down to 0.

PR #1229, opened 2024-09-22
Add client customizer interface for each supported integration.

### LINKED ISSUE #562: Make AwsClientCustomizer Flexible
**Type**: Feature

Imagine a developer wants to add a single request interceptor to a specific client. If I understand the whole thing correctly, the only way to do this is to create an `AwsClientCustomizer` bean that returns a `ClientOverrideConfiguration` instance in which the interceptor is added. This is a bit sub-optimal. The `ClientOverrideConfiguration` (COC) is huge one and I don't feel comfortable overriding the original instance (considering the fact that COC instance returned by the customizer replaces the original instance). Looking at the implementation, at least in AWSpring case, the COC instance is not an empty one, which means if I inject my bean, I have to duplicate the logic in `SpringCloudClientConfiguration`.

I think it makes sense to have more granular customizers, instead of a big fat one providing COC. (I'm fine with the rest of the existing customizer which are merely providers.)
Another alternative is to expect a `Consumer<ClientOverrideConfiguration.Builder>` instead of a ready-to-use instance of `COC`. A less optimal solution is to considering user-defined beans of type request interceptors.
Another totally different approach would be to expose the client builder in the customizer.

Right now I override the `SesClient` to get access to the client builder to inject my interceptor.

Let me know if you need more information.

maciejwalkowiak:
Fixes #562

- allows defining a customizer for each supported AWS client builder, for example, to configure `DynamoDbClientBuilder` user must define one or more `DynamoDbClientCustomizer`
- there can be multiple customizers for each builder, `@Order` is taken into consideration when applying customization
- to customize every SDK client (sync/async) - for example to configure a custom HTTP client applied to all SDK clients - users can define one or more `CommonAwsSyncClientCustomizer` (for sync) or `CommonAwsAsyncClientCustomizer` (for async) beans.

Example:

```java
@Bean
@Order(1)
DynamoDbClientCustomizer dynamoDbClientCustomizer() {
	return builder -> {
		builder.overrideConfiguration(builder.overrideConfiguration().copy(c -> {
			c.apiCallTimeout(Duration.ofMillis(2001));
		}));
	};
}

@Bean
@Order(2)
DynamoDbClientCustomizer dynamoDbClientCustomizer2() {
	return builder -> {
		builder.overrideConfiguration(builder.overrideConfiguration().copy(c -> {
			c.apiCallAttemptTimeout(Duration.ofMillis(2002));
		}));
	};
}

@Bean
CommonAwsSyncClientCustomizer commonAwsSyncClientCustomizer() {
	return builder -> {
		builder.httpClient(ApacheHttpClient.builder().connectionTimeout(Duration.ofMillis(1542)).build());
	};
}
```

### DIFF
--- a/docs/src/main/asciidoc/s3.adoc
+++ b/docs/src/main/asciidoc/s3.adoc
@@ -254,6 +254,33 @@ The Spring Boot Starter for S3 provides the following configuration options:
 | `spring.cloud.aws.s3.transfer-manager.follow-symbolic-links` | Specifies whether to follow symbolic links when traversing the file tree in `S3TransferManager#uploadDirectory` operation | No | `null` (falls back to SDK default)
 |===
 
+=== Client Customization
+
+`S3Client` can be further customized by providing a bean of type `S3ClientCustomizer`:
+
+[source,java]
+----
+@Bean
+S3ClientCustomizer customizer() {
+	return builder -> {
+		builder.overrideConfiguration(builder.overrideConfiguration().copy(c -> {
+			c.apiCallTimeout(Duration.ofMillis(1500));
+		}));
+	};
+}
+----
+
+[WARNING]
+====
+`builder.overrideConfiguration(..)` replaces the configuration object, so always make sure to use `builder.overrideConfiguration().copy(c -> ..)` to configure only certain properties and keep the already pre-configured values for others.
+====
+
+`S3ClientCustomizer` is a functional interface that enables configuring `S3ClientBuilder` before the `S3Client` is built in auto-configuration.
+
+There can be multiple `S3ClientCustomizer` beans present in single application context. `@Order(..)` annotation can be used to define the order of the execution.
+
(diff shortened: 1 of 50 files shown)


docs/src/main/asciidoc/s3.adoc
```diff
@@ -254,6 +254,33 @@ The Spring Boot Starter for S3 provides the following configuration options:
 | `spring.cloud.aws.s3.transfer-manager.follow-symbolic-links` | Specifies whether to follow symbolic links when traversing the file tree in `S3TransferManager#uploadDirectory` operation | No | `null` (falls back to SDK default)
 |===
 
+=== Client Customization
```
maciejwalkowiak, line 257: If this looks fine, similar docs will be added to each integration.

MatejNedic:
Really like the idea. Will follow up with detail Review but idea as it is I love it much better user experience

PR #1250, opened 2024-10-02
Fix setting timeout in seconds in `SqsTemplate`

joey:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
* Timeouts in SqsTemplate were being set by calling Duration#getSecondsPart which is always a number between 0 and 59
* Update to use Duration#getSeconds to get the full timeout specified, in seconds

This supports using a visibility time and polling timeout that's greater than 60 seconds.

Updated one of the unit tests for SqsTemplate

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
@@ -602,11 +602,11 @@ public class SqsTemplate extends AbstractMessagingTemplate<Message> implements S
 		ReceiveMessageRequest.Builder builder = ReceiveMessageRequest.builder().queueUrl(attributes.getQueueUrl())
 				.maxNumberOfMessages(maxNumberOfMessages).messageAttributeNames(this.messageAttributeNames)
 				.attributeNamesWithStrings(this.messageSystemAttributeNames)
-				.waitTimeSeconds(pollTimeout.toSeconds());
+				.waitTimeSeconds(toInt(pollTimeout.toSeconds()));
 		if (additionalHeaders.containsKey(SqsHeaders.SQS_VISIBILITY_TIMEOUT_HEADER)) {
 			builder.visibilityTimeout(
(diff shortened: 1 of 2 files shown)


maciejwalkowiak (approved):
LGTM but @tomazfernandes take a look in case there are any side effects i am not aware of

tomazfernandes:
Looks good! It might be worth it checking if ContainerOptions has the same problem though.

joey:
> Looks good! It might be worth it checking if ContainerOptions has the same problem though.

I didn't fine any other uses of the `to<Unit>Part` methods, so I think this the only place where this style of bug is present.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
```diff
@@ -616,6 +616,15 @@ private ReceiveMessageRequest doCreateReceiveMessageRequest(Duration pollTimeout
 		return builder.build();
 	}
 
+	// Convert a long value to an int. Values larger than Integer.MAX_VALUE are set to Integer.MAX_VALUE
+	private int toInt(long longValue) {
```
joey, line 620: I didn't want to use `Math.toIntExact()` as it will throw if the long value is out of the range and I thought pinning to the max integer would be better behavior. If we think it's unlikely enough that someone would set a timeout that high (over 63 years), then I'm happy to switch to the built in.

I didn't bother pinning too small of negative values as I don't think anyone would intentionally set a negative timeout.

MatejNedic: We can go with 2 approaches.
1. Yours by rounding to maximum integer there is, which is totally valid.

2. Or follow AWS SDK approach where we would (validate) throw exception with stating that value is above limit.

I personally would maybe want to go with approach number 2 just to be aligned, but both approaches are totally valid.
But tbh we are now nitpicking since no one will probably want or will set timeout that big. So for me this is fine!

MatejNedic (approved):

maciejwalkowiak:
Thanks @joey !

PR #1300, opened 2024-12-17
Set version to 3.3.0-RC1

MatejNedic:


maciejwalkowiak:
I think you meant 3.3.0-RC1 :-)

MatejNedic:
That is why I shouldn't do things tired 😅 . FIxed

Issue #1272, opened 2024-11-11
SnsNotificationMessage variant for EventBridge


frjonsen:
**Type**: Feature

**Is your feature request related to a problem? Please describe.**
In our projects we've opted for EventBridge rather than SNS. EventBridge also has a wrapper, similar to that of SNS, but some different fields. Currently we are unwrapping this message to get the to the actual data in our own in-house framework, but feel that there could be a case for including a variant of the existing [SnsNotificationMessage](https://github.com/awspring/spring-cloud-aws/blob/d59ca4ae58ba4b757d3ff5afd783134302064d0a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/SnsNotificationMessage.java#L4) annotation, but for the EventBridge wrapper.

**Describe the solution you'd like**
A new annotation, such as `EventBridgeMessage`, as well as a corresponding message converter, similar to [SnsMessageConverter](https://github.com/awspring/spring-cloud-aws/blob/main/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SnsMessageConverter.java).

**Describe alternatives you've considered**
Defining a model like `record EventBridgeMessage<T>(T detail) { }`.

tomazfernandes:
Hey @frjonsen, yep, I think this does make sense.

If you could provide a prototype PR with an integration test so we can have a clearer notion of how that looks that'd be great.

Thanks!

PR #1307, opened 2024-12-19
SQS Listener EventBridgeMessage annotation and converter

frjonsen:

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Core goal was to add a new annotation, @EventBridgeMessage, as well as the corresponding converter and argument resolver. I took the liberty of splitting out common functionality `SnsMessageConverter` to an abstract `WrappedMessageConverter` to reduce duplication.

Further context in this issue: https://github.com/awspring/spring-cloud-aws/issues/1272

Added tests similar to those of the @SnsNotificationMessage. The test resource `eventBridgeMessage` is taken partially from https://docs.aws.amazon.com/AmazonS3/latest/userguide/ev-events.html, though the payload was written by hand to match the POJO.

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -639,6 +639,29 @@ public void listen(@SnsNotificationMessage List<Pojo> pojos) {
 }
 ----
 
+==== EventBridge Messages
+
+Since 3.3.0, similar to the @SnsNotificationMessage, there is also a @EventBridgeMessage annotation that can be used to receive EventBridge messages. To only receive the `detail` part of the payload, you can utilize the `@EventBridgeMessage` annotation.
+As with @SnsNotificationMessage, this supports both individual messages, or lists.
+
+[source, java]
+----
+@SqsListener("my-queue")
+public void listen(@EventBridgeMessage Pojo pojo) {
+	System.out.println(pojo.field);
+}
+----
+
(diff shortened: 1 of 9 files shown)


frjonsen:
I was unsure what to fill in as version, in the places they were mentioned. I wrote 3.3.0 as that seemed to be the next unreleased version. Should this be changed to something else, or will it be updated once it is determined which version this would be included in?

tomazfernandes:
Hey @frjonsen, sorry for the delay.

We've just merged a PR that changes a few things in the SnsNotification part of the project, if you can please rebase your branch and see what needs to be changed.

For the overall design I have two suggestions.

1. If I understand correctly, the Wrapped abstract class will define the conversion hint when it's a list. How about we instead of the abstract class, we use an internal `SmartMessageConverter` to delegate to when we figure out the list's type? This way we'd be able to use it with any other message converters when it's a list. Makes sense?

2. This is optional, but how about we create a `EventBridgeMessage<MyPojo>` class that would contain both the deserialized POJO and the event bridge metadata? We could use it internally instead of manipulating the JSON, and users could have access to all metadata if they want. So we'd be able to use both `@EventBridgeMessage MyPojo` and `@EventBridgeMessage EventBridgeMessage<MyPojo>`, and their list variants.

Thanks and please let me know your thoughts.

PR #1308, opened 2024-12-20
Dynamically configure SemaphoreBackPressureHandler with BackPressureLimiter (#1251)

loicrouchon:
## :loudspeaker: Type of change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

This change enhances the `SemaphoreBackPressureHandler` with the support of a new `BackPressureLimiter` interface.

This `BackPressureLimiter` interface is to be implemented by applications. It has a single method `int limit()` that returns a number of permits that can be consumed by the `SemaphoreBackPressureHandler`.

Before each polling, the limit will be checked by the `SemaphoreBackPressureHandler` and adjust the number of permits that can be requested (in the range `[0, totalPermits]`). The limit returned by the `BackPressureLimiter#limit()` method is to be understood in terms of number of messages that can be consumed from the queue at the current instant. If `0` (or less) the queue consumption is considered to be on standby.

When a polling is performed and the consumption is on standby, the `SemaphoreBackPressureHandler` will sleep the `standbyLimitPollingInterval` before allowing for a next polling attempt (we cannot rely on the semaphore acquire timeouts here, hence the need for `standbyLimitPollingInterval`.

Both the `BackPressureLimiter` and `standbyLimitPollingInterval` can be configured via the `ContainerOptions`

## :bulb: Motivation and Context

The goal of this change is to address #1251.

#1251 aims to provide a more general solution to issues like #481 by giving control to users on how they would like to dynamically limit messages consumption from an SQS queue. Typical use cases could be rate-limiters (like #481) or more complicated setups involving measuring the load of a downstream system and adjusting or stopping the messages consumption

## :green_heart: How did you test it?

The testing was so far only tested via integration tests which tested various scenarios:

- static limit being 0, totalPermits or more, and various values in between.
- Dynamic limit changing, and checking the queue processing adjusting itself (in a "synchronized" way)
- Dynamic limit changing, and checking the queue processing adjusting itself (not in a synchronized way, so more realistic of what will happen, but harder to make strict assumptions on the moving target)

## :pencil: Checklist

- [x] I reviewed submitted code
- [x] I added tests to verify changes
- [ ] I updated reference documentation to reflect the change
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
* Collecting feedback on the approach
* Work on updating the reference documentation to mention the new capability.

Issue #1251, opened 2024-10-03
SQS: Add support for external back-pressure management


loicrouchon:
**Type**: Feature

**Is your feature request related to a problem? Please describe.**
The current SQS listener implementation has options to limit the concurrency within the application, however, it does not support an external data-source informing about downstream services back-pressure.

The use case I have in mind is to stop processing new messages from an upstream queue if a downstream queue reached a certain number of visible messages.
In our use case, several producers are publishing messages to a downstream queue. Some producers are less critical than others and should only publish in messages in the downstream queue at low peak times. For those producers we need to be informed of back-pressure of the downstream queue to decide if we should produce new messages (i.e. consume from the upstream queue).

Another use case would be the one described in https://github.com/awspring/spring-cloud-aws/issues/481

**Describe the solution you'd like**
The solution I'm thinking of would be to implement a hook for reporting downstream services back-pressure into some `BackPressureHandler`/`BatchAwareBackPressureHandler` implementation.
I would put out of scope the implementation of said hooks (downstream queue number of visible messages, rate limiter, ...) which would be (at least for now) to be implemented by applications.

I can propose to implement such feature, but before doing so, I'd like to discuss the direction.

So far I've seen there's only one implementation of the `BatchAwareBackPressureHandler`: the `SemaphoreBackPressureHandler`.
My idea is to build something around it (either a wrapper, a sub-class, or evolving the current implementation with more features, we'll see when going in the implementation phase).
The reason for keeping the initial `SemaphoreBackPressureHandler` is that it limits the concurrency within the application, while the downstream back-pressure concept is there to stop processing or reduce processing speed if downstream systems cannot follow.

So this new implementation would be a kind of `SemaphoreBackPressureHandler` which would call a `BackPressureMeter` in `request(int)` and `requestBatch()` methods. Depending on the current downstream back-pressure and the configured `maxBackPressure` it would accordingly increase (up to `totalPermits`) or reduce (down to `0`) the number of permits on the internal semaphore. It might also be needed to update the `drain(Duration)` method to take the reduced capacity into account.

New methods would be required on the `ContainerOptions` interface:

* `#downstreamBackPressureMeter(BackPressureMeter meter)`: specifies how to measure the back-pressure of downstream systems.
* `#maxDownstreamBackPressure(int maxDownstreamBackPressure)`: specifies what is the max back-pressure that can be put on the downstream system.

The `BackPressureMeter` interface could look like this.

```java
public interface BackPressureMeter {

[...]

PR #1316, opened 2025-01-01
Introduction of RateLimitSemaphoreBackPressureHandler

jeroenvandevelde:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Related to #123, this is purely to show the concept.
Still need to add tests, polish it, ...

It enables the ability to read messages from a queue with a max speed / second for rate limiting purposes.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
- [ ] Need to add documentation
- [ ] Need to verify the open TODOs
- [ ] Polish the code


tomazfernandes:
Hey @jeroenvandevelde, thanks for the PR!

One question - how is the `TimedSemaphore` different from Guava's `RateLimiter` and why did you choose this one?

I think a design goal we should keep in mind both for this PR and for https://github.com/awspring/spring-cloud-aws/pull/1308 is that this is an inherently complex area of the solution, so we should make an effort to keep complexity at check.

For this PR, this might mean trying to reuse some of the original `BackPressureHandler` logic where appropriate rather than duplicating it, perhaps extracting some of the logic to a third class, or any other approach that makes sense.

Also, I'm not sure if in your discussion with @loicrouchon you came to a conclusion whether rate limiting might be a special case of his proposal, and in that case if we should try to come up with a single abstraction to enhance back pressure capabilities.

Please let me know your thoughts, thanks.

jeroenvandevelde:
Hi @tomazfernandes 

> One question - how is the `TimedSemaphore` different from Guava's `RateLimiter` and why did you choose this one?

I picked the `TimedSemaphore` to stick as close as possible to the current implementation with [Semaphores](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/Semaphore.html).
I haven't done an in-depth check which one makes the most sense for this use-case.

I made this PR purely to show the concept, but i will take a look at both and compare.
Any preference for either the `TimedSemaphore` from apache or Guava's `RateLimiter` from your end?

> I think a design goal we should keep in mind both for this PR and for #1308 is that this is an inherently complex area of the solution, so we should make an effort to keep complexity at check.
> 
> For this PR, this might mean trying to reuse some of the original `BackPressureHandler` logic where appropriate rather than duplicating it, perhaps extracting some of the logic to a third class, or any other approach that makes sense.

Completely agree on the fact that this code needs some love, by either extracting another class and share some logic or ...
I just made this PR to showcase the idea, so i will definitely improve the design once we agree on a way forward.

> Also, I'm not sure if in your discussion with @loicrouchon you came to a conclusion whether rate limiting might be a special case of his proposal, and in that case if we should try to come up with a single abstraction to enhance back pressure capabilities.

In the discussion with @loicrouchon, i pointed out that we are trying to solve two separate challenges of rate limiting (dynamic and static).

_Static_ - this one - is where the rate limit stays the same and the _dynamic_ one - [#1308](https://github.com/awspring/spring-cloud-aws/pull/1308) - where the rate limit is dynamically changed based on events that happen outside of the library.

In _static_ rate limiting there is a need to keep track of how many messages have been requested in a certain timeframe. This is necessary to know if you go over the rate limit with requesting 5 more messages or not. 

In _dynamic_ rate limiting, you could also keep track of how many messages have been requested in a certain timeframe but it is not always needed. In [#1308](https://github.com/awspring/spring-cloud-aws/pull/1308) for example it is not done as @loicrouchon made it so that based on external input, the library should retrieve 0 x the normal amount of messages, 0,5 x the normal amount of messages or ...

I would like to keep the complexity of how many messages have been requested in a certain timeframe in the library so that the users don't need to implement this every time they need this.

So i am not certain a single abstraction will cover both situation or at least i don't see one immediately. 😊
This doesn't mean we shouldn't align both solutions as much as possible.

tomazfernandes:
Hey @jeroenvandevelde, I'm thinking we should wait for #1251 to be merged before moving ahead with this one since we might end up introducing new elements you could use here.

Please let me know if that makes sense to you.

jeroenvandevelde:
> Hey @jeroenvandevelde, I'm thinking we should wait for #1251 to be merged before moving ahead with this one since we might end up introducing new elements you could use here.
> 
> Please let me know if that makes sense to you.

Sure, this makes sense to me. 
I will follow up #1251 more closely again in the short future, should gain some more free time again soon.

tomazfernandes:
Hey @jeroenvandevelde, #1251 has been merged and now we can support a dedicated RateLimiterBackPressureHandler in conjunction with other BackPressure implementations by using a `CompositeBackPressureHandler`.

This should allow you to simplify your implementation and keep it scoped to `RateLimiter` concerns.

Let me know if it makes sense to you and if you have any thoughts or questions.

Thanks!

jeroenvandevelde:
> Hey @jeroenvandevelde, #1251 has been merged and now we can support a dedicated RateLimiterBackPressureHandler in conjunction with other BackPressure implementations by using a `CompositeBackPressureHandler`.
> 
> This should allow you to simplify your implementation and keep it scoped to `RateLimiter` concerns.
> 
> Let me know if it makes sense to you and if you have any thoughts or questions.
> 
> Thanks!

Thanks @tomazfernandes I will take a look at using a `CompositeBackPressureHandler`, and try to optimize the implementation.

PR #1318, opened 2025-01-10
Add SQSListener argument resolver to extract the subject of an SNS me…

n-n-nebbl:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Support extracting of SNS subjects in SQS listeners using an annotation on a String parameter

I looked at the following PR and made the change accordingly to the current state of the code added there:
https://github.com/awspring/spring-cloud-aws/pull/898

This functionality was there in the old spring-cloud-aws library (2.x) and is making transitioning to the new library easier.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -639,6 +639,16 @@ public void listen(@SnsNotificationMessage List<Pojo> pojos) {
 }
 ----
 
+Since 3.3.0 you can also retrieve the subject of the SNS message through the `@SnsNotificationSubject` annotation.
+
+[source, java]
+----
+@SqsListener("my-queue")
+public void listen(@SnsNotificationSubject String subject) {
+	System.out.println(subject);
+}
+----
+
 ===== Specifying a MessageListenerContainerFactory
 A `MessageListenerContainerFactory` can be specified through the `factory` property.
 Such factory will then be used to create the container for the annotated method.
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -639,13 +639,13 @@ public void listen(@SnsNotificationMessage List<Pojo> pojos) {
 }
 ----
 
-Since 3.3.0 you can also retrieve the subject of the SNS message through the `@SnsNotificationSubject` annotation.
+Since 3.3.1 you can also retrieve the subject of the SNS message through the `@SnsNotificationSubject` annotation.
 
 [source, java]
 ----
 @SqsListener("my-queue")
-public void listen(@SnsNotificationSubject String subject) {
-	System.out.println(subject);
+public void listen(@SnsNotificationSubject String subject, @SnsNotificationMessage Pojo pojo) {
+	System.out.println("received message for subject %s: %s".formatted(subject, pojo));
 }
 ----
 
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SnsJsonNode.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SnsJsonNode.java
@@ -22,7 +22,7 @@ import org.springframework.messaging.converter.MessageConversionException;
 /**
  * @author Michael Sosa
  * @author Alexander Nebel
- * @since 3.3.0
+ * @since 3.3.1
  */
 public class SnsJsonNode {
 	private final String jsonString;
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SnsJsonNode.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SnsJsonNode.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2013-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.awspring.cloud.sqs.support.converter;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.springframework.messaging.converter.MessageConversionException;
+
+/**
+ * @author Michael Sosa
+ * @author Alexander Nebel
+ * @since 3.3.0
+ */
+public class SnsJsonNode {
+	private final String jsonString;
+	private final JsonNode jsonNode;
+
+	public SnsJsonNode(ObjectMapper jsonMapper, String jsonString) {
+		try {
+			this.jsonString = jsonString;
+			jsonNode = jsonMapper.readTree(jsonString);
+		}
+		catch (Exception e) {
+			throw new MessageConversionException("Could not read JSON", e);
+		}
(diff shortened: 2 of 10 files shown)


n-n-nebbl:
It's the first time for me contributing to a bigger open-source project, so please tell me if I made some mistakes (also in the process of contributing besides adding the code)

If you agree that this functionality should be added, I will adding some tests and also have a look if I can make the code together with the SnsMessage PR having less duplicates.

@tomazfernandes Please have a look when you have time, so I know if I should put more effort into this.

n-n-nebbl:
just a bump to maybe get an answer. maybe from one of you @maciejwalkowiak or @MatejNedic

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SnsJsonNode.java
```diff
@@ -0,0 +1,67 @@
...
+			throw new MessageConversionException("Payload: '" + jsonString + "' does not contain a message", null);
+		}
+	}
+
+	public String getMessageAsString() {
+		return jsonNode.get("Message").asText();
+	}
+
+	public String getSubjectAsString() {
+		if (!jsonNode.has("Subject")) {
```
n-n-nebbl, line 62: Do SnsNotifications always have a `Subject`?
Then this could be moved to the `validate()` method.

n-n-nebbl: When sending a SNS message through AWS Console it is optional (https://docs.aws.amazon.com/sns/latest/dg/sns-publishing.html#publish-to-topic-aws-sdks)
In the `PublishRequest` class of the aws sdk sns package it is also described as optional in the javadoc.

So I think it is on the correct place already.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SnsSubjectConverter.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * @author Alexander Nebel
+ * @since 3.3.0
+ */
+public class SnsSubjectConverter implements MessageConverter {
+
+	private final ObjectMapper jsonMapper;
+
+	public SnsSubjectConverter(ObjectMapper jsonMapper) {
```
tomazfernandes, line 35: Nit: We usually just name it `objectMapper`. The variable name is not an issue, but perhaps the error message would be clearer?

docs/src/main/asciidoc/sqs.adoc
```diff
@@ -639,6 +639,16 @@ public void listen(@SnsNotificationMessage List<Pojo> pojos) {
 }
 ----
 
+Since 3.3.0 you can also retrieve the subject of the SNS message through the `@SnsNotificationSubject` annotation.
```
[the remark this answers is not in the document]
tomazfernandes: Also, perhaps we should make it clear that lists are not supported?

BTW, any particular reason not to support Lists from the start?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/resolver/NotificationSubjectArgumentResolver.java
```diff
@@ -0,0 +1,71 @@
...
+		this.converter = new SnsSubjectConverter(jsonMapper);
+	}
+
+	@Override
+	public boolean supportsParameter(MethodParameter parameter) {
+		if (parameter.hasParameterAnnotation(SnsNotificationSubject.class)) {
+			if (ClassUtils.isAssignable(parameter.getParameterType(), String.class)) {
+				return true;
+			}
+			if (logger.isWarnEnabled()) {
```
tomazfernandes, line 51: AFAIK modern logging implementations do not execute any logic e.g. placeholder resolution / String concatenation if the corresponding log level is not enabled, so I tend to favor cleaning up the code and not checking if the log level is enabled myself.

Just a nit though.

tomazfernandes (commented):
Looks great @n-n-nebbl, thanks for the PR! 

Just left a few nits / suggestions for your consideration, but none of them are blockers.

docs/src/main/asciidoc/sqs.adoc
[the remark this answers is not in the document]
n-n-nebbl: I just couldn't imagine the use case.
How would this look like?

```
void listen(@SnsNotificationSubject List<String> subjects, @SnsNotificationMessage List<String> messages) {
  ...
}
```
how to match the subject with the message then?
If this feature is needed I can implement it.

BTW: is the version correct? should it be 3.3.1?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/resolver/NotificationSubjectArgumentResolver.java
n-n-nebbl: static code analysis was complaining because I am calling `getMethodName(parameter)` even if the logging level is not activated which is true, but could also be ignored

docs/src/main/asciidoc/sqs.adoc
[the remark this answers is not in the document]
tomazfernandes: Well, in theory it should be `3.4` since it's not a bugfix, but I think `3.3.1` is the safe bet. Unless @maciejwalkowiak has a different idea?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/resolver/NotificationSubjectArgumentResolver.java
tomazfernandes: Yeah, I guess that's a fair point.

Ideally `slf4j` could provide a method signature that received a lambda which would only be executed if logging was enabled at the proper level, but I don't think it's an option at this point.

We can leave it as is.

docs/src/main/asciidoc/sqs.adoc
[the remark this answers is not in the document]
tomazfernandes: > I just couldn't imagine the use case.
> How would this look like?

Yup, fair point. We'd probably need a `Map<UUID, String>` instead of the list so it's possible to correlate, but I'm happy to leave this out of this PR and see if anyone requests that before implementing it.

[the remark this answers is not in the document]
maciejwalkowiak: Unless its a breaking change it is safe to put 3.3.1

n-n-nebbl:
From my point of view every comment is resolved and the PR is ready to get merged.

tomazfernandes (approved):
Thanks @n-n-nebbl, looking forward to more!

There are quite a few interesting SQS open issues if you're interested, please let me know if you'd like to pick any one up.

tomazfernandes:
Just one late thought @n-n-nebbl - it might be interesting for us to have a `SnsMessage<MyType>` POJO that would contain both the deserialized message and the associated metadata from SNS.

So users could have either `@SnsNotificationMessage SnsMessage<MyPojo>` or `@SnsNotificationMessage List<SnsMessage<MyPojo>>`.

This way we could support getting a list of messages with their subjects and other metadata. We could also use it with the `ObjectMapper` internally instead of operating directly on JSONs.

Please let me know your thoughts and if you'd be interested in providing a PR for that. Thanks!

n-n-nebbl:
> Just one late thought @n-n-nebbl - it might be interesting for us to have a `SnsMessage<MyType>` POJO that would contain both the deserialized message and the associated metadata from SNS.
> Please let me know your thoughts and if you'd be interested in providing a PR for that. Thanks!

That would be a nice solution!
I can provide a PR for it, but I do not know yet when I will have time for it so it could take some weeks.

n-n-nebbl:
> Thanks @n-n-nebbl, looking forward to more!
> 
> There are quite a few interesting SQS open issues if you're interested, please let me know if you'd like to pick any one up.

I will have a look at them the next time I can spare some time coding :)

PR #1322, opened 2025-01-16
Lift ByteBuddy version to 1.15.11

MatejNedic:


maciejwalkowiak:
Not valid as we don't manage ByteBuddy version anymore.

Issue #595, opened 2022-12-21
Add Support for SES V2


akefirad:
**Type**: Feature

**Is your feature request related to a problem? Please describe.**
There's a new SES API, version 2. It would be nice to have this support as well.

Happy to help with a PR.

maciejwalkowiak:
Can you describe in more details what functionality do you have in mind?

akefirad:
So there's a new [SDK](https://mvnrepository.com/artifact/software.amazon.awssdk/sesv2) for [SES API V2](https://docs.aws.amazon.com/ses/latest/APIReference-V2/Welcome.html) which provides more functionalities (including more policy statement conditions). [Here](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/java_sesv2_code_examples.html)'s a some examples from AWS. Does that help?

maciejwalkowiak:
I'm still not sure how this will affect the functionality offered by Spring Cloud AWS, or is it more about autoconfiguring the client? Feel free to drop a PR perhaps it gets more clear then

akefirad:
Imagine if I want to use version 2 (because of some specific feature available only in version 2) of SES client, I cannot use AWSpring SES module since this module is written on top of SES version 1 and won't work with version 2. Does that make sense? As for the PR, the question before attempting is how we want to support it? Duplicate the whole library? Or introduce an interface that abstract away the details of SES clients, V1 and V2. Or any other solution.

maciejwalkowiak:
I see this making sense if we just drop SES v1 and replace it with SES v2. Are there any drawbacks of using SES v2?

akefirad:
Hm good point. Since this is a major release, it wouldn't break existing code. The only thing I'm not sure is what you asked, is there anything we miss if we go with version2 altogether. I have to check. Would be nice if you have any connection with anyone on AWS side. (I'll try to dig into the docs anyway.)

maciejwalkowiak:
@msailes are you aware of any drawbacks of switching to SES v2 or is it just a natural step and v1 should be considered as deprecated?

akefirad:
As far as I can tell, they are not compatible, neither backward nor forward. Unless we hear something otherwise from Mark, my guess is we have to support both. In the mean time, what do you think about having three interfaces and one implementation for each version:
```java
public interface SendMessageResult {
  String getMessageId();
}

public interface AwsSesSimpleMessageSender {
  SendMessageResult send(SimpleMailMessage message);
}


[...]

PR #1336, opened 2025-01-31
#595 migrate to sesv2

dominik-kovacs:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
Migration of AWS SES from v1 to v2

Amazon SES v2 API and SMTP accepts email messages up to 40MB in size including any images and attachments that are part of the message. Messages larger than 10MB are subject to bandwidth throttling, and depending on your sending rate, you may be throttled to as low as 40MB/s. For example, you could send a 40MB message at the rate of 1 message per second, or two 20MB messages per second.

Amazon SES API v1 accepts messages up to 10MB in size including any images and attachments that are part of the message.

Issue #595

Unit tests
I don't have localstack pro, so no integration tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/docs/src/main/asciidoc/ses.adoc
+++ b/docs/src/main/asciidoc/ses.adoc
@@ -135,8 +135,7 @@ The Spring Boot Starter for SES provides the following configuration options:
 | `spring.cloud.aws.ses.enabled` | Enables the SES integration. | No | `true`
 | `spring.cloud.aws.ses.endpoint` | Configures endpoint used by `SesClient`. | No |
 | `spring.cloud.aws.ses.region` | Configures region used by `SesClient`. | No |
-| `spring.cloud.aws.ses.source-arn` | Configures source ARN, used only for sending authorization. | No |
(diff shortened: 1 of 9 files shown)


maciejwalkowiak:
@dominik-kovacs thanks for a PR. To avoid breaking changes, we need to keep the current integration working. Perhaps it's makes sense to have a new module for v2? @MatejNedic any opinions?

dominik-kovacs:
The only breaking change I encountered was merging `sourceArn` and `fromArn` to a single parameter `fromEmailAddressIdentityArn` https://repost.aws/en/questions/QUjSyC3Az8TmeMc3_D5CcVfg#ANZRULQLHKTEmZj1WllxW6Rw

we could keep all 3 and deprecate sourceArn and fromArn

+ I'm not sure if this is a breaking change but ses v2 is only supported in localstack pro.

MatejNedic:
I would move it to a new module as well. If someone does not want to migrate sdk, it is better if you ask me. 

This is major change lets wait few versions before deprecating old module. 

How does this sound?

PR #1356, opened 2025-03-09
Add module for SES v2

rmschots:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Add a module to allow use of SES API v2.

Amazon SES API v1 accepts messages up to 10MB in size including any images and attachments that are part of the message. SES API v2 accepts up to 40MB.
Issue https://github.com/awspring/spring-cloud-aws/issues/595
Alternative solution to https://github.com/awspring/spring-cloud-aws/pull/1336.
This solution guarantees no breaking changes with previous versions, and allows using the v2 API instead.

- Same tests as the v1 API, as the functionality is the same.
- Manually tested a new sample application, using localstack pro, as only the pro version supports ses v2.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -32,6 +32,8 @@
     - any-glob-to-any-file:
       - spring-cloud-aws-autoconfigure/src/*/java/io/awspring/cloud/autoconfigure/ses/*
       - spring-cloud-aws-ses/**/*
+      - spring-cloud-aws-autoconfigure/src/*/java/io/awspring/cloud/autoconfigure/sesv2/*
+      - spring-cloud-aws-sesv2/**/*
 "component: sns":
   - changed-files:
     - any-glob-to-any-file:
--- a/docs/src/main/asciidoc/_configprops.adoc
+++ b/docs/src/main/asciidoc/_configprops.adoc
@@ -85,6 +85,12 @@
 |spring.cloud.aws.ses.from-arn |  | Configures from ARN. Only applies to SendRawEmail operation.
 |spring.cloud.aws.ses.region |  | Overrides the default region.
 |spring.cloud.aws.ses.source-arn |  | Configures source ARN. Used only for sending authorization.
+|spring.cloud.aws.sesv2.configuration-set-name |  | Configures configuration set name.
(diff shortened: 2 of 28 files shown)


maciejwalkowiak:
Thanks @rmschots for a PR. Perhaps it could be improved to reduce the duplication, as these classes from V1 and V2 are pretty much the same with small differences. Perhaps both V1 and V2 could live in single module with optional dependencies to SES SDKs, which would allow us to have an abstract base classes for two versions, and then there would be two starters for V1 and V2 that would bring related SDKs? What do you think about it? cc @MatejNedic

huilinn93:
@maciejwalkowiak I implemented the separate module based on the feedback on https://github.com/awspring/spring-cloud-aws/pull/1336#issuecomment-2629096457.
Maybe best that there is a consensus on the solution before people create another PR.

rmschots:
@maciejwalkowiak I implemented the separate module based on the feedback on https://github.com/awspring/spring-cloud-aws/pull/1336#issuecomment-2629096457.
Maybe best that there is a consensus on the solution before people create another PR.

MatejNedic:
@maciejwalkowiak Personally I would prefer duplication since SDK is different. We can slowly than deprecate V1 version much cleaner, specific feature and implementations can be handled much better and so on.

I would go with split. Let me take a look in code and think about it, will make a review in following days.

rmschots:
@MatejNedic Maybe time to look at this one again? It's been a while since it has been forgotten.

MatejNedic:
I will take a look in following days.

@maciejwalkowiak if all is good we can target it for 4.0.0?

MatejNedic:
Since it is a new feature and they broke api I would like to either 

1.  In 4.0.0 go with breaking change and migration guide.
2. Create new module if people want to use V2 api


Since V2 is not really bringing that much to a table, at least not for our integration, I would go with an option 2. 

If someone is interested in changes V2 brings-> 
https://aws.amazon.com/blogs/messaging-and-targeting/upgrade-your-email-tech-stack-with-amazon-sesv2-api/

MajnunMan:
Hey guys, thanks for the amazing work! Really appreciate it.
 
We really need this module. 

Can you give us a timeline for when this PR will be merged? We really need it 🙏

Xyaren:
@maciejwalkowiak  Any news here ? 
We need support vor v2 in order to use the tenant functionality, which is only supported in v2 of the aws ses api.

Xyaren:
@maciejwalkowiak  This is assigned to the 4.0.0 milestone but still not incuded.
Can the community somehow help with this ?

Issue #1373, opened 2025-04-20
Introduce "FilteringSinkAdapter" in SQS integration


MatejNedic:
Sometimes when using `@SqsListener()` I would like to filter the messages before entering the method.

Good example is when messages have ttl (Time-to-live) or when we are no longer interested in certain events due business logic. Instead of polluting and replicating this logic every time in `@SqsListener()`  it would be better if we could extract it outside in `FilteringSinkAdapter`.

This way our integration would provide users a clean way to filter and choose which message is processed and which is not.

imsosleepy:
Hi @MatejNedic , I'm interested in working on this issue.
Before jumping into the implementation, I wanted to check in and get some clarification on the intended direction.

I currently have two possible approaches in mind:

1. Creating a filter interface and allowing users to register custom filters as beans, which can then be applied before processing the message. 

2. Adding a new option to @SqsListener, such as @SqsListener(filterStrategy = "ttlFilter"), to specify the filter in a more declarative way.

I initially started working on the first approach, but I realized that the second one could also be effective, especially for ease of configuration.

Would you be able to share your thoughts or suggest a preferred direction?

Thanks in advance!

imsosleepy:
> Hi [@MatejNedic](https://github.com/MatejNedic), [@tomazfernandes](https://github.com/tomazfernandes) , I'm interested in working on this issue. Before starting the actual implementation, I wanted to check in and get some clarification on the intended direction.
> 
> I currently have two possible approaches in mind:
> 
> 1. Creating a filter interface and allowing users to register custom filters as beans, which can then be applied before processing the message. [example](https://github.com/imsosleepy/spring-cloud-aws/commit/9f355e1665f3ea5a66e446c4204c3a98379451ea)
> 2. Adding a new option to @SqsListener, such as @SqsListener(filterStrategy = "ttlFilter"), to specify the filter in a more declarative way.
> 
> I initially started working on the first approach, but I realized that the second one could also be effective, especially for ease of configuration.
> 
> Would you be able to share your thoughts or suggest a preferred direction?

I realized that if I go with the first approach, the filter would be configured globally, which means that multiple listeners wouldn’t be able to have their own individual filters. This makes it less suitable. I'll proceed with the second approach and submit a PR.

tomazfernandes:
Hi @imsosleepy, thanks for looking into this!

There are a few things for us to consider.

First is what we need from the interface - we might want to have both a single message and a batch message methods.

[...]

PR #1388, opened 2025-05-20
Add FilteringAdapter in SQS

imsosleepy:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Added support for filtering SQS messages before they reach the listener using a MessageFilter.

The filter can be configured through SqsMessageListenerContainerFactory, and only messages that match the condition will be passed to the listener.

Previously, all incoming messages were forwarded to the listener, requiring manual filtering within business logic.

This change enables declarative filtering at the container level, helping reduce unnecessary processing and keeping listener logic clean.

Added integration test SqsMessageFilterIntegrationTests to verify that filtered messages are not delivered to the listener.

Confirmed that only messages satisfying the filter are passed through and acknowledged.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

This initial implementation supports filtering on a single message basis only, and does not yet account for batch or asynchronous filtering scenarios.

As suggested in [#1373](https://github.com/awspring/spring-cloud-aws/issues/1373), we plan to extend support to:

- Batch message filtering
- Asynchronous filtering
- Support for filtering based on fields other than the message payload
(e.g., message attributes, custom headers, or system metadata)


spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/config/AbstractMessageListenerContainerFactory.java
```diff
@@ -16,26 +16,22 @@
...
 import io.awspring.cloud.sqs.listener.acknowledgement.AcknowledgementResultCallback;
 import io.awspring.cloud.sqs.listener.acknowledgement.AsyncAcknowledgementResultCallback;
 import io.awspring.cloud.sqs.listener.errorhandler.AsyncErrorHandler;
 import io.awspring.cloud.sqs.listener.errorhandler.ErrorHandler;
 import io.awspring.cloud.sqs.listener.interceptor.AsyncMessageInterceptor;
 import io.awspring.cloud.sqs.listener.interceptor.MessageInterceptor;
+import io.awspring.cloud.sqs.support.filter.DefaultMessageFilter;
+import io.awspring.cloud.sqs.support.filter.MessageFilter;
+import org.springframework.messaging.Message;
+import org.springframework.util.Assert;
```
injae-kim, line 29: maybe your import order lint is different with this project? can you check?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/config/AbstractMessageListenerContainerFactory.java
```diff
@@ -180,6 +176,7 @@ public C createContainer(Endpoint endpoint) {
 		B options = this.containerOptionsBuilder.createCopy();
 		configure(endpoint, options);
 		C container = createContainerInstance(endpoint, options.build());
+
```
injae-kim, line 179: can we remove this unnecessary empty line?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/AbstractContainerOptions.java
```diff
@@ -59,6 +62,9 @@ public abstract class AbstractContainerOptions<O extends ContainerOptions<O, B>,
 
 	private final AcknowledgementMode acknowledgementMode;
 
+	private final MessageFilter<?> messageFilter;
+
+
```
injae-kim, line 67: can we remove this unnecessary empty line L67?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/AbstractContainerOptions.java
```diff
@@ -164,6 +172,9 @@ public MessagingMessageConverter<?> getMessageConverter() {
 		return this.messageConverter;
 	}
 
+	@Override
+	public MessageFilter<?> getMessageFilter() {return this.messageFilter; }
```
injae-kim, line 176:
```suggestion
	public MessageFilter<?> getMessageFilter() {
	    return this.messageFilter;
	}
```
nit. indent

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/AbstractContainerOptions.java
```diff
@@ -400,6 +414,14 @@ public B messageConverter(MessagingMessageConverter<?> messageConverter) {
...
 		}
 
+		@Override
+		public B messageFilter(MessageFilter<?> messageFilter) {
+			Assert.notNull(messageFilter, "messageFilter cannot be null");
+			this.messageFilter = messageFilter;
+			return self();
+		}
+
+
```
injae-kim, line 424: can we remove this unnecessary empty line?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/AsyncComponentAdapters.java
```diff
@@ -214,6 +220,44 @@ public CompletableFuture<Void> onMessage(Collection<Message<T>> messages) {
...
+			this.filter = filter;
+		}
+
+		@Override
+		public CompletableFuture<Void> onMessage(Message<T> message) {
+			if (filter.process(message)) {
+				return execute(() -> this.filteredMessageListener.onMessage(message));
+			}
+			else {
+				logger.debug("Message filtered out: {}", message.getPayload());
```
injae-kim, line 240: can we remove this debug log? same on L252 :)

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/ContainerOptions.java
```diff
@@ -139,6 +141,11 @@ default BackOffPolicy getPollBackOffPolicy() {
 	 */
 	MessagingMessageConverter<?> getMessageConverter();
 
+	/** Return the message filter applied before message processing.
+	 *  @return the message filter.
+	 * */
```
injae-kim, line 146:
```suggestion
	/** 
	 *  Return the message filter applied before message processing.
	 *  @return the message filter.
	 */
```
nit, indent

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/AsyncComponentAdapters.java
```diff
@@ -76,6 +78,10 @@ public static <T> AsyncMessageListener<T> adapt(MessageListener<T> messageListen
 		return new BlockingMessageListenerAdapter<>(messageListener);
 	}
 
+	public static <T> AsyncMessageListener<T> adaptFilter(MessageListener<T> messageListener, MessageFilter<T> messageFilter) {
```
injae-kim, line 81: I think we need comment on here same with `adapt` 😄

MatejNedic:
Hey @imsosleepy tnx on making the PR! Will check PR next week since I am on the conference this week and Tomaz is on vacation.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/AsyncComponentAdapters.java
```diff
@@ -214,6 +229,42 @@ public CompletableFuture<Void> onMessage(Collection<Message<T>> messages) {
...
+			}
+			else {
+				return CompletableFuture.completedFuture(null);
+			}
+		}
+
+		@Override
+		public CompletableFuture<Void> onMessage(Collection<Message<T>> messages) {
+			Collection<Message<T>> filteredMessages = messages.stream()
+															  .filter(filter::process)
```
MatejNedic, line 256: Neat pick again but please switch to for. Since size will be always maximum 10 lets switch to for loop to take benefit of performance.

tomazfernandes (changes requested):
Thanks @imsosleepy for the changes and @MatejNedic for looking into it!

The `MessageFilter` interface looks good.

I'm not sureabout filtering messages at the message listener level since we'd still have the message go through the interceptors and error handlers.

I believe it would make more sense to filter it at the Sink level, so the message never gets emitted and doesn't go through the other components, we just acknowledge it.

What do you folks think?

MatejNedic:
@tomazfernandes I agree makes more sense to me as well. Abstraction `MessageFilter` is spot on what I would expect.


What about `@SqsListener(messageFilter = "messageFilterBeanMethod" )` ? I think this makes much sense I see few use cases. 
 

Are you sure message will go through error handler if it ends without exception in filter tho?

tomazfernandes:
> What about @SqsListener(messageFilter = "messageFilterBeanMethod" ) ? I think this makes much sense I see few use cases.

There's nothing really you can do through the annotation that you can't do with the `ContainerOptions`. In general, my rule of thumb is to limit the amount of configurations on the annotation unless users explicitly ask for them, as otherwise we might end up with all configurations that are available on `ContainerOptions` in the annotation and that would get messy.

For this particular case, I would suggest we stick with the `ContainerOptions` for the scope of this PR to keep it simpler, and we can open an issue / separate PR to add the annotation configuration after we merge it if we think that's important.

What do you think, makes sense?

tomazfernandes:
> Are you sure message will go through error handler if it ends without exception in filter tho?

Yeah, if it doesn't return an error it won't go through the error handler, but if it does return an error, we'll now have a message we don't want to process in the error handler, or perhaps no message at all which would likely trigger another error.

If we prevent the message from entering the processing pipeline altogether I believe it'll be simpler to manage the outcomes.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/config/AbstractMessageListenerContainerFactory.java
```diff
@@ -16,26 +16,20 @@
...
 import io.awspring.cloud.sqs.ConfigUtils;
-import io.awspring.cloud.sqs.listener.AbstractMessageListenerContainer;
-import io.awspring.cloud.sqs.listener.AsyncComponentAdapters;
-import io.awspring.cloud.sqs.listener.AsyncMessageListener;
-import io.awspring.cloud.sqs.listener.ContainerComponentFactory;
-import io.awspring.cloud.sqs.listener.ContainerOptions;
-import io.awspring.cloud.sqs.listener.ContainerOptionsBuilder;
-import io.awspring.cloud.sqs.listener.MessageListener;
-import io.awspring.cloud.sqs.listener.MessageListenerContainer;
+import io.awspring.cloud.sqs.listener.*;
```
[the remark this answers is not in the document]
maciejwalkowiak: running `make format` will solve it. Can be done after merge.

imsosleepy:
Thanks so much for the detailed feedback @tomazfernandes and @MatejNedic!

I agree that filtering at the Sink level makes a lot more sense and helps avoid unnecessary processing through interceptors and error handlers.

That said, I’m still wrapping my head around the architecture around MessageSink and how exactly to plug in filtering at that level. If you have a more concrete suggestion or example of where the filtering logic should live and how it should be wired into the container, that would help me move faster and avoid misalignment.

I’ll dig a bit deeper on my end in the meantime. Thanks again for the guidance — really appreciate it!

imsosleepy:
As suggested, I’ve moved the message filtering logic to the MessageSink layer to avoid unnecessary processing through interceptors and error handlers.

Before proceeding with additional test for batch and FIFO modes, I’d appreciate a review of the current structure to ensure we're aligned.

MatejNedic (commented):
LGTM flexible easy to use I like it!

MatejNedic:
Hey @tomazfernandes can you check this PR when you get time?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/filter/MessageFilter.java
```diff
@@ -0,0 +1,7 @@
+package io.awspring.cloud.sqs.support.filter;
+
+import org.springframework.messaging.Message;
+
+public interface MessageFilter<T> {
+    boolean process(Message<T> message);
```
tomazfernandes: On a second thought, I think the interface should receive and return a `Collection` of messages rather than a single one.

Something like:

```java
public interface MessageFilter<T> {
    Collection<Message<T>> filter(Collection<Message<T>> messages);
}
```

The reason is that if the filter includes some I/O (e.g. querying the DB to check if the users exist, or making an http request), that'll be more performant as a batch query than in a for loop, and if the user is receiving e.g. 500 messages in a batch that can make a sizeable difference.

If the user wants to filter individual messages, it's simple enough for them to iterate on the collection themselves.

What do you folks think?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/sink/FilteredBatchMessageSink.java
```diff
@@ -0,0 +1,35 @@
...
+import io.awspring.cloud.sqs.listener.MessageProcessingContext;
+import io.awspring.cloud.sqs.support.filter.MessageFilter;
+import org.springframework.messaging.Message;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+public class FilteredBatchMessageSink<T> extends BatchMessageSink<T> {
```
tomazfernandes: The `MessageSink` is a composable component, so we should favor composition over inheritance.

In practical terms, it can extend from `AbstractDelegatingMessageListeningSinkAdapter` instead - see `MessageGroupingSinkAdapter` for an example and how it's assembled in the `ContainerFactory`.

Makes sense?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/sink/FilteredBatchMessageSink.java
```diff
@@ -0,0 +1,35 @@
...
+
+    private final MessageFilter<T> filter;
+
+    public FilteredBatchMessageSink(MessageFilter<T> filter) {
+        this.filter = filter;
+    }
+
+    @Override
+    protected CompletableFuture<Void> doEmit(Collection<Message<T>> messages, MessageProcessingContext<T> context) {
+        List<Message<T>> filtered = new ArrayList<>(messages.size());
```
tomazfernandes: This part of the code needs to be non-blocking, otherwise it may block the polling thread e.g. if the filter makes a request or DB query.

I'm not 100% sure what's the best approach, but we could:
- Add the filtering logic in the `execute` method in `AbstractMessageProcessingPipelineSink` in a way that filtering happens in a separate thread
- Make the filtering interface non-blocking (`CompletableFuture`-based)

I would probably go with the first approach.

For that, we'd probably need to have the `MessageFilter` in the `AbstractMessageProcessingPipelineSink` directly.

Let me know your thoughts.

tomazfernandes (commented):
Hey @imsosleepy @MatejNedic, I apologize for the delay.

I left a few comments, let me know your thoughts!

imsosleepy:
> Hey @imsosleepy @MatejNedic, I apologize for the delay.
> 
> I left a few comments, let me know your thoughts!

@tomazfernandes Thanks for the suggestions — I agree with the points you’ve raised on composition over inheritance and non-blocking filtering. I’ll work on these changes over the weekend and share the results. The non-blocking filtering part in the BatchMessageSink area might need a bit more input from you as I work through the details.

imsosleepy:
@tomazfernandes  I’d like to get feedback on whether my current approach makes sense, although FIFO is not yet working.

1. I felt it was too heavy to pass `options` all the way down to the `execute` level, so I implemented an asyncFilter.
2. While working on this, I found that wrapping `OrderedMessageSink`’s `execute` with `asyncFilter` caused issues — the ordering broke and some tests started failing.
3. It seems possible to set the filter in `AbstractPipelineMessageListenerContainer`’s `configureMessageSink`. This way, the filter can be configured at a higher level without pushing it all the way down to `execute`. However, if we do pass it down to `execute`, improving `execute` might be a better approach, because:
4. If we keep the current structure, the `doEmit` implementation in `OrderedMessageSink` will become quite complex.

In summary, the main question is whether the filter should be passed down to `execute`.

I’d appreciate any advice from others.

PR #1419, opened 2025-07-05
feat: add SNS notification to SQS module

chomatdam:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Adds `SnsNotification<T>` wrapper class. This class allows users to receive the complete SNS notification with all its metadata when using `@SqsListener`.

This feature allows users to access both the message payload and all the metadata of an SNS notification (such as messageId, topicArn, subject, timestamp, and message attributes) in a type-safe manner.

Previously, users had to use separate annotations like `@SnsNotificationMessage` and `@SnsNotificationSubject` to access only a subset of an SNS notification. The new wrapper class provides a more comprehensive and convenient way to access all aspects of the notification in a single object.

### Unit Tests
- [x] `SnsNotificationConverterTest`
- [x] `SnsNotificationArgumentResolverTest`

### Integration Tests
- [x] `SnsNotificationIntegrationTests`
   - String message with required SNS notification fields only
   - String message with all SNS notification fields
   - JSON message (`SnsNotification<TestPayload>`)

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
This PR is submitted for preemptive review to gather early feedback while working on the following improvements:
- [x] Test real-world SNS subscription scenarios to identify which notification fields are actually optional
- [x] Add robustness against missing fields that could cause `NullPointerException`

### DIFF
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -694,6 +694,20 @@ public void listen(@SnsNotificationSubject String subject, @SnsNotificationMessa
 }
 ----
 
+Since 3.4.1, you can receive the complete SNS notification with all its metadata using the `SnsNotification<T>` wrapper class. This provides access to both the message payload and all the metadata such as messageId, topicArn, subject, timestamp, and message attributes.
+
+[source, java]
+----
+@SqsListener("my-queue")
+public void listen(SnsNotification<Pojo> notification) {
+    String messageId = notification.getMessageId();
+    String topicArn = notification.getTopicArn();
+    notification.getSubject().ifPresent(subject -> System.out.println("Subject: " + subject));
+    Pojo message = notification.getMessage();
+    System.out.println("Message from topic " + topicArn + " with ID " + messageId + ": " + message);
+}
+----
+
 ===== Specifying a MessageListenerContainerFactory
 A `MessageListenerContainerFactory` can be specified through the `factory` property.
 Such factory will then be used to create the container for the annotated method.
@@ -738,6 +752,8 @@ A number of possible argument types are allowed in the listener method's signatu
 - `Message<MyPojo>` - Provides a `Message<MyPojo>` instance with the deserialized payload and `MessageHeaders`.
 - `List<MyPojo>` - Enables batch mode and receives the batch that was polled from SQS.
 - `List<Message<MyPojo>>` - Enables batch mode and receives the batch that was polled from SQS along with headers.
+- `SnsNotification<MyPojo>` - Provides access to both the message payload and all the metadata of an SNS notification.
+See <<SNS Messages>> for more information.
 - `@Header(String headerName)` - provides the specified header.
 - `@Headers` - provides the `MessageHeaders` or a `Map<String, Object>`
 - `Acknowledgement` - provides methods for manually acknowledging messages for single message listeners.
@@ -1948,7 +1964,7 @@ SqsListenerObservation.Convention sqsListenerObservationConvention() {
                    .and("messaging.operation", "receive")
                    .and("custom.tag", "custom-value");
         }
-        
+
         @Override
         public KeyValues getHighCardinalityKeyValues(SqsListenerObservation.Context context) {
             String paymentId = MessageHeaderUtils.getHeaderAsString(context.getMessage(), "payment-id-header-name");
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -694,7 +694,7 @@ public void listen(@SnsNotificationSubject String subject, @SnsNotificationMessa
 }
 ----
 
-Since 3.4.1, you can receive the complete SNS notification with all its metadata using the `SnsNotification<T>` wrapper class. This provides access to both the message payload and all the metadata such as messageId, topicArn, subject, timestamp, and message attributes.
+Since 3.4.1, the complete SNS notification can be received with all its metadata using the `SnsNotification<T>` wrapper class. This provides access to both the message payload and all the metadata such as messageId, topicArn, subject, timestamp, and message attributes.
 
 [source, java]
 ----
--- a/spring-cloud-aws-samples/spring-cloud-aws-sqs-sample/src/main/java/io/awspring/cloud/sqs/sample/SnsNotificationSample.java
+++ b/spring-cloud-aws-samples/spring-cloud-aws-sqs-sample/src/main/java/io/awspring/cloud/sqs/sample/SnsNotificationSample.java
@@ -16,9 +16,12 @@
 package io.awspring.cloud.sqs.sample;
 
 import io.awspring.cloud.sqs.annotation.SqsListener;
+import io.awspring.cloud.sqs.operations.SqsTemplate;
 import io.awspring.cloud.sqs.support.converter.SnsNotification;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.boot.ApplicationRunner;
+import org.springframework.context.annotation.Bean;
 import org.springframework.stereotype.Component;
 
 /**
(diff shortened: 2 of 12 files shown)


spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SnsNotificationConverter.java
```diff
@@ -0,0 +1,209 @@
...
+            GenericMessage<String> genericMessage = new GenericMessage<>(messageContent);
+            Object convertedPayload = payloadConverter instanceof SmartMessageConverter smartMessageConverter
+                    ? smartMessageConverter.fromMessage(genericMessage, payloadClass, conversionHint)
+                    : payloadConverter.fromMessage(genericMessage, payloadClass);
+
+            // Extract metadata
+			String type = jsonNode.get("Type").asText();
+			String messageId = jsonNode.get("MessageId").asText();
+            String topicArn = jsonNode.get("TopicArn").asText();
+            String subject = jsonNode.has("Subject") ?
```
tomazfernandes: Can you elaborate on the reasoning behind this behavior? Why would we want to have e.g. UnsubscribeURL as the subject?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SnsNotificationConverter.java
```diff
@@ -0,0 +1,209 @@
...
+                ? mp.nested()
+                : conversionHint;
+
+        return messages.stream().map(message -> fromGenericMessage(message, resolvedClazz, hint)).toList();
+    }
+
+    private Object fromGenericMessage(GenericMessage<?> message, Class<?> targetClass,
+            @Nullable Object conversionHint) {
+        try {
+            JsonNode jsonNode = jsonMapper.readTree(message.getPayload().toString());
```
tomazfernandes: Why do we need to use `JsonNode` here? Wouldn't it be simpler to have a POJO / record to represent the message and use `ObjectMapper` to deserialize it, and then map to `SnsNotification`?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SnsNotificationConverter.java
chomatdam:
```java
String subject = jsonNode.has("Subject") ? jsonNode.get("Subject").asText() : null;
```
I don't see the UnsubscribeURL as subject?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SnsNotificationConverter.java
chomatdam: yep, totally agree! I started off discovering the codebase from SnsMessageConverter as base, and I kind of followed a pattern not being one 🙂, will fix that! (today or over the weekend)

chomatdam: looking at it, I now remember why I tried and reverted back at some point to `JsonNode`. I have in `SnsNotification` an Instant that requires the `jackson-datatype-jsr310` module, I think the user can provide its own ObjectMapper making the deserialization not guaranteed.

Any thoughts if we should make the timestamp String instead of Instant, have an ObjectMapper able to register this module, or sticking to JsonNode?

tomazfernandes: We could inform the user that they’d need to provide an `ObjectMapper` with the JSR-310 module if we wanted to support that, but in this case I believe we’re better off keeping it as a `String`, as it preserves the original data without transformation and keeps the solution simpler and more robust.

Makes sense to you?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SnsNotificationConverter.java
tomazfernandes: Oh, my bad, the way it's formatted makes it looks like a long chained ternary if, rather than separate statements.

Please disconsider.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SnsNotificationConverter.java
chomatdam: Sounds reasonable, I made the change (+ formatting) 👍

docs/src/main/asciidoc/sqs.adoc
```diff
@@ -694,6 +694,20 @@ public void listen(@SnsNotificationSubject String subject, @SnsNotificationMessa
 }
 ----
 
+Since 3.4.1, you can receive the complete SNS notification with all its metadata using the `SnsNotification<T>` wrapper class. This provides access to both the message payload and all the metadata such as messageId, topicArn, subject, timestamp, and message attributes.
```
tomazfernandes: Let's avoid using "you" - `Since 3.4.1, the complete SNS notification can be received...`

spring-cloud-aws-samples/spring-cloud-aws-sqs-sample/src/main/java/io/awspring/cloud/sqs/sample/SnsNotificationSample.java
```diff
@@ -0,0 +1,93 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+/**
+ * Sample demonstrating how to receive SNS notifications from SQS.
+ *
+ * @author Damien Chomat
+ */
+@Component
```
tomazfernandes, line 32: How about we add an `ApplicationRunner` to actually trigger sending a message?

https://github.com/awspring/spring-cloud-aws/blob/02e997cfd6e3f54da9e301133a83d6a175045765/spring-cloud-aws-samples/spring-cloud-aws-sqs-sample/src/main/java/io/awspring/cloud/sqs/sample/SpringSqsHandlerSample.java#L60-L67

We could have a `json` file representing the SNS message, or a similar approach.

What do you think?

spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration/SnsNotificationIntegrationTests.java
```diff
@@ -0,0 +1,353 @@
...
+		Instant timestamp = Instant.parse("2023-01-01T00:00:00Z");
+		String unsubscribeUrl = "https://sns.region.amazonaws.com/?Action=Unsubscribe&SubscriptionArn=arn:aws:sns:region:accountId:topicName:uuid";
+		String subject = "subject-json";
+
+		// Create SNS notification with JSON payload
+		String snsJson = "{" + "\"Type\": \"" + type + "\"," + "\"MessageId\": \"" + messageId + "\","
+				+ "\"SequenceNumber\": \"" + sequenceNumber + "\"," + "\"TopicArn\": \"" + topicArn + "\","
+				+ "\"Message\": \"" + messageContent.replace("\"", "\\\"") + "\"," + "\"Timestamp\": \"" + timestamp
+				+ "\"," + "\"UnsubscribeURL\": \"" + unsubscribeUrl + "\"," + "\"Subject\": \"" + subject + "\"," // TODO:
+																													// validate
```
tomazfernandes: TODO + formatting here seems off.

What behavior exactly do you want to clarify?

chomatdam: I removed the TODO, it was initially for the few scenarios I still had to run to validate what's required/optional in the `SnsNotification` object
I moved to text blocks / multiline strings to help spotless doing a better job 🙂

spring-cloud-aws-samples/spring-cloud-aws-sqs-sample/src/main/java/io/awspring/cloud/sqs/sample/SnsNotificationSample.java
chomatdam: I updated the sample with an `ApplicationRunner`, replaced the static data class by a record, and used a text block for the `SnsNotification` to send

chomatdam:
@tomazfernandes @maciejwalkowiak lmk if there is anything else I can do for this PR, or feel free to edit directly for any last mile pre-merge adjustment!

tomazfernandes:
Thanks for the PR @chomatdam! Looking forward to more!

Issue #814, opened 2023-05-12
Why uisin UUID type for ID message header?


ivanfomkin:
**Type**: Bug

**Component**:
"SQS"

**Describe the bug**
I'm try to work with yandex cloud with spring-cloud-aws 3.0.1 and I'm get error when I'm send and receive messages from Yandex Message Queue:
`java.lang.IllegalArgumentException: Invalid UUID string: 92898073-7bd6a160-5797b060-54a7e539` (any random String)
For receiving message root cause is:
```
java.lang.IllegalArgumentException: Invalid UUID string: 92898073-7bd6a160-5797b060-54a7e539
	at java.base/java.util.UUID.fromString1(UUID.java:280) ~[na:na]
	at java.base/java.util.UUID.fromString(UUID.java:258) ~[na:na]
	at io.awspring.cloud.sqs.support.converter.SqsHeaderMapper.toHeaders(SqsHeaderMapper.java:156) ~[spring-cloud-aws-sqs-3.0.1.jar:3.0.1]
	at io.awspring.cloud.sqs.support.converter.SqsHeaderMapper.toHeaders(SqsHeaderMapper.java:57) ~[spring-cloud-aws-sqs-3.0.1.jar:3.0.1]
```

When I'm try to send message to my queue I'm gett this root stacktrace (but messages success posting to queue)

```
Caused by: java.lang.IllegalArgumentException: Invalid UUID string: cbc2063e-e13308bd-4ff0b3a3-b36398c6
	at java.base/java.util.UUID.fromString1(UUID.java:280) ~[na:na]
	at java.base/java.util.UUID.fromString(UUID.java:258) ~[na:na]
	at io.awspring.cloud.sqs.operations.SqsTemplate.lambda$doSendAsync$3(SqsTemplate.java:285) ~[spring-cloud-aws-sqs-3.0.1.jar:3.0.1]
	at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:646) ~[na:na]
```

In AWS docs I found description of MessageId field:
```
A unique identifier for the message. A MessageIdis considered unique across all AWS accounts for an extended period of time.

Type: String
```
It's placed here: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_Message.html

May be should not use UUID-type and just use String?

tomazfernandes:
Hi @ivanfomkin, thanks for bringing this up.

[...]

PR #1433, opened 2025-08-09
Add configurable UUID conversion for non-AWS SQS-compatible services

jm0514:
# Add configurable UUID conversion for non-AWS SQS-compatible services

- [x] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
Added configurable support for SQS-compatible cloud services (like Yandex Message Queue) that use non-UUID MessageId formats. The framework now supports both UUID and non-UUID MessageId handling through a new configuration option.

### Configuration:
```properties
# For AWS SQS (default behavior)
spring.cloud.aws.sqs.convert-message-id-to-uuid=true

# For non-AWS SQS-compatible services (Yandex, etc.)
spring.cloud.aws.sqs.convert-message-id-to-uuid=false
```

Solves #814 

- **Unit Tests**: Added comprehensive tests for `SqsHeaderMapper` and `MessageHeaderUtils`

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -11,6 +11,8 @@ Compared to JMS or other message services Amazon SQS has limitations that should
 Spring Cloud AWS has dedicated support to transfer Java objects with Amazon SQS messages by converting them to JSON.
 * Amazon SQS has a maximum message size of 1MiB per message, so bigger messages will fail to be sent.
 
+NOTE: Since 4.1.0 Spring Cloud AWS SQS also supports SQS-compatible brokers such as Yandex Message Queue. See <<sqs-compatible-brokers-support>>.
+
 A Spring Boot starter is provided to auto-configure SQS integration beans.
 Maven coordinates, using <<index.adoc#bill-of-materials, Spring Cloud AWS BOM>>:
 
@@ -319,28 +321,6 @@ If `SendBatchFailureStrategy#DO_NOT_THROW` is configured, no exception is thrown
 
 For convenience, the `additionalInformation` parameters can be found as constants in the `SqsTemplateParameters` class.
 
-===== Non-UUID Message IDs
-
-By default, Spring Cloud AWS SQS expects the message ID returned by SQS to be a valid UUID.
-If a non-UUID message ID is received, an error is thrown with instructions to enable non-UUID support.
-
-The raw provider message ID is always stored in the `Sqs_RawMessageId` header, regardless of whether UUID conversion is enabled or disabled.
-Access the raw ID via `MessageHeaderUtils.getRawMessageId(message)`.
-
-To enable non-UUID message ID support (e.g., for Yandex Message Queue or other SQS-compatible providers):
-
-[source,properties]
-----
-spring.cloud.aws.sqs.convert-message-id-to-uuid=false
-----
-
-When disabled:
-
-* **Receive side**: A deterministic UUID derived from the raw ID is used as the Spring `MessageHeaders.ID`.
-* **Send side**: If the send response contains a non-UUID message ID,
-  `SendResult.messageId()` returns a deterministic UUID and the raw ID is available
-  in `SendResult.additionalInformation()` under the `SqsTemplateParameters.SQS_RAW_MESSAGE_ID_PARAMETER_NAME` key.
-
 [[template-message-conversion]]
 ==== Template Message Conversion
 
@@ -423,6 +403,29 @@ In case type mapping information is not available, the payload class can be spec
 Optional<Message<SampleRecord>> receivedMessage = template
 			.receive(queue, SampleRecord.class);
 ```
+
+[[sqs-compatible-brokers-support]]
+==== SQS-Compatible Brokers Support
+
+By default, Spring Cloud AWS SQS expects the message ID returned by SQS to be a valid UUID.
+If a non-UUID message ID is received, an error is thrown with instructions to enable non-UUID support.
+
+The raw provider message ID is always stored in the `Sqs_RawMessageId` header, regardless of whether UUID conversion is enabled or disabled.
+Access the raw ID via `MessageHeaderUtils.getRawMessageId(message)`.
+
+To enable non-UUID message ID support (e.g., for Yandex Message Queue or other SQS-compatible providers):
+
+[source,properties]
+----
+spring.cloud.aws.sqs.convert-message-id-to-uuid=false
+----
+
+When disabled:
+
+* **Receive side**: A deterministic UUID derived from the raw ID is used as the Spring `MessageHeaders.ID`.
+* **Send side**: If the send response contains a non-UUID message ID,
+  `SendResult.messageId()` returns a deterministic UUID and the raw ID is available
+  in `SendResult.additionalInformation()` under the `SqsTemplateParameters.SQS_RAW_MESSAGE_ID_PARAMETER_NAME` key.
 
 === Receiving Messages
 


spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SqsHeaderMapper.java
```diff
@@ -156,9 +163,27 @@ public MessageHeaders toHeaders(Message source) {
...
+
+		if (convertMessageIdToUuid && isValidUuid(source.messageId())) {
+			MessageHeaders messageHeaders = accessor.toMessageHeaders();
+			logger.trace("Mapped headers {} for message {}", messageHeaders, source.messageId());
+			return new MessagingMessageHeaders(messageHeaders, UUID.fromString(source.messageId()));
+		} else {
+			accessor.setHeader(SqsHeaders.SQS_AWS_MESSAGE_ID_HEADER, source.messageId());
+			MessageHeaders messageHeaders = accessor.toMessageHeaders();
+			logger.trace("Mapped headers {} for message {}", messageHeaders, source.messageId());
+			return new MessagingMessageHeaders(messageHeaders, UUID.randomUUID());
```
maciejwalkowiak: Passing a random UUID does not feel right. Perhaps `MessagingMessageHeaders` can have id set as String instead of `UUID`? cc @tomazfernandes

jm0514: Thanks for the suggestion! Updated to generate UUID from Message ID instead of random UUID.

tomazfernandes:
Hey @jm0514, thanks for the PR!

> Perhaps MessagingMessageHeaders can have id set as String instead of UUID? cc @tomazfernandes

@maciejwalkowiak, the issue there is that the `MessagingMessageHeaders` inherits from `MessageHeaders` from Spring, and that one requires an `UUID` type in the constructor.

Might be worth raising with the Spring Messaging team?

In the meantime, I think we can make the solution simpler if we leverage the fact that most ID-related operations are made through the `MessageHeaderUtils#getId` method, which already returns a `String`.

Instead of requiring a configuration change to support other ID types, we could simplify the logic by trying to parse the UUID, and if it fails we add the new header as suggested in the PR.

Then in the `MessageHeaderUtils#getId` method, we check for the header first, and return either that, or the Spring Message ID if the former is absent.

The caveat would be that for non UUID ids, the result from getting the ID header directly would be a random UUID, but I don't think there's a way around it with the existing `MessageHeaders` implementation, so properly documenting the behavior should suffice. And since we don't currently support that anyway, we wouldn't really need a breaking change.

What do you folks think?

jm0514:
- **Fail-fast**: When `convert-message-id-to-uuid=true` (default), receiving a non-UUID message ID now throws a `MessagingException` with a clear error     
  message and instructions.
  - **Send-side parity**: `SqsTemplate` now handles non-UUID message IDs in send responses. `SendResult.messageId()` returns a deterministic UUID, and the raw
   ID is available via `SendResult.additionalInformation()` under the `rawMessageId` key.
  - **Naming consistency:** Renamed `Sqs_AWSMessageId` → `Sqs_RawMessageId` and `getAwsMessageId()` → `getRawMessageId()` to be provider-neutral and consistent
  across receive/send sides.
  - **Documentation:** Added Non-UUID Message IDs section in sqs.adoc and property entry in `_configprops.adoc`.

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sqs/SqsAutoConfiguration.java
```diff
@@ -159,10 +161,15 @@ private void configureProperties(SqsContainerOptionsBuilder options) {
 	static class SqsJacksonConfiguration {
 		@ConditionalOnMissingBean
 		@Bean
-		public MessagingMessageConverter<Message> messageConverter(ObjectProvider<JsonMapper> jsonMapperProvider) {
-			JsonMapper jsonMapper = jsonMapperProvider.getIfAvailable();
-			return jsonMapper != null ? new SqsMessagingMessageConverter(jsonMapper)
+		public MessagingMessageConverter<Message> messageConverter(ObjectProvider<JsonMapper> jsonMapperProvider,
```
tomazfernandes: Currently the auto-config manually creates a SqsHeaderMapper and sets the flag on it. Instead, let's add `convertMessageIdToUuid` as a property on both `SqsContainerOptions` and `SqsTemplateOptions`, so  auto-config just does:

```
  factory.configure(options -> options.convertMessageIdToUuid(sqsProperties.getConvertMessageIdToUuid()));
  builder.configure(options -> options.convertMessageIdToUuid(sqsProperties.getConvertMessageIdToUuid()));
```

This keeps the configuration surface consistent between container and template, removes the manual `SqsHeaderMapper` creation from auto-config, and gives programmatic users (without auto-config) a clean way  to set it through the standard options API.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SqsHeaderMapper.java
```diff
@@ -157,9 +164,36 @@ public MessageHeaders toHeaders(Message source) {
 		accessor.copyHeadersIfAbsent(getMessageAttributesAsHeaders(source));
 		accessor.copyHeadersIfAbsent(createDefaultHeaders(source));
 		accessor.copyHeadersIfAbsent(createAdditionalHeaders(source));
-		MessageHeaders messageHeaders = accessor.toMessageHeaders();
-		logger.trace("Mapped headers {} for message {}", messageHeaders, source.messageId());
-		return new MessagingMessageHeaders(messageHeaders, UUID.fromString(source.messageId()));
+
```
tomazfernandes, line 166: Let's consolidate this logic so we get consistent behavior across both listener and template.

My suggestion would be to create a `SqsMessageIdResolver` utility class with a method such as:

```java
 public static MessageHeaders resolveAndAddMessageId(String messageId, MessageHeaders headers,
          boolean convertMessageIdToUuid) {
      MessageHeaders withRawId = MessageHeaderUtils.addHeaderIfAbsent(
          headers, SqsHeaders.SQS_RAW_MESSAGE_ID_HEADER, messageId);
      if (isValidUuid(messageId)) {
          return new MessagingMessageHeaders(withRawId, UUID.fromString(messageId));
      }
      if (convertMessageIdToUuid) {
          throw new MessagingException(String.format(
              "Message ID '%s' is not a valid UUID. To support non-UUID message IDs, "
                  + "set 'spring.cloud.aws.sqs.convert-message-id-to-uuid=false'. "
                  + "The raw message ID will be available via the '%s' header.",
              messageId, SqsHeaders.SQS_RAW_MESSAGE_ID_HEADER));
      }
      return new MessagingMessageHeaders(withRawId,
          UUID.nameUUIDFromBytes(messageId.getBytes(StandardCharsets.UTF_8)));
  }

 private static boolean isValidUuid(String value) {
          try {
              UUID.fromString(value);
              return true;
          }
          catch (IllegalArgumentException e) {
              return false;
          }
      }
```

Then we'd use it in both `SqsHeaderMapper` and `SqsTemplate`.

Open to suggestions if you have a different idea.

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sqs/SqsAutoConfiguration.java
jm0514: Thanks for the great suggestion! This makes the configuration much cleaner and consistent with the existing patterns.

Applied in 1d32b0c1. Moved convertMessageIdToUuid to both SqsContainerOptions and SqsTemplateOptions, so auto-config now uses factory.configure() and builder.configure() instead of manually creating SqsHeaderMapper. 
Added getHeaderMapper() to AbstractMessagingMessageConverter to propagate the option to the header mapper.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SqsHeaderMapper.java
jm0514: Your suggestion was clear and well thought out, so I followed it as proposed.

Applied in f1d7a50c. Created SqsMessageIdResolver with resolveAndAddMessageId() for the listener side and resolveUuid() / isValidUuid() for the template side. 
Both SqsHeaderMapper and SqsTemplate now delegate to this utility class.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
```diff
@@ -327,16 +342,22 @@ private CompletableFuture<Boolean> handleAutoDeduplication(String endpointName)
...
 
-	private <T> SendResult<T> createSendResult(UUID messageId, @Nullable String sequenceNumber, String endpointName,
-			org.springframework.messaging.Message<T> originalMessage) {
+	private <T> SendResult<T> createSendResult(String rawMessageId, @Nullable String sequenceNumber,
+			String endpointName, org.springframework.messaging.Message<T> originalMessage) {
+		Map<String, Object> additionalInfo = new HashMap<>();
+		if (sequenceNumber != null) {
+			additionalInfo.put(SqsTemplateParameters.SEQUENCE_NUMBER_PARAMETER_NAME, sequenceNumber);
+		}
+		UUID messageId = SqsMessageIdResolver.resolveUuid(rawMessageId);
```
tomazfernandes: I'd suggest always adding the new header to `originalMessage` through `resolveAndAddMessageId`, then reading the UUID value back from the resulting message.

That keeps the behavior consistent between receive and send paths and avoids silently handling the UUID conversion when not explicitly configured.

That also allows us to make the remaining methods of `SqsMessageIdResolver` private.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplateParameters.java
```diff
@@ -39,4 +39,9 @@ public class SqsTemplateParameters {
 	 */
 	public static final String ERROR_CODE_PARAMETER_NAME = "code";
 
+	/**
+	 * The raw provider message ID when it is not a valid UUID.
+	 */
+	public static final String RAW_MESSAGE_ID_PARAMETER_NAME = "rawMessageId";
```
tomazfernandes: Let's name it "SQS_RAW_MESSAGE_ID_PARAMETER_NAME" for consistency with the header.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
```diff
@@ -327,16 +342,22 @@ private CompletableFuture<Boolean> handleAutoDeduplication(String endpointName)
...
+		if (!SqsMessageIdResolver.isValidUuid(rawMessageId)) {
+			additionalInfo.put(SqsTemplateParameters.RAW_MESSAGE_ID_PARAMETER_NAME, rawMessageId);
```
tomazfernandes: If this id is already added to the `MessageHeaders`, we may not need to also expose it through `additionalInfo`.

If we keep this parameter, we should always add it so users can rely on it regardless of whether they are using SQS or another implementation.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/AbstractMessagingMessageConverter.java
```diff
@@ -119,6 +120,15 @@ public void setPayloadTypeHeaderValueFunction(Function<Message<?>, String> paylo
 		this.payloadTypeHeaderFunction = payloadTypeHeaderFunction;
 	}
 
+	/**
+	 * Get the {@link HeaderMapper} used to convert headers for
+	 * {@link software.amazon.awssdk.services.sqs.model.Message} instances.
+	 * @return the header mapper instance.
+	 */
+	public HeaderMapper<S> getHeaderMapper() {
```
tomazfernandes: We could change to a configuration-style API here, for example `configureHeaderMapper(Consumer<HeaderMapper<S>> configurer)`.

That preserves encapsulation and keeps the API aligned with the configuration patterns used in the project.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/source/AbstractSqsMessageSource.java
```diff
@@ -109,6 +114,17 @@ protected void doConfigure(ContainerOptions<?, ?> containerOptions) {
 		this.messageVisibility = sqsContainerOptions.getMessageVisibility() != null
 				? (int) sqsContainerOptions.getMessageVisibility().getSeconds()
 				: MESSAGE_VISIBILITY_DISABLED;
+		configureHeaderMapper(sqsContainerOptions);
+	}
+
+	private void configureHeaderMapper(SqsContainerOptions sqsContainerOptions) {
```
tomazfernandes: Let's move the configuration logic to `SqsMessageIdResolver` and centralize all message id resolution logic there and keep behavior consistent between the template and message source.

I suggest something along these lines:
```java
 public static void configureMessageIdResolution(
          MessagingMessageConverter<Message> converter, boolean convertMessageIdToUuid)`.
```

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplateOptions.java
```diff
@@ -24,6 +24,7 @@
 /**
  * Sqs specific options for the {@link SqsTemplate}.
  *
+ * @author Jeongmin Kim
```
tomazfernandes, line 28: Please add me as an author here as well, along with `@since 3.0.0`. I missed it when the class was originally created.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
jm0514: `createSendResult()` now delegates to `resolveAndAddMessageId()`, which adds `SQS_RAW_MESSAGE_ID_HEADER `to the message headers and applies the same fail-fast behavior as the receive path.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplateParameters.java
jm0514: The `RAW_MESSAGE_ID_PARAMETER_NAME `constant was removed since `createSendResult()` now delegates to `resolveAndAddMessageId()`, which stores the raw ID in `MessageHeaders` via `SQS_RAW_MESSAGE_ID_HEADER`. 
This made the `additionalInfo` usage and the constant itself unnecessary.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SqsMessageIdResolver.java
```diff
@@ -0,0 +1,98 @@
...
+	 * @param convertMessageIdToUuid whether to enforce UUID message IDs.
+	 * @return the resolved {@link MessageHeaders} with the message ID set.
+	 * @throws MessagingException if the message ID is not a valid UUID and conversion is enabled.
+	 */
+	public static MessageHeaders resolveAndAddMessageId(String messageId, MessageHeaders headers,
+			boolean convertMessageIdToUuid) {
+		MessageHeaders withRawId = MessageHeaderUtils.addHeaderIfAbsent(headers, SqsHeaders.SQS_RAW_MESSAGE_ID_HEADER,
+				messageId);
+		if (isValidUuid(messageId)) {
+			return new MessagingMessageHeaders(withRawId, UUID.fromString(messageId));
```
tomazfernandes: Not critical, but we're converting `UUID.fromString` twice.

Perhaps the `isValidUuid` method could return an `Optional<UUID>` so we convert it only once. Method name might need to be adjusted as well.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
```diff
@@ -327,16 +334,21 @@ private CompletableFuture<Boolean> handleAutoDeduplication(String endpointName)
...
-			org.springframework.messaging.Message<T> originalMessage) {
+	private <T> SendResult<T> createSendResult(String rawMessageId, @Nullable String sequenceNumber,
+			String endpointName, org.springframework.messaging.Message<T> originalMessage) {
+		MessageHeaders resolvedHeaders = SqsMessageIdResolver.resolveAndAddMessageId(rawMessageId,
+				originalMessage.getHeaders(), this.convertMessageIdToUuid);
+		UUID messageId = resolvedHeaders.getId();
+		Map<String, Object> additionalInfo = new HashMap<>();
+		if (sequenceNumber != null) {
+			additionalInfo.put(SqsTemplateParameters.SEQUENCE_NUMBER_PARAMETER_NAME, sequenceNumber);
+		}
```
tomazfernandes, line 350: Looks good, but we're returning the original message which does not contain the new header.

jm0514:
@tomazfernandes Thanks for catching these — I missed the header propagation issue in SendResult while refactoring.
 
 Here's a summary of the changes:                                                                                                                     
                             
  - `SqsMessageIdResolver`: Renamed `isValidUuid` to `tryParseUuid` returning `Optional<UUID>` to avoid duplicate `UUID.fromString` parsing.                     
  - `SqsTemplate`: `createSendResult` now builds a new message with `resolvedHeaders` so `Sqs_RawMessageId` is included in `SendResult.message()`.
  - `AbstractListenerObservation`: Use `getRawMessageId()` in `Context` so observation tags reflect the broker ID for non-UUID environments.                 
  - `Doc`: Updated to clarify that `Sqs_RawMessageId` header is always present regardless of the `convert-message-id-to-uuid` flag.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
tomazfernandes: @jm0514, this last change is causing `shouldSendAndReceiveBatchFifo` integration test to break.

This raises an important point: while returning the message with the id from the broker would make ids consistent, returning the original message in the `SendResult` allows the sender to correlate the send results with the original `Message` instances they sent, which is a stronger reason for keeping the original behavior.

Let's go with your original direction - add the original message to `SendResult`, and have an additional parameter for the raw message id. Please update the documentation accordingly.

Thanks.

jm0514: Restored the original message in `SendResult`, and added the raw message ID to `additionalInformation` via `SQS_RAW_MESSAGE_ID_PARAMETER_NAME`. Updated the documentation and tests accordingly.

tomazfernandes:
@maciejwalkowiak, the CI is failing due to Localstack requiring auth: https://github.com/awspring/spring-cloud-aws/actions/runs/23697734758/job/69074091139?pr=1433#step:4:7455

Related: https://github.com/awspring/spring-cloud-aws/issues/1586#issuecomment-4148136287

I've ran the tests locally and they pass, so I'll proceed with the merge:
image

jm0514:
Thanks for the review and merge!

PR #1444, opened 2025-08-24
Do not log warning message for STS credentials when properties are not configured

### LINKED ISSUE #862: "Skipping creating `StsCredentialsProvider`" is misleading when only using StsAssumeRoleCredentialsProvider
**Type**: Bug

**Component**:
STS/Autoconfiguration

**Describe the bug**
We are only using an `StsAssumeRoleCredentialsProvider` to assume a role and not the `StsWebIdentityTokenFileCredentialsProvider`. So there is no need to provide an `webIdentityTokenFile` or other `StsProperties`.

Now the `CredentialsProviderAutoConfiguration#createCredentialsProvider` will log this warning because it founds the `StsWebIdentityTokenFileCredentialsProvider` dependency but no proper config for it: 
```
Skipping creating `StsCredentialsProvider`. `software.amazon.awssdk:sts` is on the classpath, but neither `spring.cloud.aws.credentials.sts` properties are configured nor `AWS_WEB_IDENTITY_TOKEN_FILE` or the javaproperty `aws.webIdentityTokenFile` is set
```
But this config issue is only important for the case you want to use this `StsWebIdentityTokenFileCredentialsProvider`. So assuming we want to do the `StsWebIdentityTokenFileCredentialsProvider` based login only because this class is inside the class-path is not correct.

Seen in Spring AWS Cloud 3.0.1

**Sample**
Just put this dependency in your project and the log message will be shown:
```
    <dependency>
      <groupId>software.amazon.awssdk</groupId>
      <artifactId>sts</artifactId>
    </dependency>
```

I would suggest to only log this warning if the `StsProperties` are not null and an error happens.

maciejwalkowiak:
Fixes #862

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/core/CredentialsProviderAutoConfiguration.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/core/CredentialsProviderAutoConfiguration.java
@@ -44,6 +44,8 @@ import software.amazon.awssdk.regions.providers.AwsRegionProvider;
 import software.amazon.awssdk.services.sts.StsClient;
 import software.amazon.awssdk.services.sts.auth.StsWebIdentityTokenFileCredentialsProvider;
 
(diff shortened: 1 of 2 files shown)


MatejNedic:
Also valid solution. Check -> #1523

PR #1483, opened 2025-09-25
#1457-Introduce-kinesis-autoconfiguration

MatejNedic:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


maciejwalkowiak (approved):
Assuming it is compatible with https://github.com/awspring/spring-cloud-aws/pull/1479 we can merge! 

But if we are adding Kinesis integration we can also think about:
- https://github.com/awslabs/amazon-kinesis-client
- https://github.com/awslabs/amazon-kinesis-producer

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/kinesis/KinesisAsyncClientCustomizer.java
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright 2013-2022 the original author or authors.
```
artembilan: `2013-2025`?

artembilan (changes requested):
Just couple hygiene nit-picks.
This affects my PR for `spring-cloud-aws-kinesis` module only at the dependency .
The KPL/KCL could be auto-configured in other PRs.
Thanks

MatejNedic:
Hey @artembilan I will add KPL/KCL auto configuration after I merge yours PR. Will then align dependencies and other stuff.

spring-cloud-aws-dependencies/pom.xml
```diff
@@ -33,6 +33,8 @@
 		<eclipse.jakarta.mail.version>2.0.3</eclipse.jakarta.mail.version>
 		<spring-modulith.version>2.0.0-M2</spring-modulith.version>
 		<amazon.s3.accessgrants>2.3.0</amazon.s3.accessgrants>
+		<amazon.kinesis.client>2.7.0</amazon.kinesis.client>
+		<amazon.kinesis.producer>0.15.12</amazon.kinesis.producer>
```
artembilan: Looks like this PR was not rebased to main.
I agree that these property names are much better, than I did in my Kinesi contribution, however the versions are old.
Why did you decide to go with these versions instead of moving to their newer?

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/kinesis/KinesisAutoConfiguration.java
```diff
@@ -0,0 +1,122 @@
...
+@AutoConfiguration
+@ConditionalOnClass({ KinesisAsyncClient.class })
+@EnableConfigurationProperties({ KinesisProperties.class })
+@AutoConfigureAfter({ CredentialsProviderAutoConfiguration.class, RegionProviderAutoConfiguration.class })
+@ConditionalOnProperty(value = "spring.cloud.aws.kinesis.enabled", havingValue = "true", matchIfMissing = true)
+public class KinesisAutoConfiguration {
+
+	@ConditionalOnMissingBean
+	@Bean
+	public KinesisAsyncClient kinesisAsyncClient(KinesisProperties properties,
```
artembilan, line 42: There is an interesting situation.
The KPL does not require AWS `kinesis` dependency.
Even if it works with Kinesis service, it does everything itself.
What I mean that this bean could be also optional if we are interested only in KPL where the whole AWS `kinesis` dependency could be omitted.
Just a thought if you still explore more options here.
I mean the whole `KinesisProducerAutoConfiguration` could go to its top-level class, since technically there is nothing to do with `kinesis` dependencies at all.
And when end-user chose something like `spring-cloud-aws-starter-kpl`, we would have only a KPL as dependency and its auto-configuration.

artembilan (changes requested):

spring-cloud-aws-dependencies/pom.xml
MatejNedic: Will rebase the main.

Tbh have not checked the versioins againat yours will upgrade!

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/kinesis/KinesisAutoConfiguration.java
MatejNedic: This is very interesting. Will split autoconfigurarion than as you suggested. Was not aware of it tnx!

spring-cloud-aws-starters/spring-cloud-aws-starter-kinesis/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+			<groupId>software.amazon.awssdk</groupId>
+			<artifactId>kinesis</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>software.amazon.kinesis</groupId>
+			<artifactId>amazon-kinesis-client</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>com.amazonaws</groupId>
+			<artifactId>amazon-kinesis-producer</artifactId>
```
artembilan: I think KCL and KPL should go to their own starters.
While KCL might just depend on the Kinesis SDK itself, that does not mean that I want the KCL if I'm not interested in it.
The KPL, as we discussed before, fully does not depend on the Kinesis SDK.
Therefore, if I'm interested only in KPL, I'd like not see extra unnecessary deps.

artembilan (changes requested):

spring-cloud-aws-starters/spring-cloud-aws-starter-kinesis/pom.xml
MatejNedic: Tnx, I have started moving them I have not yet finished everything to be ready for commit.

One thing I am noticing in KCL we will need to add DynamoDB and Cloudwatch as a optional dependecnies as well. Since both Scheduler can have them or not. 

Anyways I have split them in:

```
	<module>spring-cloud-aws-starters/spring-cloud-aws-starter-kinesis</module>
		<module>spring-cloud-aws-starters/spring-cloud-aws-starter-kinesis-producer</module>
		<module>spring-cloud-aws-starters/spring-cloud-aws-starter-kinesis-client-library</module>
```

I hope I will have everything written with tests till end of this week! I am working on one more integration so I am doing them in parallel.

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/kinesis/KinesisClientLibraryAutoConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+@EnableConfigurationProperties({ KinesisClientLibraryProperties.class })
+@AutoConfigureAfter({ CredentialsProviderAutoConfiguration.class, RegionProviderAutoConfiguration.class, KinesisAutoConfiguration.class })
+@ConditionalOnProperty(value = "spring.cloud.aws.kinesis.client.library.enabled", havingValue = "true", matchIfMissing = true)
+public class KinesisClientLibraryAutoConfiguration {
+
+
+
+	@ConditionalOnMissingBean
+	@Bean
+	public Scheduler scheduler(ObjectProvider<DynamoDbAsyncClient> dynamoDbClient, ObjectProvider<CloudWatchAsyncClient> cloudWatchClient,
```
MatejNedic: Hey @artembilan  I would appreciate your thoughts on this one.

I am not done yet there is still much work. I was thinking to create multiple Scheduler beans and have the first one Primary.

Now in properties I would support 
```
	private String streamName;
	private String applicationName; 
```

As a list of objects so users could do.

```
spring.cloud.aws.kinesis.client.library.streamName[0]=firstStream
spring.cloud.aws.kinesis.client.library.applicationName[0]=applicationName
spring.cloud.aws.kinesis.client.library.streamName[1]=secondStream
spring.cloud.aws.kinesis.client.library.applicationName[1]=applicationName
```

This configuration would result in two Scheduler beans being created, named after their respective streamName values.
The first one would be the default (`@Primary`) bean, and the rest could be injected using `@Qualifier`(streamName).

artembilan: We don't do that in Spring Boot.
More over, the KCL supports several streams.
See `MultiStreamTracker` abstraction.
I'm about to PR KCL with Spring Integration.
So, there you would see how is that used and so on.

The main point of this my comment is "no, we don't do several distinguished beans from list of props".

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/kinesis/KinesisClientLibraryAutoConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+@ConditionalOnClass({ KinesisAsyncClient.class, Scheduler.class })
+@EnableConfigurationProperties({ KinesisClientLibraryProperties.class })
+@AutoConfigureAfter({ CredentialsProviderAutoConfiguration.class, RegionProviderAutoConfiguration.class,
+		KinesisAutoConfiguration.class })
+@ConditionalOnProperty(value = "spring.cloud.aws.kinesis.client.library.enabled", havingValue = "true", matchIfMissing = true)
+public class KinesisClientLibraryAutoConfiguration {
+
+	@ConditionalOnMissingBean
+	@Bean
+	public Scheduler scheduler(ObjectProvider<DynamoDbAsyncClient> dynamoDbClient,
```
artembilan: This is an active components which requires some lifecycle control.
Plus you cannot guess if end-user is interested only in one of this.
We may go something like `SchedulerFactory` to hide some common properties, but the rest should be left up to end-user configuration.

This about this like the listener container behind `@SqsListener`.
And I believe eventually we can come up with something like `@KclListener`

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/kinesis/KinesisClientLibraryAutoConfiguration.java
MatejNedic: Thanks! I will keep it simple then. I see where this can be useful but than again, I am aware Spring Boot does not do this so I will follow the patterns. Same is with datasource support.

Was just thinking if it would ever make sense in this case.

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/kinesis/KinesisClientLibraryAutoConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient;
+import software.amazon.awssdk.services.kinesis.KinesisAsyncClient;
+import software.amazon.kinesis.coordinator.Scheduler;
+import software.amazon.kinesis.processor.ShardRecordProcessorFactory;
+
+@AutoConfiguration
+@ConditionalOnClass({ KinesisAsyncClient.class, Scheduler.class })
+@EnableConfigurationProperties({ KinesisClientLibraryProperties.class })
+@AutoConfigureAfter({ CredentialsProviderAutoConfiguration.class, RegionProviderAutoConfiguration.class,
+		KinesisAutoConfiguration.class })
```
artembilan: Cannot we use `CloudWatchExportAutoConfiguration` instead of `ObjectProvider`?
I'm not sure if `DynamoDbAsyncClient` is the goal in this project yet, though...

artembilan (changes requested):
Looks great so far!

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/kinesis/KinesisClientLibraryAutoConfiguration.java
MatejNedic: Will move away from ObjectProvider since both DynamoDbAsyncClient and CloudWatchAsync client are not nullabe for Scheduler.

Meaning will have to play little bit with async autoconfiguration in general since most of our clients are sync

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/kinesis/KinesisClientLibraryAutoConfiguration.java
MatejNedic: I had a same idea for 4.1 to introduce KclListener and work on it! :)

Edit:

Actually thinking about it I will take little bit longer do dig into SqsContainer and KafkaContainer and will do it as part od 4.0.0

MatejNedic: I will take some time reading through Spring Messaging, Spring Kafka and SQS integration. I need more reading on this, but i understand the point behind SchedularFactory and containers. Tnx on constructive feedback Artem, I appreciate it much.

MatejNedic:
Closing in favour of KclListener pr

PR #1498, opened 2025-10-26
Lift version of spring cloud 2025 m4

MatejNedic:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description
@tomazfernandes  everything regarding converters I have changed. I am not sure what is best way to tackle JsonMappers since JacksonJsonMessageConverter hides it. There is no way to set only to propagate one when using constructor.
I am tbh not sure what is best action here but it is a breaking change so please look and lets agree how to move. Be free to checkout PR and push directly :) 

There are also few tests I have put to ignore till we settle the approach.

I tbh would go that users provide their own converters personally. JsonMapper default one already is registering few modules so yea.

I have to check cloudwatch integration currently it is failing I will check it tomorrow evening.

Also I have to rework docs since a autoconfiguration will prolly change!

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/core/CredentialsProviderAutoConfiguration.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/core/CredentialsProviderAutoConfiguration.java
@@ -18,7 +18,6 @@ package io.awspring.cloud.autoconfigure.core;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Objects;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.ObjectProvider;
@@ -163,8 +162,7 @@ public class CredentialsProviderAutoConfiguration {
 				propertyMapper.from(stsProperties::getRoleArn).to(builder::roleArn);
 				propertyMapper.from(stsProperties::getWebIdentityTokenFile)
 						.to(b -> builder.webIdentityTokenFile(Paths.get(b)));
-				propertyMapper.from(stsProperties::getRoleSessionName)
-						.to(builder::roleSessionName);
+				propertyMapper.from(stsProperties::getRoleSessionName).to(builder::roleSessionName);
 			}
 			return builder.build();
 		}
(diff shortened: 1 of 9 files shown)


spring-cloud-aws-autoconfigure/pom.xml
```diff
@@ -32,6 +32,7 @@
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-metrics</artifactId>
+			<version>4.0.0-M1</version>
```
maciejwalkowiak: As far as i can see `spring-boot-metrics` is replaced by `spring-boot-micrometer-metrics`  and we should use this one?

tomazfernandes:
Hey @MatejNedic,

I suggest that the project follows the pattern the Spring team did for their migration, in order to support both Jackson v2 and v3 as they do.

The pattern is, roughly:
- Duplicate all classes that use Jackson v2
- Adjust the imports in the duplicate to the new v3 namespace
- Change ObjectMapper usages to JsonMapper in the duplicates

Here's an example with [DefaultKafkaHeaderMapper](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java) and [JsonKafkaHeaderMapper](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/support/JsonKafkaHeaderMapper.java)

At some decision points, they check which Jackson versions are in the classpath, such as in [MessagingMessageConverter](https://github.com/spring-projects/spring-kafka/blob/6594f5b7ab3d26cf01d03aed58e40c47f950b8ac/spring-kafka/src/main/java/org/springframework/kafka/support/converter/MessagingMessageConverter.java#L93-L105)

This is the PR from Spring Kafka's migration for reference: https://github.com/spring-projects/spring-kafka/pull/3996

Let me know if that makes sense to you.

MatejNedic:
Hey @tomazfernandes 

Sounds good to me!

I have seen while moving to jackson 3 Deprecated on top of jackson 2 convrter classes.

Can you make the changes required against this PR? Just check it out and push freely. I sadly won't have time till middle of next week.

tomazfernandes:
Unfortunately, I won't have the time either. I think it's fine for us to leave this migration for M2. I'll open an issue to track this, so the community can pitch in as well.

tomazfernandes:
Created the issue: https://github.com/awspring/spring-cloud-aws/issues/1511

Feel free to edit if you'd like.

MatejNedic:
Lets go with M1 release with Jackson 2 support!

MatejNedic:
Closing in favour https://github.com/awspring/spring-cloud-aws/pull/1521

4.0.0-M1 will be Jackson2 support only 
4.0.0-RC1 will be both Jackson2 and Jackson3 

Will aim RC1 to be released next week and M1 friday/saturday

PR #1506, opened 2025-11-01
SQS Cherry Picks to 3.4.1

tomazfernandes:
- Validation Message
- Exponential Error Handlers
- Fix: Polling future removed on exception


tomazfernandes:
I'll open a separate PR for each of the 3 commits so it's cleaner.

PR #1518, opened 2025-11-18
Fix @Value Usage

DennisRippinger:
## :scroll: Description
The documentation example refers to the usage of `@Value`, but without the SpEl expression implementers get the value verbatim.


MatejNedic (approved):
Thanks @DennisRippinger on PR!

PR #1536, opened 2025-12-06
Fix: Nondeterminism behavior in SqsListenerAnnotationBeanPostProcessorTests

yonghanlin:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

This pull request updates the test ```shouldRegisterClassLevelSqsListenerWithSqsHandlers``` in
```SqsListenerAnnotationBeanPostProcessorTests``` to avoid relying on the implicit method ordering returned by reflection.

Previously, the test compared methods using fixed index positions:
```
methods.get(0)
methods.get(1)
```
However, it does not guarantee a deterministic ordering of methods returned by [Class#getDeclaredMethods()](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getDeclaredMethods--) or collections derived from it. As a result, the test becomes nondeterministic and fails under [NonDex tool](https://github.com/TestingResearchIllinois/NonDex) or repeated runs. The updated version checks only the presence of the two ```@SqsHandler``` overloads by validating their parameter types (```String.class``` and ```Integer.class```) in any order, making the test deterministic.

This change ensures that the test remains stable under randomized iteration orders. When running with NonDex, the test failed intermittently because it assumed a fixed order for handler method discovery, even though this order is not guaranteed. By relaxing the assertion to be order-independent, we preserve the original intent of the test and eliminating nondeterministic behavior. This improves test reliability in continuous integration environments.

Run NonDex on ```spring-cloud-aws-sqs``` module using the following command:
```
mvn -pl spring-cloud-aws-sqs edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=io.awspring.cloud.sqs.annotation.SqsListenerAnnotationBeanPostProcessorTests#shouldRegisterClassLevelSqsListenerWithSqsHandlers
```

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/annotation/SqsListenerAnnotationBeanPostProcessorTests.java
+++ b/spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/annotation/SqsListenerAnnotationBeanPostProcessorTests.java
@@ -257,10 +257,8 @@ class SqsListenerAnnotationBeanPostProcessorTests {
 		assertThat(endpoint.getLogicalNames()).containsExactly("classLevelQueue");
 		assertThat(endpoint).isInstanceOfSatisfying(MultiMethodSqsEndpoint.class, multiMethodSqsEndpoint -> {
 			assertThat(multiMethodSqsEndpoint.getMethods()).hasSize(2);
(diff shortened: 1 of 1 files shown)


maciejwalkowiak:
Thanks @yonghanlin!

PR #1539, opened 2025-12-11
Introduce jackson 3 support

MatejNedic:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/config/secretsmanager/SecretsManagerExceptionHappenedAnalyzer.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/config/secretsmanager/SecretsManagerExceptionHappenedAnalyzer.java
@@ -1,13 +1,31 @@
+/*
+ * Copyright 2013-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package io.awspring.cloud.autoconfigure.config.secretsmanager;
 
 import org.springframework.boot.diagnostics.AbstractFailureAnalyzer;
 import org.springframework.boot.diagnostics.FailureAnalysis;
 
-public class SecretsManagerExceptionHappenedAnalyzer extends AbstractFailureAnalyzer<AwsSecretsManagerPropertySourceNotFoundException> {
+public class SecretsManagerExceptionHappenedAnalyzer
+		extends AbstractFailureAnalyzer<AwsSecretsManagerPropertySourceNotFoundException> {
 
 	@Override
 	protected FailureAnalysis analyze(Throwable rootFailure, AwsSecretsManagerPropertySourceNotFoundException cause) {
-		return new FailureAnalysis("Could not import properties from AWS Secrets Manager. Exception happened while trying to load the keys: " + cause.getMessage(),
-			"Depending on error message determine action course", cause);
+		return new FailureAnalysis(
+				"Could not import properties from AWS Secrets Manager. Exception happened while trying to load the keys: "
+						+ cause.getMessage(),
+				"Depending on error message determine action course", cause);
 	}
 }
--- a/spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/config/parameterstore/ParameterStoreConfigDataLoaderIntegrationTests.java
+++ b/spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/config/parameterstore/ParameterStoreConfigDataLoaderIntegrationTests.java
@@ -150,21 +150,21 @@ class ParameterStoreConfigDataLoaderIntegrationTests {
 		SpringApplication application = new SpringApplication(App.class);
 		application.setWebApplicationType(WebApplicationType.NONE);
 		String applicationYaml = """
-first:
-  message: value from tests
-  another-parameter: another parameter value
-second:
-  secondMessage: second value from tests
-""";
+				first:
+				  message: value from tests
+				  another-parameter: another parameter value
+				second:
+				  secondMessage: second value from tests
+				""";
 		putParameter(localstack, "/test/second/message", applicationYaml, REGION);
 
 		try (ConfigurableApplicationContext context = runApplication(application,
-			"aws-parameterstore:/test/second/?extension=yaml")) {
+				"aws-parameterstore:/test/second/?extension=yaml")) {
 			assertThat(context.getEnvironment().getProperty("first.message")).isEqualTo("value from tests");
 			assertThat(context.getEnvironment().getProperty("first.another-parameter"))
-				.isEqualTo("another parameter value");
+					.isEqualTo("another parameter value");
 			assertThat(context.getEnvironment().getProperty("second.secondMessage"))
-				.isEqualTo("second value from tests");
+					.isEqualTo("second value from tests");
 			assertThat(context.getEnvironment().getProperty("non-existing-parameter")).isNull();
 		}
 	}
--- a/spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/config/secretsmanager/SecretsManagerConfigDataLoaderIntegrationTests.java
+++ b/spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/config/secretsmanager/SecretsManagerConfigDataLoaderIntegrationTests.java
@@ -216,7 +216,8 @@ class SecretsManagerConfigDataLoaderIntegrationTests {
 		SpringApplication application = new SpringApplication(App.class);
 		application.setWebApplicationType(WebApplicationType.NONE);
 
-		try (ConfigurableApplicationContext context = runApplication(application, "aws-secretsmanager:/some/random/config")) {
+		try (ConfigurableApplicationContext context = runApplication(application,
+				"aws-secretsmanager:/some/random/config")) {
 			fail("Context without keys should fail to start");
 		}
 		catch (Exception e) {
@@ -225,7 +226,7 @@ class SecretsManagerConfigDataLoaderIntegrationTests {
 			// error message
 			// Ensure that new line character should be platform independent
 			String errorMessage = "Description:%1$s%1$sCould not import properties from AWS Secrets Manager. Exception happened while trying to load"
-				.formatted(NEW_LINE_CHAR);
+					.formatted(NEW_LINE_CHAR);
 			assertThat(output.getOut()).contains(errorMessage);
 		}
 	}
--- a/spring-cloud-aws-dynamodb/src/main/java/io/awspring/cloud/dynamodb/DefaultDynamoDbTableSchemaResolver.java
+++ b/spring-cloud-aws-dynamodb/src/main/java/io/awspring/cloud/dynamodb/DefaultDynamoDbTableSchemaResolver.java
@@ -19,7 +19,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
-
 import software.amazon.awssdk.enhanced.dynamodb.TableSchema;
 
 /**
--- a/spring-cloud-aws-dynamodb/src/test/java/io/awspring/cloud/dynamodb/DynamoDbTableSchemaResolverTest.java
+++ b/spring-cloud-aws-dynamodb/src/test/java/io/awspring/cloud/dynamodb/DynamoDbTableSchemaResolverTest.java
@@ -19,7 +19,6 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import java.util.List;
-
 import org.junit.jupiter.api.Test;
 import software.amazon.awssdk.enhanced.dynamodb.TableSchema;
 import software.amazon.awssdk.enhanced.dynamodb.mapper.StaticTableSchema;
--- a/spring-cloud-aws-kinesis/src/test/java/io/awspring/cloud/kinesis/integration/DynamoDbStreamToKclIntegrationTests.java
+++ b/spring-cloud-aws-kinesis/src/test/java/io/awspring/cloud/kinesis/integration/DynamoDbStreamToKclIntegrationTests.java
@@ -89,15 +89,15 @@ class DynamoDbStreamToKclIntegrationTests implements LocalstackContainerTest {
 		DYNAMODB_STREAM_ARN = createDemoTable();
 
 		AmazonDynamoDBStreamsAdapterClient streamsAdapterClient = new AmazonDynamoDBStreamsAdapterClient(
-			DYNAMODB_STREAMS, null);
+				DYNAMODB_STREAMS, null);
 
 		await().atMost(Duration.ofMinutes(2))
-			.untilAsserted(() -> assertThat(
-				streamsAdapterClient.describeStream(builder -> builder.streamName(DYNAMODB_STREAM_ARN)))
-				.succeedsWithin(Duration.ofSeconds(60))
-				.extracting(describeStreamResponse -> describeStreamResponse.streamDescription()
-					.streamStatusAsString())
-				.isEqualTo("ENABLED"));
+				.untilAsserted(() -> assertThat(
+						streamsAdapterClient.describeStream(builder -> builder.streamName(DYNAMODB_STREAM_ARN)))
+						.succeedsWithin(Duration.ofSeconds(60))
+						.extracting(describeStreamResponse -> describeStreamResponse.streamDescription()
+								.streamStatusAsString())
+						.isEqualTo("ENABLED"));
 	}
 
 	private static String createDemoTable() {
--- a/spring-cloud-aws-kinesis/src/test/java/io/awspring/cloud/kinesis/integration/DynamoDbStreamToKinesisIntegrationTests.java
+++ b/spring-cloud-aws-kinesis/src/test/java/io/awspring/cloud/kinesis/integration/DynamoDbStreamToKinesisIntegrationTests.java
@@ -22,7 +22,6 @@ import com.amazonaws.services.dynamodbv2.streamsadapter.AmazonDynamoDBStreamsAda
 import io.awspring.cloud.kinesis.LocalstackContainerTest;
 import java.time.Duration;
 import java.util.Map;
-
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.junit.jupiter.api.BeforeAll;
@@ -86,15 +85,15 @@ class DynamoDbStreamToKinesisIntegrationTests implements LocalstackContainerTest
 		DYNAMODB_STREAM_ARN = createDemoTable();
 
 		AmazonDynamoDBStreamsAdapterClient streamsAdapterClient = new AmazonDynamoDBStreamsAdapterClient(
-			DYNAMODB_STREAMS, null);
+				DYNAMODB_STREAMS, null);
 
 		await().atMost(Duration.ofMinutes(2))
-			.untilAsserted(() -> assertThat(
-				streamsAdapterClient.describeStream(builder -> builder.streamName(DYNAMODB_STREAM_ARN)))
-				.succeedsWithin(Duration.ofSeconds(60))
-				.extracting(describeStreamResponse -> describeStreamResponse.streamDescription()
-					.streamStatusAsString())
-				.isEqualTo("ENABLED"));
+				.untilAsserted(() -> assertThat(
+						streamsAdapterClient.describeStream(builder -> builder.streamName(DYNAMODB_STREAM_ARN)))
+						.succeedsWithin(Duration.ofSeconds(60))
+						.extracting(describeStreamResponse -> describeStreamResponse.streamDescription()
+								.streamStatusAsString())
+						.isEqualTo("ENABLED"));
 	}
 
 	private static String createDemoTable() {
@@ -133,11 +132,8 @@ class DynamoDbStreamToKinesisIntegrationTests implements LocalstackContainerTest
 				.join();
 
 		Message<?> receive = this.kinesisReceiveChannel.receive(120_000);
-		assertThat(receive)
-			.satisfies(LOGGER::warn)
-			.extracting(Message::getPayload)
-			.asString()
-			.contains("some_id", "some value", "\"eventName\":\"INSERT\"", "\"eventSource\":\"aws:dynamodb\"");
+		assertThat(receive).satisfies(LOGGER::warn).extracting(Message::getPayload).asString().contains("some_id",
+				"some value", "\"eventName\":\"INSERT\"", "\"eventSource\":\"aws:dynamodb\"");
 
 		DYNAMODB.updateItem(
 				builder -> builder.tableName(TEST_TABLE).key(Map.of(TABLE_KEY, AttributeValue.fromS("some_id")))
@@ -146,22 +142,16 @@ class DynamoDbStreamToKinesisIntegrationTests implements LocalstackContainerTest
 				.join();
 
 		receive = this.kinesisReceiveChannel.receive(30_000);
-		assertThat(receive)
-			.satisfies(LOGGER::warn)
-			.extracting(Message::getPayload)
-			.asString()
-			.contains("some_id", "some value", "updated value", "\"eventName\":\"MODIFY\"");
+		assertThat(receive).satisfies(LOGGER::warn).extracting(Message::getPayload).asString().contains("some_id",
+				"some value", "updated value", "\"eventName\":\"MODIFY\"");
 
 		DYNAMODB.deleteItem(
 				builder -> builder.tableName(TEST_TABLE).key(Map.of(TABLE_KEY, AttributeValue.fromS("some_id"))))
 				.join();
 
 		receive = this.kinesisReceiveChannel.receive(30_000);
-		assertThat(receive)
-			.satisfies(LOGGER::warn)
-			.extracting(Message::getPayload)
-			.asString()
-			.contains("some_id", "\"eventName\":\"REMOVE\"");
+		assertThat(receive).satisfies(LOGGER::warn).extracting(Message::getPayload).asString().contains("some_id",
+				"\"eventName\":\"REMOVE\"");
 	}
 
 	@Configuration(proxyBeanMethods = false)
--- a/spring-cloud-aws-parameter-store/src/main/java/io/awspring/cloud/parameterstore/ParameterStorePropertySource.java
+++ b/spring-cloud-aws-parameter-store/src/main/java/io/awspring/cloud/parameterstore/ParameterStorePropertySource.java
@@ -16,11 +16,9 @@
 package io.awspring.cloud.parameterstore;
 
 import io.awspring.cloud.core.config.AwsPropertySource;
-
 import java.io.ByteArrayInputStream;
 import java.io.InputStream;
 import java.util.*;
-
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.jspecify.annotations.Nullable;
@@ -105,8 +103,9 @@ public class ParameterStorePropertySource extends AwsPropertySource<ParameterSto
 			if (propertiesType) {
 				Properties props;
 				if (prefixType.equals("properties")) {
- 					props = readProperties(parameter.value());
-				} else {
+					props = readProperties(parameter.value());
+				}
+				else {
 					props = readYaml(parameter.value());
 				}
 				for (Map.Entry<Object, Object> entry : props.entrySet()) {
@@ -156,11 +155,13 @@ public class ParameterStorePropertySource extends AwsPropertySource<ParameterSto
 		prefixIndex = context.indexOf(PREFIX_PROPERTIES_LOAD);
 		if (prefixIndex != -1) {
 			this.propertiesType = true;
-			String extracted =  context.substring(prefixIndex + PREFIX_PROPERTIES_LOAD.length()).toLowerCase();
+			String extracted = context.substring(prefixIndex + PREFIX_PROPERTIES_LOAD.length()).toLowerCase();
 			if (ALLOWED_TYPES.contains(extracted)) {
 				this.prefixType = extracted;
-			} else {
-				throw new IllegalArgumentException("Invalid prefixType: " + extracted + ". Must be one of properties, json, or yaml.");
+			}
+			else {
+				throw new IllegalArgumentException(
+						"Invalid prefixType: " + extracted + ". Must be one of properties, json, or yaml.");
 			}
 			return context.substring(0, prefixIndex);
 		}
@@ -181,9 +182,7 @@ public class ParameterStorePropertySource extends AwsPropertySource<ParameterSto
 
 	private Properties readYaml(String input) {
 		YamlPropertiesFactoryBean yaml = new YamlPropertiesFactoryBean();
-		yaml.setResources(
-			new ByteArrayResource(input.getBytes())
-		);
+		yaml.setResources(new ByteArrayResource(input.getBytes()));
 		return yaml.getObject();
 	}
 
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/config/EndpointRegistrar.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/config/EndpointRegistrar.java
@@ -15,7 +15,6 @@
  */
 package io.awspring.cloud.sqs.config;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
 import io.awspring.cloud.sqs.ConfigUtils;
 import io.awspring.cloud.sqs.listener.MessageListenerContainer;
 import io.awspring.cloud.sqs.listener.MessageListenerContainerRegistry;
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/AbstractContainerOptions.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/AbstractContainerOptions.java
@@ -255,10 +255,12 @@ public abstract class AbstractContainerOptions<O extends ContainerOptions<O, B>,
 		private static MessagingMessageConverter<?> createDefaultMessageConverter() {
 			if (JacksonPresent.isJackson3Present()) {
 				return new SqsMessagingMessageConverter();
-			} else if (JacksonPresent.isJackson2Present()) {
+			}
+			else if (JacksonPresent.isJackson2Present()) {
 				return new LegacyJackson2SqsMessagingMessageConverter();
 			}
-			throw new IllegalStateException("Sqs integration requires a Jackson 2 or Jackson 3 library on the classpath");
+			throw new IllegalStateException(
+					"Sqs integration requires a Jackson 2 or Jackson 3 library on the classpath");
 		}
 
 		private static final AcknowledgementMode DEFAULT_ACKNOWLEDGEMENT_MODE = AcknowledgementMode.ON_SUCCESS;
--- a/spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration/SqsMessageConversionIntegrationTests.java
+++ b/spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration/SqsMessageConversionIntegrationTests.java
@@ -28,8 +28,8 @@ import io.awspring.cloud.sqs.listener.SqsHeaders;
 import io.awspring.cloud.sqs.listener.interceptor.AsyncMessageInterceptor;
 import io.awspring.cloud.sqs.operations.SqsTemplate;
 import io.awspring.cloud.sqs.support.converter.MessagingMessageHeaders;
-import io.awspring.cloud.sqs.support.converter.legacy.LegacyJackson2MessageConverterMigration;
 import io.awspring.cloud.sqs.support.converter.SqsMessagingMessageConverter;
+import io.awspring.cloud.sqs.support.converter.legacy.LegacyJackson2MessageConverterMigration;
 import java.time.Duration;
 import java.util.Collections;
 import java.util.List;
@@ -429,8 +429,8 @@ class SqsMessageConversionIntegrationTests extends BaseSqsIntegrationTest {
 
 		@Bean
 		SqsListenerConfigurer customizer(ObjectMapper objectMapper) {
-			return registrar -> registrar.setJacksonMessageConverterMigration(
-					new LegacyJackson2MessageConverterMigration(objectMapper));
+			return registrar -> registrar
+					.setJacksonMessageConverterMigration(new LegacyJackson2MessageConverterMigration(objectMapper));
 		}
 
 		@Bean


spring-cloud-aws-sns/src/main/java/io/awspring/cloud/sns/core/JsonStringEncoderDelegator.java
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.awspring.cloud.sns.core;
+
+import io.awspring.cloud.core.support.JacksonPresent;
+
+public interface JsonStringEncoderDelegator {
```
maciejwalkowiak: Wouldn't name `JsonStringEncoder` suit better?

spring-cloud-aws-sqs/pom.xml
```diff
@@ -42,6 +42,12 @@
 		<dependency>
 			<groupId>com.fasterxml.jackson.core</groupId>
 			<artifactId>jackson-databind</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>tools.jackson.core</groupId>
+			<artifactId>jackson-databind</artifactId>
+			<optional>true</optional>
```
maciejwalkowiak, line 50: This one should not be optional i think?

maciejwalkowiak:
In general great job @MatejNedic! Left two comments plus all new public classes need proper Javadocs

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/AbstractMessageConverterFactory.java
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.awspring.cloud.sqs.support.converter;
+
+import org.springframework.messaging.converter.MessageConverter;
+
+public abstract class AbstractMessageConverterFactory {
```
tomazfernandes: I’m curious: could we use direct instantiation based on classpath detection instead of introducing this factory abstraction?

For instance, Spring Kafka’s MessagingMessageConverter handles Jackson version selection via classpath detection:

https://github.com/spring-projects/spring-kafka/blob/fcca99f6ae3d4f370817cfd1e24f2ae5db93efbf/spring-kafka/src/main/java/org/springframework/kafka/support/converter/MessagingMessageConverter.java#L93-L105

Since we already have `JacksonPresent` too, could we apply the same pattern here so version selection is automatic and users don’t have to manually choose? If there’s a downside or constraint in our case, what is it?

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sqs/SqsAutoConfiguration.java
```diff
@@ -157,13 +158,26 @@ private void configureProperties(SqsContainerOptionsBuilder options) {
 		mapper.from(this.sqsProperties.getListener().getAutoStartup()).to(options::autoStartup);
 	}
 
+	@ConditionalOnMissingBean
+	@Bean
+	public MessagingMessageConverter<Message> messageConverter() {
+		return new SqsMessagingMessageConverter();
```
tomazfernandes: I wonder if this could be simpler if we detect which Jackson version is on the classpath and just return the appropriate converter directly, while still auto-wiring any user-provided mapper. Something like:

```java
@Bean
@ConditionalOnMissingBean
public MessagingMessageConverter<Message> messageConverter(
        ObjectProvider<JsonMapper> jsonMapper,
        ObjectProvider<ObjectMapper> objectMapper) {

    if (JacksonPresent.isJackson3Present()) {
        SqsMessagingMessageConverter c = new SqsMessagingMessageConverter();
        jsonMapper.ifAvailable(c::setJsonMapper);
        return c;
    }
    else if (JacksonPresent.isJackson2Present()) {
        LegacySqsMessagingMessageConverter c = new LegacySqsMessagingMessageConverter();
        objectMapper.ifAvailable(c::setObjectMapper);
        return c;
    }
  }
```

MatejNedic: I will refactor with static classes. Sadly this can lead to ClassNotFoundException if I am not wrong so I will have to add more code.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/AbstractMessageConverterFactory.java
MatejNedic: I was thinking that people can extend AbstractMessageConverterFactory and do their own implementation of MessageConverter. Also, it is much more readable when I refactored to this way considering deletions and so on.

tomazfernandes: Hey @MatejNedic, I understand what you're aiming at, and we already support users providing their own `MessageConverters` through the `SqsListenerConfigurer`.

As far as I can tell, the only place we're using the `create()` method from `AbstractMessageConverterFactory` is in `AbstractListenerAnnotationBeanPostProcessor.java:345`.

I think we could just detect the Jackson version on the classpath there and decide what converter to use automatically.

This might eliminate the need for `AbstractMessageConverterFactory`, `JacksonJsonMessageConverterFactory`, and `LegacyJackson2MessageConverterFactory`, and make overall converter configuration simpler, while still preserving the customization options for users.

What do you think?

MatejNedic: Hey @tomazfernandes , I was thinking about it but I came up with following scenario.

What if you have Jackson 3 and Jackson 2 on classpath. Since this is messaging I want to keep Jackson 2 for SQS integration. currently if we put JacksonPresent users are not able to change this behaviour. 


Notice inside of `SqsListenerAnnotationBeanPostProcessor` we are also based on instace getting ObjectMapper or JsonMapper. I want to allow users to control this. I don't also want to polute `EndpointRegistar` with ObjectMapper and JsonMapper (we had objectMapper there). So for me best way was to do a wrapper and then extract either Jackson 2 or Jackson 3 based on whichever Bean users defined. This allows users themselves to have on Classpath Jackson 3 and 2 and opt in for 2.


```
if (wrapper instanceof JacksonJsonMessageConverterFactory) {
			argumentResolvers.add(new NotificationMessageArgumentResolver(messageConverter,
					((JacksonJsonMessageConverterFactory) wrapper).getJsonMapperWrapper().getJsonMapper()));
			argumentResolvers.add(new NotificationSubjectArgumentResolver(
					((JacksonJsonMessageConverterFactory) wrapper).getJsonMapperWrapper().getJsonMapper()));
			argumentResolvers.add(new SnsNotificationArgumentResolver(messageConverter,
					((JacksonJsonMessageConverterFactory) wrapper).getJsonMapperWrapper().getJsonMapper()));
		}
		else if (wrapper instanceof LegacyJackson2MessageConverterFactory) {
			argumentResolvers.add(new LegacyJackson2NotificationMessageArgumentResolver(messageConverter,
					((LegacyJackson2MessageConverterFactory) wrapper).getObjectMapper()));
			argumentResolvers.add(new LegacyJackson2NotificationSubjectArgumentResolver(
					((LegacyJackson2MessageConverterFactory) wrapper).getObjectMapper()));
			argumentResolvers.add(new LegacyJackson2SnsNotificationArgumentResolver(messageConverter,
					((LegacyJackson2MessageConverterFactory) wrapper).getObjectMapper()));
		}
```
		
		
		@tomazfernandes Wdyt?

tomazfernandes: > I want to allow users to control this. 

If the goal is to enable users to force Jackson 2 over Jackson 3 when both are present, a more consistent approach would be exposing this as an option on the `EndpointRegistrar`, so users can explicitly toggle it via `SqsListenerConfigurer`. For example `JacksonVersion` {`AUTO`, `JACKSON_2`, `JACKSON_3`} or something similar. We can also expose the same option as an auto-configuration property if we’d like.

This way we rely on the existing configuration path instead of introducing a new one, and we don’t require users to declare a bean just to flip a configuration toggle. It also makes the selection explicit user intent and keeps the wiring simpler.

If it helps unblock the release we can make this PR focused on the default behavior only (no factory, detect Jackson 2 vs 3, if both are present default to Jackson 3), and add the explicit opt-in toggle in a separate PR.

Let me know your thoughts.

tomazfernandes: @maciejwalkowiak since `AbstractMessageConverterFactory` introduces a new public API surface and diverges from the existing SQS configuration path, I suggest we treat it as temporary for RC1: mark it @Deprecated(forRemoval = true) and avoid documenting it as a supported API for now. Then we can align on the long-term direction and adjust for GA.

Let me know if that works for RC1.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/AbstractListenerAnnotationBeanPostProcessor.java
```diff
@@ -15,14 +15,11 @@
...
 package io.awspring.cloud.sqs.annotation;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
 import io.awspring.cloud.sqs.ConfigUtils;
-import io.awspring.cloud.sqs.config.Endpoint;
-import io.awspring.cloud.sqs.config.EndpointRegistrar;
-import io.awspring.cloud.sqs.config.HandlerMethodEndpoint;
-import io.awspring.cloud.sqs.config.SqsEndpoint;
-import io.awspring.cloud.sqs.config.SqsListenerConfigurer;
+import io.awspring.cloud.sqs.config.*;
```
tomazfernandes: Nit: remove wildcard

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/AbstractMessagingMessageConverter.java
```diff
@@ -15,23 +15,16 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.lang.Nullable;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageHeaders;
-import org.springframework.messaging.converter.CompositeMessageConverter;
-import org.springframework.messaging.converter.MappingJackson2MessageConverter;
-import org.springframework.messaging.converter.MessageConverter;
-import org.springframework.messaging.converter.StringMessageConverter;
+import org.springframework.messaging.converter.*;
```
tomazfernandes: Nit: remove wildcard.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/resolver/legacy/LegacyJackson2NotificationMessageArgumentResolver.java
```diff
@@ -0,0 +1,52 @@
+/*
```
tomazfernandes, line 1: nit: folder name is misspelled.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/AbstractMessageConverterFactory.java
MatejNedic: Thing is I wouldn't then include it in RC1 since it has to be close to release as possible. 

Lets rethink this but if we go Tomaz with your proposal we are missing that ObjectMapper can be set later in lifecycle on MessageConverter as we had in 3.x.x 

If you are fine with excluding that I can completely rework this for RC1.

tomazfernandes: Hey @MatejNedic, can you share the exact 3.x scenario you're referring to, ideally a snippet?

Thanks.

tomazfernandes: >  I can completely rework this for RC1.

@MatejNedic, appreciate you reworking this for RC1.

For the `set later` behavior with ObjectMapper, we can keep the 3.x wiring, with both Jackson 2's ObjectMapper (when present) alongside Jackson's 3 JsonMapper in the `EndpointRegistrar`, and build the appropriate converter.

Let me know if you run into any issues.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/SqsListenerAnnotationBeanPostProcessor.java
```diff
@@ -106,12 +111,23 @@ protected Collection<HandlerMethodArgumentResolver> createAdditionalArgumentReso
...
 	@Override
 	protected Collection<HandlerMethodArgumentResolver> createAdditionalArgumentResolvers(
-			MessageConverter messageConverter, ObjectMapper objectMapper) {
+			MessageConverter messageConverter, AbstractMessageConverterFactory wrapper) {
 		List<HandlerMethodArgumentResolver> argumentResolvers = new ArrayList<>(createAdditionalArgumentResolvers());
-		if (objectMapper != null) {
-			argumentResolvers.add(new NotificationMessageArgumentResolver(messageConverter, objectMapper));
-			argumentResolvers.add(new NotificationSubjectArgumentResolver(objectMapper));
-			argumentResolvers.add(new SnsNotificationArgumentResolver(messageConverter, objectMapper));
+		if (wrapper instanceof JacksonJsonMessageConverterFactory) {
```
tomazfernandes: How about instead of having these instanceof calls, we add a `createAdditionalArgumentResolvers` method in the Factory?

This way we keep Jackson2 and Jackson3-specific code encapsulated and avoid having a CNFE from loading the specific classes.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/AbstractMessageConverterFactory.java
tomazfernandes: @MatejNedic, after looking more closely at the classpath and classloading implications, I think this path could introduce CNFE/classpath edge cases across certain dependency combinations, including native/AOT setups.

I also think the current implementation has similar risk: some code paths (for example, instanceof checks against Jackson-specific impl classes) may trigger eager loading/linking of those impls, which can fail when the relevant Jackson dependency is not present.

If timeline allows, I’d prefer releasing this as an M2 so we have room to iterate and gather feedback before RC/GA. If RC1 is the priority, I’m ok shipping as-is and then reassessing for an RC2 if needed.

maciejwalkowiak: @tomazfernandes let's release it as RC1. I believe this solution is good as it is even for GA - the mentioned factory class should be deprecated from the start as this is anyway only intermediary solution for backward compatibility with Jackson 2. 

Of course, if we have ideas and capacity to implement a better one, we could do it, but seems like time flies faster than we are able to catch up.

Are you able to give more specific places in code that can introduce CNFE? (I don't see it but i might have missed it).

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/resolver/legacy/LegacyJackson2NotificationMessageArgumentResolver.java
MatejNedic: Hey @tomazfernandes what would be folder name? I am putting all legacy jackson2 into legacy folder. So later we can just delete folders and that is it

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sqs/SqsAutoConfiguration.java
```diff
@@ -135,18 +144,13 @@ public SqsMessageListenerContainerFactory<Object> defaultSqsListenerContainerFac
 		return factory;
 	}
 
-	private void setMapperToConverter(MessagingMessageConverter<?> messagingMessageConverter, ObjectMapper om) {
-		if (messagingMessageConverter instanceof SqsMessagingMessageConverter sqsConverter) {
-			sqsConverter.setObjectMapper(om);
+	private void setMapperToConverter(MessagingMessageConverter<?> messagingMessageConverter,
+			AbstractMessageConverterFactory factory) {
+		if (messagingMessageConverter instanceof LegacyJackson2SqsMessagingMessageConverter sqsConverter) {
```
tomazfernandes: This branch can still fail with `NoClassDefFoundError` in Jackson3-only setups (or AOT/native) because the `instanceof LegacyJackson2SqsMessagingMessageConverter` may trigger resolution/linking of the Jackson2 converter type when Jackson2 isn’t on the runtime classpath.

I suggest guarding before the instanceof:
```java
if (JacksonPresent.isJackson2Present()
        && messagingMessageConverter instanceof LegacyJackson2SqsMessagingMessageConverter sqsConverter) {
    sqsConverter.setObjectMapper(((LegacyJackson2MessageConverterFactory) factory).getObjectMapper());
}
```

Also, let's create a similar path for Jackson 3 (see comment for the converter below).

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sqs/SqsAutoConfiguration.java
```diff
@@ -157,13 +161,48 @@ private void configureProperties(SqsContainerOptionsBuilder options) {
...
-	public SqsListenerConfigurer objectMapperCustomizer(ObjectProvider<ObjectMapper> objectMapperProvider) {
-		ObjectMapper objectMapper = objectMapperProvider.getIfUnique();
+	public SqsListenerConfigurer objectMapperCustomizer(
+			ObjectProvider<AbstractMessageConverterFactory> objectProviderWrapper) {
+		AbstractMessageConverterFactory wrapper = objectProviderWrapper.getIfUnique();
 		return registrar -> {
-			// Object Mapper for SqsListener annotations handler method
-			if (registrar.getObjectMapper() == null && objectMapper != null) {
-				registrar.setObjectMapper(objectMapper);
+			if (wrapper != null) {
```
tomazfernandes: We should also check if the registrar doesn't have a user-defined wrapper before assigning this one to it.

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sqs/SqsAutoConfiguration.java
```diff
@@ -114,7 +122,8 @@ public SqsMessageListenerContainerFactory<Object> defaultSqsListenerContainerFac
 			ObjectProvider<AsyncMessageInterceptor<Object>> asyncInterceptors,
 			ObjectProvider<ObservationRegistry> observationRegistry,
 			ObjectProvider<SqsListenerObservation.Convention> observationConventionProvider,
-			ObjectProvider<MessageInterceptor<Object>> interceptors, ObjectProvider<ObjectMapper> objectMapperProvider,
+			ObjectProvider<MessageInterceptor<Object>> interceptors,
+			ObjectProvider<AbstractMessageConverterFactory> objectMapperProvider,
```
tomazfernandes: Variable name is inconsistent since this is not an ObjectMapper anymore.

spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/sqs/SqsAutoConfigurationTest.java
```diff
@@ -202,6 +203,7 @@ void withObservationDisabled() {
 		});
 	}
 
+	@Disabled("We should see what we want with this test!")
```
tomazfernandes: Not sure I understand the question - why did you need to disable it?

Why not have a `JsonMapperConfiguration`?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/AbstractListenerAnnotationBeanPostProcessor.java
```diff
@@ -335,7 +343,9 @@ protected CompositeMessageConverter createCompositeMessageConverter() {
 		List<MessageConverter> messageConverters = new ArrayList<>();
 		messageConverters.add(new StringMessageConverter());
 		messageConverters.add(new SimpleMessageConverter());
-		messageConverters.add(createDefaultMappingJackson2MessageConverter(this.endpointRegistrar.getObjectMapper()));
+		if (endpointRegistrar.getAbstractMessageConverterFactory() != null) {
```
tomazfernandes: SQS needs to work standalone without Spring Boot and autoconfiguration, so we can't rely on having a wrapper here to offer a working OOTB solution.

My suggestion is we branch here based on `JacksonPresent` and instantiate the corresponding default wrapper if the registrar doesn't have one.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/config/JacksonAbstractMessageConverterFactory.java
```diff
@@ -0,0 +1,71 @@
...
+import org.springframework.messaging.converter.MappingJackson2MessageConverter;
+import tools.jackson.databind.json.JsonMapper;
+
+/**
+ * Factory util class to construct {@link org.springframework.messaging.converter.MessageConverter} either Jackson 2 or
+ * Jackson 3 specific.
+ * @author Matej Nedic
+ * @since 4.0.0
+ */
+public class JacksonAbstractMessageConverterFactory {
```
tomazfernandes: This class references both Jackson 2 and Jackson 3 classes, which might introduce `NoClassDefFoundError` if either is absent in the classpath.

Suggestion: keep the logic directly where they're used, or split into two separate classes for each Jackson version and make sure they're only accessed after a JacksonPresent check.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
```diff
@@ -649,6 +650,10 @@ private static SqsMessagingMessageConverter createDefaultMessageConverter() {
 		return new SqsMessagingMessageConverter();
 	}
 
+	private static LegacyJackson2SqsMessagingMessageConverter createDefaultLegacyJackson2MessageConverter() {
```
tomazfernandes: Let's avoid the hard reference to Jackson 2 / Jackson 3 specific classes in the method signature, which might trigger class loading/linking of the classes and trigger `NoClassDefFoundError` if the jackson dependency is not in the classpath.

Since these are private methods only used in one place, I suggest we move the instantiation to the places they're actually being used instead.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
```diff
@@ -750,6 +755,17 @@ public SqsTemplateBuilder configureDefaultConverter(
 			return this;
 		}
 
+		@Override
+		public SqsTemplateBuilder configureLegacyJackson2Converter(
```
tomazfernandes: From an API design standpoint, exposing this builder method means we’re offering users an entry point that can fail at runtime when the legacy Jackson 2 dependency isn’t present (at class load time or when invoked), without compile-time guarantees.

Let’s document this constraint explicitly in the interface javadocs and in the SQS reference docs, and add a guard clause here, e.g.:

`Assert.state(JacksonPresent.isJackson2Present(), "...");`

Same for the Jackson 3 path.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplateBuilder.java
```diff
@@ -51,6 +52,15 @@ public interface SqsTemplateBuilder {
 	 */
 	SqsTemplateBuilder configureDefaultConverter(Consumer<SqsMessagingMessageConverter> messageConverterConfigurer);
 
+	/**
+	 * Configure the default message converter.
+	 *
+	 * @param messageConverterConfigurer a {@link LegacyJackson2SqsMessagingMessageConverter} consumer.
+	 * @return the builder.
+	 */
+	SqsTemplateBuilder configureLegacyJackson2Converter(
```
tomazfernandes: See comment about the javadoc above.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/AbstractMessageConverterFactory.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Converter factory used to create MessageConverter either for Jackson 2 or Jackson 3. Used also to provide
+ * Implementation for Jackson 3 {@link JacksonJsonMessageConverterFactory} Implementation for Jackson 2
+ * {@link LegacyJackson2MessageConverterFactory} Which provide either: {@link ObjectMapper} or {@link JsonMapper} for
+ * SQS integration. Note that you can declare either of implementations as a Bean which will be picked by
+ * autoconfiguration.
+ * @author Matej Nedic
+ * @since 4.0.0
+ */
+public interface AbstractMessageConverterFactory {
```
tomazfernandes: Since this is now an interface and does more than create a `MessageConverter`, the Abstract / Factory naming is misleading. 

Since this is an interim Jackson2 compatibility surface, please rename it to something Jackson and migration-scoped (e.g. JacksonMigrationHelper) and mark it @Deprecated(forRemoval=true) with javadoc like “helper to assist with Jackson 3 migration, will be removed”.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/AbstractMessagingMessageConverter.java
```diff
@@ -51,7 +45,7 @@ public abstract class AbstractMessagingMessageConverter<S> implements ContextAwa
 
 	private String typeHeader = SqsHeaders.SQS_DEFAULT_TYPE_HEADER;
 
-	private MessageConverter payloadMessageConverter;
+	public MessageConverter payloadMessageConverter;
```
tomazfernandes: Please move this back to private and pass it through the constructor instead of setting it directly in the subclass.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/AbstractMessagingMessageConverter.java
```diff
@@ -237,31 +207,16 @@ private MessageHeaders getMessageHeaders(Message<?> message) {
...
-	private CompositeMessageConverter createDefaultCompositeMessageConverter() {
-		List<MessageConverter> messageConverters = new ArrayList<>();
-		messageConverters.add(createClassMatchingMessageConverter());
-		messageConverters.add(createStringMessageConverter());
-		messageConverters.add(createDefaultMappingJackson2MessageConverter());
-		return new CompositeMessageConverter(messageConverters);
-	}
-
-	private SimpleClassMatchingMessageConverter createClassMatchingMessageConverter() {
+	public SimpleClassMatchingMessageConverter createClassMatchingMessageConverter() {
```
tomazfernandes: The logic for `createDefaultCompositeMessageConverter` should remain in the `AbstractMessagingMessageConverter`.

Each subclass can have a "createDefaultJacksonMessageConverter" implementation, and either have the logic for creating the specific Jackson converter there, or use a specific `JacksonAbstractMessageConverterFactory` for the corresponding Jackson version.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/AbstractMessagingMessageConverter.java
```diff
@@ -87,29 +80,6 @@ public void setPayloadMessageConverter(MessageConverter messageConverter) {
 		this.payloadMessageConverter = messageConverter;
 	}
 
-	/**
-	 * Set the {@link ObjectMapper} instance to be used for converting the {@link Message} instances payloads.
-	 * @param objectMapper the object mapper instance.
-	 */
-	public void setObjectMapper(ObjectMapper objectMapper) {
```
tomazfernandes, line 104: With this diff we're removing the ability for the user to set their own mapper in the converter through direct configuration.

I suggest that instead we make it so that if the user sets a `JsonMapper`, we'll use that to create a new `JacksonJsonMessageConverter` with the default settings, and then follow a similar path to the current one.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/AbstractMessageConverterFactory.java
tomazfernandes: @maciejwalkowiak agreed on RC1.

While Matej's refactor addressed the instanceof CNFE concerns, I've made some review comments that point out a few remaining points.

> the mentioned factory class should be deprecated from the start as this is anyway only intermediary solution for backward compatibility with Jackson 2

Agreed. Let's reflect that in the code by marking it as deprecated and renaming it to something more migration-scoped to make the long-term direction clearer.

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sqs/SqsAutoConfiguration.java
MatejNedic: I have on purpose made interface public. I wanted users to be able to extend and implement their own implementation. If they want to use different converter or something like that.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/config/JacksonAbstractMessageConverterFactory.java
MatejNedic: AFAIK this shouldn't cause this error but I will split for clarity makes sense

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/AbstractMessagingMessageConverter.java
MatejNedic: User should pass objectMapper through the factory. Since now Factory objectMapper is used to be set later on MessageConverter.

Thing is I don't want in this class anything Jackson 2 or Jackson 3 specific. That is why I went with Factroy approach to completely remove any Jackson 2 or 3 classes from here.

In JacksonJsonMessageConverter for Jacskon 3 we cannot set JsonMapper after Converter is created that is why I moved it to factory creation as well so JsonMapper user passed is respected.

MatejNedic:
@tomazfernandes Thanks on review will apply the changes today! Please check the comments I left.

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sqs/SqsAutoConfiguration.java
MatejNedic: I thought of better way.

Once I push refactor lets check together design and see if you like it :)

spring-cloud-aws-autoconfigure/src/test/java/io/awspring/cloud/autoconfigure/sqs/SqsAutoConfigurationTest.java
MatejNedic: Since in configuresFactoryComponentsAndOptions we were checking if user provided `objectMapper` was set. 
Issue is in Spring Boot 4 Jackson 3 MessageConverter cannot accept JsonMapper after it was created. Only through the constructor. So I disabled the test.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
MatejNedic: I will create enum to pass. With enum then we can decide to either configure one or another.

Wdyt @tomazfernandes

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
MatejNedic: We can do that but check comment above you made. I am not sure how to approach this. Let me think on it few hours but enum is only way I see.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/AbstractMessagingMessageConverter.java
MatejNedic: I want to reuse them because they are used in composite. I am not seeing the difference tbh. 

I refactored mb this will make more sense please check.

tomazfernandes (changes requested):
Thanks Matej. I’ll take a look.

SQS acceptance criteria for this PR:
	•	CNFE/classpath risks are mitigated
	•	No new long-term supported user-facing APIs or extension points are introduced as part of the migration
	•	No loss of existing user functionality

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/JacksonMessageConverterFactory.java
```diff
@@ -0,0 +1,54 @@
...
+ * Converter factory used to create MessageConverter either for Jackson 2 or Jackson 3. Used also to provide
+ * Implementation for Jackson 3 {@link JacksonJsonMessageConverterFactory} Implementation for Jackson 2
+ * {@link LegacyJackson2MessageConverterFactory} Which provide either: {@link ObjectMapper} or {@link JsonMapper} for
+ * SQS integration. Note that you can declare either of implementations as a Bean which will be picked by
+ * autoconfiguration.
+ * To be removed with Jackson 2 support
+ * @author Matej Nedic
+ * @since 4.0.0
+ */
+public interface JacksonMessageConverterFactory {
```
tomazfernandes: As previously aligned, for SQS, we don’t intend to support this factory type as a long-term API. This should be a migration-only hook.

Please rename it to something migration-specific, annotate it with @Deprecated(forRemoval = true), and add migration-only Javadoc.

spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sqs/SqsAutoConfiguration.java
```diff
@@ -157,13 +155,48 @@ private void configureProperties(SqsContainerOptionsBuilder options) {
...
-	public SqsListenerConfigurer objectMapperCustomizer(ObjectProvider<ObjectMapper> objectMapperProvider) {
-		ObjectMapper objectMapper = objectMapperProvider.getIfUnique();
+	public SqsListenerConfigurer objectMapperCustomizer(
+			ObjectProvider<JacksonMessageConverterFactory> factoryProvider) {
+		JacksonMessageConverterFactory factory = factoryProvider.getIfUnique();
 		return registrar -> {
-			// Object Mapper for SqsListener annotations handler method
-			if (registrar.getObjectMapper() == null && objectMapper != null) {
-				registrar.setObjectMapper(objectMapper);
+			if (factory != null) {
```
tomazfernandes: We should check if the user has set a wrapper to the registrar already before overriding it.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
tomazfernandes: Not sure what you mean, can you be more specific about how this enum would work?

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
tomazfernandes: I don't see how an enum could solve this public API issue, can you be more specific?

If nothing comes to mind, my suggestions stands - we can add a guard clause and document it.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/AbstractMessagingMessageConverter.java
tomazfernandes: > User should pass objectMapper through the factory. Since now Factory objectMapper is used to be set later on MessageConverter.

This is not true for user-instantiated converters, users need a way to manually set it.

> Thing is I don't want in this class anything Jackson 2 or Jackson 3 specific.

We can move this method (and the one for the JsonMapper) to the respective subclasses.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/AbstractMessagingMessageConverter.java
```diff
@@ -237,31 +208,16 @@ private MessageHeaders getMessageHeaders(Message<?> message) {
...
-	private CompositeMessageConverter createDefaultCompositeMessageConverter() {
-		List<MessageConverter> messageConverters = new ArrayList<>();
-		messageConverters.add(createClassMatchingMessageConverter());
-		messageConverters.add(createStringMessageConverter());
-		messageConverters.add(createDefaultMappingJackson2MessageConverter());
-		return new CompositeMessageConverter(messageConverters);
-	}
-
-	private SimpleClassMatchingMessageConverter createClassMatchingMessageConverter() {
+	public static SimpleClassMatchingMessageConverter createClassMatchingMessageConverter() {
```
tomazfernandes: Please move this method back to private so it doesn't create a public-facing API.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/AbstractMessagingMessageConverter.java
```diff
@@ -237,31 +208,16 @@ private MessageHeaders getMessageHeaders(Message<?> message) {
...
 		SimpleClassMatchingMessageConverter matchingMessageConverter = new SimpleClassMatchingMessageConverter();
 		matchingMessageConverter.setSerializedPayloadClass(String.class);
 		return matchingMessageConverter;
 	}
 
-	private StringMessageConverter createStringMessageConverter() {
+	public static StringMessageConverter createStringMessageConverter() {
```
tomazfernandes: Please move this method back to private so it doesn't create a public-facing API.

tomazfernandes (changes requested):
@MatejNedic, requesting a couple more changes.

Since this is a migration-scoped PR, let’s not introduce new long-term user-facing APIs or extension points. For SQS, we don’t plan to support `JacksonMessageConverterFactory` long term, so it should be scoped as migration-only.

So, please:
• rename it to something migration-scoped, and
• annotate it with @Deprecated(forRemoval = true) plus migration-only Javadoc.

Once updated, I’ll take another look.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
MatejNedic: configureLegacyJackson2Converter is referencing new LegacyJackson2SqsMessagingMessageConverter which should be used with Jackson 2. Now having two configureDefaultConverter one for SqsMessagingMessageConverter and other for LegacyJackson2SqsMessagingMessageConverter in SqsTempalte I don't think is best option.

From my perspective I was thinking to go with something like this ->

```
@Override
		public SqsTemplateBuilder configureDefaultConverter(
			Consumer<AbstractMessagingMessageConverter> messageConverterConfigurer) {
			return configureDefaultConverter(messageConverterConfigurer, null);
		}

		@Override
		public SqsTemplateBuilder configureDefaultConverter(
				Consumer<AbstractMessagingMessageConverter> messageConverterConfigurer, SqsJacksonVersion sqsJacksonVersion) {
			Assert.notNull(messageConverterConfigurer, "messageConverterConfigurer must not be null");
			Assert.isNull(this.messageConverter, "messageConverter already configured");
			AbstractMessagingMessageConverter defaultMessageConverter;
			if (sqsJacksonVersion == SqsJacksonVersion.JACKSON_3) {
				defaultMessageConverter = new SqsMessagingMessageConverter();
			} else if (sqsJacksonVersion == SqsJacksonVersion.JACKSON_2)  {
				defaultMessageConverter = new LegacyJackson2SqsMessagingMessageConverter();
			} else {
				defaultMessageConverter = createDefaultBasedOnJacksonPresent();
			}
			messageConverterConfigurer.accept(defaultMessageConverter);
			this.messageConverter = defaultMessageConverter;
			return this;
		}
```


Or maybe we can duplicate SqsTemplate. Jackson 3 SqsTemplate and Jacskon 2 LegacyJackson2SqsTemplate
all based on SqsOperations.

Wdyt @tomazfernandes

MatejNedic:
@tomazfernandes I pushed sample of SQSTemplate change. Please check it and let me know what approach would you like.

I have applied other changes as you suggested!

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
```diff
@@ -741,15 +740,29 @@ public SqsTemplateBuilder messageConverter(MessagingMessageConverter<Message> me
 
 		@Override
 		public SqsTemplateBuilder configureDefaultConverter(
-				Consumer<SqsMessagingMessageConverter> messageConverterConfigurer) {
+			Consumer<AbstractMessagingMessageConverter> messageConverterConfigurer) {
+			return configureDefaultConverter(messageConverterConfigurer, null);
+		}
+
+		@Override
+		public SqsTemplateBuilder configureDefaultConverter(
```
tomazfernandes, line 748: I think this is an interesting solution, let's go with it.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/JacksonMessageConverterFactoryAndEnricher.java
```diff
@@ -0,0 +1,55 @@
...
+ * Implementation for Jackson 3 {@link JacksonJsonMessageConverterFactoryAndEnricher} Implementation for Jackson 2
+ * {@link LegacyJackson2MessageConverterFactoryAndEnricher} Which provide either: {@link ObjectMapper} or {@link JsonMapper} for
+ * SQS integration. Note that you can declare either of implementations as a Bean which will be picked by
+ * autoconfiguration.
+ * NOTE this is transition only api.
+ * @author Matej Nedic
+ * @since 4.0.0
+ */
+@Deprecated(forRemoval = true)
+public interface JacksonMessageConverterFactoryAndEnricher {
```
tomazfernandes: *> No new long-term supported user-facing APIs or extension points are introduced as part of the migration*

This name is too broad for a migration-only API. Suggest renaming to `JacksonMigrationHelper` (or `JacksonMessageConverterMigration`) to make the transitional scope explicit. The current name reads like a long-term extension point.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/JacksonJsonMessageConverterFactoryAndEnricher.java
```diff
@@ -0,0 +1,74 @@
...
+import tools.jackson.databind.json.JsonMapper;
+
+/**
+ * Used to create {@link JacksonJsonMessageConverter} and provide {@link JsonMapper} to
+ * {@link SqsListenerAnnotationBeanPostProcessor}.
+ *
+ * @author Matej Nedic
+ * @since 4.0.0
+ */
+public class JacksonJsonMessageConverterFactoryAndEnricher implements JacksonMessageConverterFactoryAndEnricher {
```
tomazfernandes: *> No new long-term supported user-facing APIs or extension points are introduced as part of the migration*

Please add `@Deprecated(forRemoval=true)` and add to the javadocs that this is a internal, temporary implementation for the migration.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/JacksonMessageConverterFactoryAndEnricher.java
```diff
@@ -0,0 +1,55 @@
...
+
+	MessageConverter create();
+
+	/**
+	 * Used by {@link SqsListenerAnnotationBeanPostProcessor} to add resolvers which will be used when resolving
+	 * message.
+	 * @param argumentResolvers List of argument resolvers
+	 * @param messageConverter MessageConverters which will be used by resolvers.
+	 */
+	void enrichResolvers(List<HandlerMethodArgumentResolver> argumentResolvers, MessageConverter messageConverter);
```
tomazfernandes: *> No new long-term supported user-facing APIs or extension points are introduced as part of the migration*

Same point here: `enrichResolvers` is very generic for a migration-only API. Please rename it to something intent-revealing and migration-scoped, for example `addJacksonMigrationResolvers(...)`.

Also, please remove the reference to `SqsListenerAnnotationBeanPostProcessor` from the Javadoc (internal infrastructure)and point to `SqsListenerConfigurer` as the supported customization entry point for managing handler method argument resolvers.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/JacksonMessageConverterFactoryAndEnricher.java
```diff
@@ -0,0 +1,55 @@
...
+
+	/**
+	 * Used to enrich {@link MessagingMessageConverter} with Jackson implementation
+	 * @param messageConverter Which will be enriched
+	 */
+	default void enrichMessageConverter(MessagingMessageConverter messageConverter){}
```
tomazfernandes: *> No new long-term supported user-facing APIs or extension points are introduced as part of the migration*

Also very generic name for a migration-only API. In the legacy implementation it only sets the ObjectMapper, so please rename it to something more specific, for example setLegacyObjectMapper(...) (or configureLegacyObjectMapper(...)).

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/JacksonMessageConverterFactoryAndEnricher.java
```diff
@@ -0,0 +1,55 @@
...
+ * SQS integration. Note that you can declare either of implementations as a Bean which will be picked by
+ * autoconfiguration.
+ * NOTE this is transition only api.
+ * @author Matej Nedic
+ * @since 4.0.0
+ */
+@Deprecated(forRemoval = true)
+public interface JacksonMessageConverterFactoryAndEnricher {
+
+	MessageConverter create();
```
tomazfernandes: *> No new long-term supported user-facing APIs or extension points are introduced as part of the migration*

create() is very generic for a migration-only API. Please rename to something more specific, such as `createMigrationMessageConverter()` (or `createMigrationConverter()`) to keep the intent clear.

Also, let's add javadoc stating this is an internal-only, temporary method.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/JacksonMessageConverterFactoryAndEnricher.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.awspring.cloud.sqs.support.converter;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
```
tomazfernandes: *> CNFE/classpath risks are mitigated*

Please remove the Jackson version-specific imports from the Javadocs so we don't accidentally load the classes and risk a `NoClassDefFoundError`.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
```diff
@@ -741,15 +740,29 @@ public SqsTemplateBuilder messageConverter(MessagingMessageConverter<Message> me
...
+				Consumer<AbstractMessagingMessageConverter> messageConverterConfigurer, SqsJacksonVersion sqsJacksonVersion) {
 			Assert.notNull(messageConverterConfigurer, "messageConverterConfigurer must not be null");
 			Assert.isNull(this.messageConverter, "messageConverter already configured");
-			SqsMessagingMessageConverter defaultMessageConverter = createDefaultMessageConverter();
+			AbstractMessagingMessageConverter defaultMessageConverter;
+			if (sqsJacksonVersion == SqsJacksonVersion.JACKSON_3) {
+				defaultMessageConverter = new SqsMessagingMessageConverter();
```
tomazfernandes, line 755: *> CNFE/classpath risks are mitigated*

Let's just check the corresponding version is in the classpath before instantiating the converters.

MatejNedic:
Thnaks @tomazfernandes!

Will implement this changes today and lets get this merged and shipped!

tomazfernandes:
Cool, thanks @MatejNedic.

I’ll re-review after the updates are pushed.

Once the remaining requested changes are addressed, I'll approve the PR and we'll be good to merge.

tomazfernandes:
> No new long-term supported user-facing APIs or extension points are introduced as part of the migration

@MatejNedic just one more small ask for when you get to it: after the rename and javadoc changes, please move `JacksonMessageConverterFactoryAndEnricher` and its implementations into the `legacy` package.

That keeps the migration-only intent clear and makes it easy to remove later.

spring-cloud-aws-secrets-manager/src/main/java/io/awspring/cloud/secretsmanager/JacksonSecretValueReader.java
```diff
@@ -0,0 +1,46 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.awspring.cloud.secretsmanager;
+
+import com.fasterxml.jackson.core.JacksonException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.json.JsonMapper;
```
rafmar-dna: Please use Jackson 3 equivalents

MatejNedic: Good catch thanks @rafmar-dna

MatejNedic:
@tomazfernandes all done!

MatejNedic:
@tomazfernandes I am afraid I don't understand the part about Registrar.

When I was thinking what abstraction to bring I was solving main problem some of our users will have.

Mutliple Jackson versions on classpath. Now this scenario was already brought to me by 1 of our library users.

Imagine you have.

1. Scenario
`SqsListener` -> Jackson 2
`SqsTemplate` -> Jacskon 2
And a third bean `SqsTemplate` -> Jackson 3

2 Scenario. 
Rolling out services slowly firstly drainign queue and then having producer and consumer upgraded.
SqsListener -> Jackson 2
SqsTemplate -> Jacskon 3

This is something I want to be able to achieve. Currently `SqsAutoConfiguration` will provide registrar and `SqsTemplate` with same `MessagingMessageConverter`. They will be completely the same. I want to allow users to diverge and use both Jacksons in same time. 

Autoconfiguration is here only to enforce 2 or 3 not more.

For SqsTemplate
```java
@ConditionalOnMissingBean
@Bean
public SqsTemplate sqsTemplate(SqsAsyncClient sqsAsyncClient,
	ObjectProvider<JacksonMessageConverterMigration> messageConverterFactory,
	ObjectProvider<ObservationRegistry> observationRegistryProvider,
	ObjectProvider<SqsTemplateObservation.Convention> observationConventionProvider,
	MessagingMessageConverter<Message> messageConverter) {
SqsTemplateBuilder builder = SqsTemplate.builder().sqsAsyncClient(sqsAsyncClient)
		.messageConverter(messageConverter);
messageConverterFactory.ifAvailable(mcf -> mcf.configureLegacyObjectMapper(messageConverter));
}
```

For containerFactory

```java
@ConditionalOnMissingBean
@Bean
public SqsMessageListenerContainerFactory<Object> defaultSqsListenerContainerFactory(
        ObjectProvider<SqsAsyncClient> sqsAsyncClient, ObjectProvider<AsyncErrorHandler<Object>> asyncErrorHandler,
        ObjectProvider<ErrorHandler<Object>> errorHandler,
        ObjectProvider<AsyncMessageInterceptor<Object>> asyncInterceptors,
        ObjectProvider<ObservationRegistry> observationRegistry,
        ObjectProvider<SqsListenerObservation.Convention> observationConventionProvider,
        ObjectProvider<MessageInterceptor<Object>> interceptors,
        ObjectProvider<JacksonMessageConverterMigration> messageConverterFactory,
        MessagingMessageConverter<?> messagingMessageConverter) {

    SqsMessageListenerContainerFactory<Object> factory = new SqsMessageListenerContainerFactory<>();
    factory.configure(this::configureProperties);
    sqsAsyncClient.ifAvailable(factory::setSqsAsyncClient);
    asyncErrorHandler.ifAvailable(factory::setErrorHandler);
    errorHandler.ifAvailable(factory::setErrorHandler);
    interceptors.forEach(factory::addMessageInterceptor);
    asyncInterceptors.forEach(factory::addMessageInterceptor);
    messageConverterFactory.ifAvailable(mcf -> mcf.configureLegacyObjectMapper(messagingMessageConverter));
}		
```


JacksonMessageConverterMigration is only used to create defaults which were created before and setObjectMapper. Nothing more.

Tbh I wouldn't put MessagingMessageConverter inside JacksonMessageConverterMigration. Personally since at some point we are going to delete that class and instead of it just put JsonMapper and autoconfiguration will look the same.

Also I am trying to follow here Spring Boot way of autoconfiguration. Only simple scenarios are covered by it such as I want only Jackson 3 or Jackson 2 for both. If you want something custom create the Bean of what you need and tweak it how you want. I think this should be intention of our Autoconfiguration IMHO.

tomazfernandes:
Hey @MatejNedic, thanks for the changes.

As the code you pasted shows, a user-provided MessagingMessageConverter bean is picked up and used by both SqsTemplate and the container factory.

The same MessagingMessageConverter is not wired into the registrar/BPP @SqsListener argument-resolver flow, which creates a discrepancy.

The fix should be straightforward: propagate the selected MessagingMessageConverter into the registrar/BPP path as well, using JacksonMessageConverterMigration as an internal adapter.

This keeps auto-configuration converter customization for the BPP flow and user Jackson 2 vs 3 selection at the MessagingMessageConverter level rather than the temporary migration API, which is what AC #2 requires.

On use cases and auto-configuration, agreed it should stay simple and predictable. The more complex scenarios you mentioned should be handled by users wiring separate beans explicitly.

I'll re-review once the update is pushed. Thanks again.

tomazfernandes:
@MatejNedic quick update on conflicts:

I resolved the current merge conflicts against `awspring/main` in a helper PR: MatejNedic/spring-cloud-aws-1#2.

Scope is conflict resolution only (merge `upstream/main` into the branch and resolve), no intended functional changes beyond that. Tests are green on the SQS module.

I do not have permission to push to your fork branch directly (403), so the easiest path is to merge that helper PR into `Introduce-jackson-3-support`.

Let me know if you hit any issues merging it.

MatejNedic:
Hey @tomazfernandes are we good to go after I merged your changes?

tomazfernandes:
@MatejNedic thanks for merging the conflict resolution PR.

Clarifying my earlier note on `MessagingMessageConverter` wiring: no changes needed for the registrar/BPP path, current wiring is good as-is for this PR.

Non-RC blocking follow-up (to be addressed before GA): some Jackson migration Javadocs and reference docs can read like a long-term supported extension surface. That is at odds with our acceptance criteria (“no new long-term supported user-facing APIs or extension points introduced as part of the migration”). Tracked in [#1550.](https://github.com/awspring/spring-cloud-aws/issues/1550)

MatejNedic:
Thank @tomazfernandes on all the support!

Lets not prolong the release and merge this and do release today.

Plan would be the following: 
Since it is only docs and classes have been marked with deprecate already, I will create an issue to go through Javadocs and documentation for SQS  integration. Will state that that is is only migration api which will be supported till end of 2026. This will be then part of 4.0.0-RC2 release. Since we are not gonna change anything except docs this is not breaking change for RC2 release. I will start the work on it right after we release and give you for the review!

I have got multiple emails about RC1 and jackson 3 support and I want to get this out as soon as possible. Also, API won't change only docs will so I think this should be path forward.

You already created the issue my bad. Merging releasing and getting to work regarding docs!

PR #1609, opened 2026-04-20
Avoid reflective copying in SqsContainerOptions builder

### LINKED ISSUE #1598: Native SQS listener container loses custom messageConverter
**Type**: Bug

  **Component**:
  SQS

  **Describe the bug**
  In GraalVM native image, a custom `MessagingMessageConverter` configured on
  `SqsMessageListenerContainerFactory` is lost before the actual listener container is created.

  Setup:
  - Spring Cloud AWS SQS 4.0.0
  - Spring Boot 4.0.0
  - Java 25
  - GraalVM 25.0.2
  - reproduced locally with LocalStack

  What happens:
  1. Application config customizes the listener container factory using
     `io.awspring.cloud.sqs.config.SqsMessageListenerContainerFactory.Builder#configure(...)`
     and sets a custom converter via
  `io.awspring.cloud.sqs.listener.AbstractContainerOptions.Builder#messageConverter(io.awspring.cloud.sqs.support.converter.MessagingMessageConverter<?>)`.
  2. On JVM, the listener container uses that custom converter, and message processing works.
  3. In the native image, the built listener container ends up using the default
     `SqsMessagingMessageConverter` instead.

  This causes the custom receive-side conversion to be bypassed in native mode.

  In my case:
  - JVM used the custom converter and correctly processed an SNS-wrapped FIFO message
  - native used the default converter and deserialized the wrapper type incorrectly (`payload == null`)

  The problem appears to be in the builder copy path. `SqsContainerOptions.BuilderImpl.createCopy()`
  uses reflective field copying, and in the native image, the superclass state (including `messageConverter`)
  does not seem to survive correctly.

  A local patch that replaced the reflective copy with explicit field copying made native behave like JVM.

  A cleaner upstream fix may be to replace the reflective copy in
  `SqsContainerOptions.BuilderImpl.createCopy()` with explicit builder-state copying
  (for example via copy constructors), instead of relying on
  `ReflectionUtils.shallowCopyFieldState(...)`. A local patch in that direction
  made native image preserve the configured custom `messageConverter`.

  Another possible fix direction would be to make the native runtime hints explicit for the
  builder types involved in the reflective copy, but explicit copying seems like the cleaner
  and a more robust solution.

  **Sample**
  Minimal repro:
  1. Define a custom `MessagingMessageConverter<Message>` bean
  2. Configure `io.awspring.cloud.sqs.config.SqsMessageListenerContainerFactory`
     using `io.awspring.cloud.sqs.config.SqsMessageListenerContainerFactory.Builder#configure(...)`
  3. Set the custom converter via
     `io.awspring.cloud.sqs.listener.AbstractContainerOptions.Builder#messageConverter(io.awspring.cloud.sqs.support.converter.MessagingMessageConverter<?>)`
  4. Add an `@SqsListener`
  5. Run on JVM: custom converter is used
  6. Build native image: listener container uses default `SqsMessagingMessageConverter` instead

  Observed behavior:
  - JVM: `source.convertMessage converter=<custom converter>`
  - native: `source.convertMessage converter=io.awspring.cloud.sqs.support.converter.SqsMessagingMessageConverter`

sondemar:
  - [x] Bugfix
  - [ ] New feature
  - [ ] Enhancement
  - [ ] Refactoring

  ## :scroll: Description
  Replace reflective builder state copying in `SqsContainerOptions` with explicit field copying.

  This change updates the SQS container options builder copy paths to avoid `ReflectionUtils.shallowCopyFieldState(...)` and preserve custom `messageConverter` configuration when options are copied through builder/container creation flows.

  It also adds regression tests for:
  - builder `createCopy()`
  - `fromBuilder(...)`
  - container creation through `SqsMessageListenerContainerFactory`

  This fixes the issue where custom `messageConverter` configuration could be lost or behave incorrectly when `SqsContainerOptions` were copied.

  It also removes reflection from this builder-copy path, which is a better fit for AOT/native scenarios.

  Fixes #1598

  Ran focused SQS tests covering the changed code paths:

  - `mvn -o -pl spring-cloud-aws-sqs -Dtest=ContainerOptionsTests,SqsMessageListenerContainerFactoryTests test`

  Also verified the project still packages successfully with tests skipped:

  - `mvn -o -DskipTests clean package`

  ## :pencil: Checklist
  - [ ] All tests passing
  - [x] No breaking changes

  ## :crystal_ball: Next steps
  None.

### DIFF
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/AbstractContainerOptions.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/AbstractContainerOptions.java
@@ -315,6 +315,10 @@ public abstract class AbstractContainerOptions<O extends ContainerOptions<O, B>,
 		protected Builder() {
 		}
 
+		protected Builder(Builder<?, ?> builder) {
+			copyStateFrom(builder);
+		}
+
 		protected Builder(AbstractContainerOptions<?, ?> options) {
 			this.maxConcurrentMessages = options.maxConcurrentMessages;
 			this.maxMessagesPerPoll = options.maxMessagesPerPoll;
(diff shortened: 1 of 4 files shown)


tomazfernandes (commented):
Hey @sondemar, thanks for the PR.

I wonder what your thoughts are about a simpler approach such as:

```java
Builder(Builder) { this(builder.build()); }
```

This would avoid the copy duplication. The main concern I see would be the firing the `this.maxMessagesPerPoll <= this.maxConcurrentMessages` on an potentially incomplete builder.

What do you think?

sondemar:
> Hey @sondemar, thanks for the PR.
> 
> I wonder what your thoughts are about a simpler approach such as:
> 
> ```java
> Builder(Builder) { this(builder.build()); }
> ```
> 
> This would avoid the copy duplication. The main concern I see would be the firing the `this.maxMessagesPerPoll <= this.maxConcurrentMessages` on an potentially incomplete builder.
> 
> What do you think?

Hi @tomazfernandes, yes, this validation time window is exactly the reason I avoided `Builder(Builder) { this(builder.build()); }`. A builder can be temporarily invalid while it is **still being configured**. For example, `maxMessagesPerPoll(11)` is invalid with the default `maxConcurrentMessages(10)`, but it becomes valid when the caller sets `maxConcurrentMessages(11)`. Calling `build()` inside `createCopy()` would trigger validation **too early**. I kept the builder-to-builder state copying for that reason and added a regression test covering this incomplete-builder scenario. 
We can extract some shared copy logic later, but I would like to keep it separate from this fix since it would be mostly polish and would not remove the need for builder-state copying.

tomazfernandes (approved):
Sounds good @sondemar. Thanks for the PR, looking forward to the next one.

PR #1622, opened 2026-05-12
Feature: SQS Heartbeating

### LINKED ISSUE #1554: SQS: Heartbeat for long-running handlers
**Type**: Feature

Without a heartbeat / keep-alive mechanism, a given message may be picked up multiple times by handlers (if it is long-running with variable time) and executed concurrently.

A new `SqsListener.UseHeartbeat` property that - when set to true - keeps a given sqs message alive by extending its (in)visibility in the background. 

An alternative is to inject an `Visibility` instance and do this manually in each handler.

**Additional context**
Implementing a heartbeat is recommended by AWS. 

> Implement a heartbeat mechanism to periodically extend the visibility timeout, ensuring the message remains invisible until processing is complete.

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html#visibility-timeout-best-practices

As an aside, we've implemented this pattern in C# using a background thread (via `Task.Run()`) that manages heartbeat failures via CancellationTokens. This has proven effective.

kylez-ithaka:

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
Implements / Fixes #1554 : Implement heartbeat for long running tasks. 

This allows a consumer of SQS which doesn't know up front the maximum time it will process a message, to use a "heartbeat" mechanism where it periodically checks back in with SQS to extend the visibility timeout, per AWS best practices https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html

This has _benefits_ over setting a really long maximum visibility timeout
- If process fails long before maximum visibility timeout you set, it reduces latency of the message getting picked back up by a consumer, or moving into the dead letter queue (if configured)
- If process goes beyond the maximum visibility timeout, but is still in progress, it prevents the message from getting picked up by another consumer, which would've caused extra work/cost and potentially other issues (including dead lettering things that actually completed)
- Reduces effort to implement SQS listeners overall, you can set the default visibility timeout for rough expected amount for it to complete processing (or just default of 30 seconds, or your own standard default), as long as you setup the heartbeat to happen before visibility timeout

The _costs_ over setting a really long maximum visibility timeout
- There is overhead of making the heartbeat, it's a rather small overhead individually (simple call to SQS, single future), but if configured too frequently and running large amounts of SQS messages on the same server, it could add up (especially if you under provision for your CPU and misconfigure the heartbeat to happen too frequently)

This is pretty standard practice for SQS, especially for longer or variable time processing. Motivating by having issues with messages ending up in dead letter because timeout wasn't long enough for some items, and caused it to be processed multiple times

Unit tests, and I made a queue and ran it locally, I setup a queue and put messages on it, that normally would timeout but now doesn't. I could see the logs about heart beating as well.

Note this was to test it out, one shouldn't do a heartbeat every 1 second
```java
 @SqsListener(value = "test-sqs-heartbeat-queue", messageVisibilityHeartbeatIntervalSeconds ="1",messageVisibilityHeartbeatSeconds = "30")
    public void listenToTest(String message) {
        LOGGER.info("Start processing {}", message);
        for(int i = 0; i < 10; i++){
            try {
                LOGGER.info("Processing message, task #{} - {}", i, message);
                Thread.sleep(6000);
            } catch (InterruptedException e) {}
        }
        LOGGER.info("End processing {}", message);
    }
```

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

Make some release that has it in it, and include in changelog?

### DIFF
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -815,6 +815,10 @@ See <<FIFO Support>> for more information.
 - `acknowledgementMode` - Set the acknowledgement mode for the container.
 If any value is set, it will take precedence over the acknowledgement mode defined for the container factory options.
 See <<Acknowledgement Mode>> for more information.
+- `messageVisibilityHeartbeatIntervalSeconds` - Set the interval for sending visibility heartbeats for messages that are being processed.
+Note that this is optional and used in conjunction with `messageVisibilityHeartbeatSeconds`.
+- `messageVisibilityHeartbeatSeconds` - Set the visibility timeout extension to be applied in the heartbeats for messages that are being processed.
+Note that this is optional and used in conjunction with `messageVisibilityHeartbeatIntervalSeconds`, and make sure the interval is shorter than this value.
 
 ===== Listener Method Arguments
 
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/SqsListener.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/SqsListener.java
@@ -141,6 +141,16 @@ public @interface SqsListener {
 	String messageVisibilitySeconds() default "";
 
 	/**
+	 * Interval between visibility heartbeat requests sent while listener execution is in progress.
+	 */
+	String messageVisibilityHeartbeatIntervalSeconds() default "";
+
+	/**
+	 * Visibility timeout to apply on each visibility heartbeat request.
+	 */
+	String messageVisibilityHeartbeatSeconds() default "";
+
+	/**
 	 * The acknowledgement mode to be used for the provided queues. If not specified, the acknowledgement mode defined
 	 * for the container factory will be used.
 	 */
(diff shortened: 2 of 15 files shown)


kylez-ithaka:
this would be a type:feature or type:enhancement, not documentation

kylez-ithaka:
@maciejwalkowiak @MatejNedic @tomazfernandes  is there anything I should adjust for this to get it queued up for a Review?

tomazfernandes:
Hi @kylez-ithaka, thanks for the PR.

I've started reviewing it already, will get back to you soon.

tomazfernandes (changes requested):
@kylez-ithaka thanks for the PR.

The main issue I see is that scheduling the heartbeat at intercept level only covers messages that have started processing. For ordered sinks (FIFO queues), messages from the same batch wait for their turn, so their visibility won't be extended when a previous message goes long.

We could instead schedule a single heartbeat for the whole batch in the Sink when the messages are emitted, and use a lightweight interceptor only to signal that the message is done processing and remove it from the heartbeat set.

What do you think?

kylez-ithaka:
@tomazfernandes good catch and that makes sense to be the change, I don't know if i'll get around to that update until a few weeks from now

kylez-ithaka:
I'm starting to update this

Issue #1605, opened 2026-04-16
SendBatchOperationFailedException is created with empty message string


krechanski:
**Type**: Bug

**Component**:
SQS

**Describe the bug**
`AbstractMessagingTemplate.handleFailedSendBatch` creates `SendBatchOperationFailedException` with a hardcoded empty string as the exception message (spring-cloud-aws-sqs 4.0.0):
                                                                                                                   
  //AbstractMessagingTemplate.java                                                                                
 `private <T> CompletableFuture<SendResult.Batch<T>> handleFailedSendBatch(String endpoint,                        
          SendResult.Batch<T> result) {                                                                            
      return CompletableFuture.failedFuture(new SendBatchOperationFailedException("", endpoint, result));          
  }`                                                                                                          
                                                                                                                   
  `e.getMessage()` always returns "", making the exception silent in logs. The failure details (error code,          
  senderFault) are accessible via `e.getSendBatchResult().failed()` but are never included in the message string.
                                                                                                                  
**Sample**
  1. Send a batch via SqsTemplate.sendMany() where at least one message fails at the AWS level (e.g. invalid       
  attribute type, missing FIFO headers)
  2. Catch SendBatchOperationFailedException and call e.getMessage()                                               
  3. Result: "" — no diagnostic information

tomazfernandes:
@krechanski, thanks for bringing this up.

Would you like to contribute a PR fixing it?

krechanski:
Yeah, sure. I can provide a PR 😄

PR #1624, opened 2026-05-23
fix: add descriptive error message to SendBatchOperationFailedException

Baqirrizvidev:
This PR addresses Issue #1605 by constructing a detailed, descriptive exception message when creating a `SendBatchOperationFailedException`.

### Problem
Previously, in `AbstractMessagingTemplate.handleFailedSendBatch`, a `SendBatchOperationFailedException` was created with a hardcoded empty string `""` as the exception message. As a result, `e.getMessage()` returned `""`, rendering the exception silent in logs and requiring developers to manually extract details via `e.getSendBatchResult().failed()`.

### Solution
This PR updates `AbstractMessagingTemplate` to dynamically construct a diagnostic error message when a batch send operation fails. The message now contains:
- The target SQS queue/endpoint.
- The raw SQS message ID (or Spring message ID) of each failed message in the batch.
- The associated failure error message returned from the SQS client.

### Testing
- Added assertions to the existing `shouldThrowIfHasFailedMessagesInBatchByDefault` unit test in `SqsTemplateTests` to verify that `SendBatchOperationFailedException` contains a message with the expected queue name, failed message ID, and SQS error details.
- Verified that all unit tests pass successfully.

### DIFF
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/AbstractMessagingTemplate.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/AbstractMessagingTemplate.java
@@ -401,7 +401,12 @@ public abstract class AbstractMessagingTemplate<S> implements MessagingOperation
 
 	private <T> CompletableFuture<SendResult.Batch<T>> handleFailedSendBatch(String endpoint,
 			SendResult.Batch<T> result) {
-		return CompletableFuture.failedFuture(new SendBatchOperationFailedException("", endpoint, result));
+		String errorMessage = "Send batch operation failed for endpoint %s. Failed messages: %s.".formatted(endpoint,
+				result.failed().stream()
+						.map(failed -> "[Message ID: %s, Error: %s]"
+								.formatted(MessageHeaderUtils.getRawMessageId(failed.message()), failed.errorMessage()))
+						.collect(Collectors.joining(", ")));
+		return CompletableFuture.failedFuture(new SendBatchOperationFailedException(errorMessage, endpoint, result));
 	}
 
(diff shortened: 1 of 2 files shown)


maciejwalkowiak (approved):

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/AbstractMessagingTemplate.java
```diff
@@ -401,7 +401,12 @@ private <T> CompletableFuture<SendResult.Batch<T>> wrapSendException(Collection<
 
 	private <T> CompletableFuture<SendResult.Batch<T>> handleFailedSendBatch(String endpoint,
 			SendResult.Batch<T> result) {
-		return CompletableFuture.failedFuture(new SendBatchOperationFailedException("", endpoint, result));
+		String errorMessage = "Send batch operation failed for endpoint %s. Failed messages: %s.".formatted(endpoint,
+				result.failed().stream()
+						.map(failed -> "[Message ID: %s, Error: %s]"
+								.formatted(MessageHeaderUtils.getRawMessageId(failed.message()), failed.errorMessage()))
```
tomazfernandes, line 407: Let's add a small guard against `null` here in case correlation fails in a previous step and the message is not included in the batch:

E.g.:
```java
failed.message() != null ? MessageHeaderUtils.getRawMessageId(failed.message())                                                                                                                                                                                   
                                                                        : "unknown"
```

tomazfernandes (commented):
@Baqirrizvidev thanks for the PR.

Left a comment for a null guard to be on the safe side. We're changing correlation in a few places. Should be good to merge with that change.

PR #1644, opened 2026-06-30
#1587 Support routing for Message<T> parameters in @SqsHandler

### LINKED ISSUE #1587: Support @SqsHandler routing for Message<T> parameters
**Type**: Bug

**Component**:
"SQS" 

**Describe the bug**
 When using @SqsListener with multiple @SqsHandler methods, if a handler method is Message<'Pojo'> instead of just 'Pojo', routing always falls through to the default handler. 

**Sample**

Using Spring boot 3.x and spring cloud aws 4.0.0

```
// This WORKS — Pojo parameter
 @SqsListener(queueNames = "myQueue")
 public class WorkingListener {

     @SqsHandler
     public void onCreated(MyPojo event) {
         // Correctly routed
     }

     @SqsHandler
     public void onUpdated(MyOtherPojo event) {
         // Correctly routed
     }

     @SqsHandler(isDefault = true)
     public void onDefault(Event event) {
         // Only receives truly unmatched events
     }
 }

 // This FAILS — Message<Pojo> parameter
 @SqsListener(queueNames = "myQueue")
 public class BrokenListener {

     @SqsHandler
     public void onCreated(Message<Pojo> message) {
         // Never reached — falls through to default
     }

     @SqsHandler
     public void onUpdated(Message<MyOtherPojo> message) {
         // Never reached — falls through to default
     }

     @SqsHandler(isDefault = true)
     public void onDefault(Event event) {
         // ALL events end up here
     }
 }

```

**Workaround**

 Add the POJO as the first parameter for routing, and Message as a second parameter:

```
 @SqsHandler
 public void onCreated(Pojo pojo, Message<Pojo> message) {
     // 'pojo' enables correct routing with access to message metadata
 }

``

alexisgra:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
When using `@SqsListener` with multiple `@SqsHandler` methods, routing is unable to route messages to a handler with `Message<T>` envelope as parameter.

The cause is in `CompositeInvocableHandler.isPayloadAssignable`: for a `Message<T>` parameter it compared
the payload against `Message.class`, which never matches a concrete payload. So the handler was never selected. Following Tomas comment in #1587 , the fix detects a `Message` typed parameter and resolves its generic type, then matches the payload against that generic instead.

⚠️ Known breaking change — open for discussion.

Because a `Message<T>` parameter is now considered payload-assignable, a single handler that declares both the Pojo and its Message<Pojo> (the workaround suggested in #1587) now resolves two payload-assignable parameters and throws `"Ambiguous payload parameter for ..."` instead of routing on the Pojo.

Handler (payload = Pojo) | Before | This PR
-- | -- | --
`handle(Pojo)` | matches 🟢 | matches 🟢 
`handle(Message<Pojo>)` | no match 🔴 | matches 🟢 
`handle(Pojo, Message<Pojo>)` | matches 🟢 | throws (breaking change) ⚠️ 

A change in `findCandidate` method can make the last case route on the Pojo again (treat `Message<T>` as the
envelope, not a second payload). I've left it out so you can decide the intended semantics. 
The behavior is pinned by `throwsWhenPayloadAndMessageParametersOnSameMethod`.

Fix #1587

- Added unit tests in `CompositeInvocableHandlerTest` covering : 
   - `Message<T>` routing.
   - Ambiguity across handlers when a plain and a `Message<T>` handler match the same payload type.
   - Ambiguity within one method declaring both a plain and a `Message<T>` parameter of the same type.

- Added integration test in `SqsIntegrationTests`, asserting a `Message<String>` handler is selected and the default handler is not invoked.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
Decide the intended behavior for a handler declaring both a `Pojo` and a `Message<Pojo>` as parameters:
1. Keep it as an ambiguity error (current behavior - breaking change)
2. Route on the Pojo and treat the `Message<T>` as the envelope. 

In my opinion, `handle(Message<Pojo>)` should now replace `handle(Pojo, Message<Pojo>)`; the dual-parameter
form becomes redundant once the envelope routes on its own. However, since the workaround was publicly
documented, I understand you may prefer to avoid a breaking change here.

I'm happy to push an adjustment for 2, if that's preferred.

### DIFF
--- a/docs/src/main/asciidoc/sqs.adoc
+++ b/docs/src/main/asciidoc/sqs.adoc
@@ -720,6 +720,15 @@ public class MyListener {
 
 The `isDefault = true` parameter designates a method as the fallback handler for messages that don't match any other handler's parameter type.
 
+Handler methods can also accept the payload wrapped in a `Message<T>`; routing is based on the generic type, so it works the same way as a plain payload parameter:
+[source, java]
+----
+@SqsHandler
+public void handle(Message<MyPojo> message) {
+	System.out.println(message.getPayload());
+}
+----
+
 To determine which handler method to invoke, the framework needs to know the payload type before deserialization.
 A custom `PayloadTypeMapper` should be configured to map incoming messages to their concrete types.
 See <<Custom Payload Type Mapping>> for an example.
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/adapter/CompositeInvocableHandler.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/adapter/CompositeInvocableHandler.java
@@ -22,6 +22,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import org.jspecify.annotations.Nullable;
 import org.springframework.core.MethodParameter;
+import org.springframework.core.ResolvableType;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.handler.invocation.InvocableHandlerMethod;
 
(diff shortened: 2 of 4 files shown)


tomazfernandes (commented):
Thanks for the PR @alexisgra, looks good.

> I'm happy to push an adjustment for 2, if that's preferred.

Yes, let's keep backwards compatibility. The specific semantics would be a plain payload parameter match should take precedence. Message<T> parameters only participate in routing when no plain parameter matches the payload. That keeps every configuration that routes today routing identically.

We should be good to merge then.

PR #1654, opened 2026-07-21
Fix duplicate DeleteMessageBatchRequestEntry ids when SQS redelivers a message within one ack batch

### LINKED ISSUE #1634: SQS batched acknowledgement can fail when a standard queue redelivers the same message in one ack batch
**Type**: Bug

**Component**:
SQS

**Describe the bug**

When using SQS listener acknowledgement batching with a Standard SQS queue, acknowledgement can fail if the same SQS message is delivered more than once before the acknowledgement batch is flushed.

Spring Cloud AWS version: 4.0.2

Standard SQS queues provide at-least-once delivery, so the same message can legitimately be received more than once. If duplicate deliveries of the same SQS message are included in one acknowledgement batch, Spring Cloud AWS appears to use the SQS message id as the `DeleteMessageBatchRequestEntry.id`.

AWS requires each `DeleteMessageBatchRequestEntry.id` to be unique within a batch request. If two entries use the same id, AWS rejects the request with `BatchEntryIdsNotDistinctException`, causing the acknowledgement batch to fail.

Relevant code path:

- `SqsHeaderMapper.toHeaders(...)` maps the SQS message id into the Spring message header id:  `new MessagingMessageHeaders(messageHeaders, UUID.fromString(source.messageId()))`
- `SqsAcknowledgementExecutor.toDeleteMessageEntry(...)` uses that id as the delete batch entry id: `.id(MessageHeaderUtils.getId(message))`

Actual behavior:

If an acknowledgement batch contains duplicate deliveries of the same SQS message, the generated `DeleteMessageBatchRequest` contains duplicate entry ids and AWS rejects the batch.

Expected behavior:

Spring Cloud AWS should avoid generating duplicate `DeleteMessageBatchRequestEntry.id` values within a single batch request.

The batch entry id does not need to be the SQS message id. It only needs to be unique within the request and usable for correlating batch results. For example, Spring Cloud AWS could generate unique per-entry ids and keep an internal mapping back to the original message/receipt handle for partial failure handling.

Disabling acknowledgement batching works around the issue, but batching should not fail only because SQS redelivered the same message in the same batch window.

**Sample**

I do not have a standalone sample application yet, but the issue should be reproducible by testing the acknowledgement executor directly:

1. Create two Spring `Message<?>` instances representing two deliveries of the same SQS message.
2. Give both messages the same Spring message id, matching the same SQS `messageId`.
3. Give each message a valid `SqsHeaders.SQS_RECEIPT_HANDLE_HEADER`; the receipt handles may be different.
4. Pass both messages to `SqsAcknowledgementExecutor.execute(...)`.
5. Inspect the `DeleteMessageBatchRequest` sent to `SqsAsyncClient.deleteMessageBatch(...)`.

Expected request entries should have unique `DeleteMessageBatchRequestEntry.id` values.

Actual request entries have duplicate ids because both are derived from `MessageHeaderUtils.getId(message)`.

Pseudo-test shape:

```java
@Test
void batchedAcknowledgementShouldUseUniqueDeleteBatchEntryIdsForDuplicateSqsMessageIds() {
    UUID sqsMessageId = UUID.randomUUID();

    Message<String> firstDelivery = messageWithIdAndReceiptHandle(sqsMessageId, "receipt-handle-1");
    Message<String> secondDelivery = messageWithIdAndReceiptHandle(sqsMessageId, "receipt-handle-2");

    SqsAsyncClient sqsAsyncClient = mock(SqsAsyncClient.class);
    ArgumentCaptor<DeleteMessageBatchRequest> requestCaptor =
            ArgumentCaptor.forClass(DeleteMessageBatchRequest.class);

    when(sqsAsyncClient.deleteMessageBatch(requestCaptor.capture()))
            .thenReturn(CompletableFuture.completedFuture(DeleteMessageBatchResponse.builder().build()));

    SqsAcknowledgementExecutor<String> executor = new SqsAcknowledgementExecutor<>();
    executor.setSqsAsyncClient(sqsAsyncClient);
    executor.setQueueAttributes(queueAttributes("test-queue", "https://sqs.example/test-queue"));

    executor.execute(List.of(firstDelivery, secondDelivery)).join();

    List<String> batchEntryIds = requestCaptor.getValue().entries().stream()
            .map(DeleteMessageBatchRequestEntry::id)
            .toList();

    assertThat(batchEntryIds).doesNotHaveDuplicates();
}
```

The exact test setup may differ because Spring message ids are normally generated by the framework. Another way to reproduce this is to use the real `SqsHeaderMapper.toHeaders(...)` conversion path with two AWS SDK SQS `Message` objects that have the same `messageId`.

BK202503:
## Summary
`SqsAcknowledgementExecutor` used `MessageHeaderUtils.getId(message)` — the Spring message id derived from the SQS `messageId` — as the `DeleteMessageBatchRequestEntry.id`. On a Standard queue, if the same SQS message is redelivered within one acknowledgement batch, two entries end up with the same id and AWS rejects the entire batch with `BatchEntryIdsNotDistinctException`. The reporter's analysis and full repro are in #1634.

## Fix
Batch entry id is now a positional index into the snapshot list of messages passed to `deleteMessageBatch`. SQS only requires the id to be unique within a single request and doesn't otherwise interpret it, so a positional index is sufficient and keeps the mapping trivial.

Partial-failure correlation is updated accordingly: `createPartialFailureException` parses each `BatchResultErrorEntry.id` as an index into the ordered batch and buckets messages by index. If any returned id doesn't parse to a valid index — which shouldn't happen but preserves the previous defensive behaviour — the executor still returns the existing "cannot correlate all failure ids" case, treating every message as failed.

## Tests
- **New regression test**: `shouldUseUniqueBatchEntryIdsWhenMessageIdIsDuplicated` — two mock messages sharing the same Spring message id (via `MessagingMessageHeaders` with an explicit shared UUID) but with distinct receipt handles → asserts the captured `DeleteMessageBatchRequest` entries have distinct ids and preserve both receipt handles in order. Fails on `main` (`BatchEntryIdsNotDistinctException` path exercised), passes with this change.
- **Existing partial-failure test updated**: `shouldWrapPartialBatchFailure` now uses `id("0")` (positional) instead of the message-header UUID.
- Full sqs module test suite: **634 tests, 0 failures, 0 errors, 5 skipped** (`./mvnw -pl spring-cloud-aws-sqs -am test` — includes localstack integration tests).

Fixes #1634

### DIFF
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/acknowledgement/SqsAcknowledgementExecutor.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/acknowledgement/SqsAcknowledgementExecutor.java
@@ -25,6 +25,7 @@ import io.awspring.cloud.sqs.listener.SqsHeaders;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
@@ -98,14 +99,15 @@ public class SqsAcknowledgementExecutor<T>
 				MessageHeaderUtils.getId(messagesToAck));
 		StopWatch watch = new StopWatch();
 		watch.start();
+		List<Message<T>> orderedMessages = new ArrayList<>(messagesToAck);
 		return CompletableFutures.exceptionallyCompose(this.sqsAsyncClient
-			.deleteMessageBatch(createDeleteMessageBatchRequest(messagesToAck)).thenCompose(
-					response -> handleDeleteMessageBatchResponse(messagesToAck, response)),
-			t -> toAcknowledgementFailure(messagesToAck, t))
-			.whenComplete((v, t) -> logAckResult(messagesToAck, t, watch));
+			.deleteMessageBatch(createDeleteMessageBatchRequest(orderedMessages)).thenCompose(
+					response -> handleDeleteMessageBatchResponse(orderedMessages, response)),
+			t -> toAcknowledgementFailure(orderedMessages, t))
+			.whenComplete((v, t) -> logAckResult(orderedMessages, t, watch));
 	}
 
-	private CompletableFuture<Void> handleDeleteMessageBatchResponse(Collection<Message<T>> messagesToAck,
+	private CompletableFuture<Void> handleDeleteMessageBatchResponse(List<Message<T>> messagesToAck,
 			DeleteMessageBatchResponse response) {
 		if (!response.failed().isEmpty()) {
 			return CompletableFutures.<Void>failedFuture(createPartialFailureException(messagesToAck, response));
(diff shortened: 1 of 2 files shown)


tomazfernandes (approved):
Thanks for the PR @BK202503, looking forward to more.

PR #1656, opened 2026-07-22
feat(s3): add copy method to S3Template

### LINKED ISSUE #1157: Add copy to S3Template
**Type**: Feature

When uploading a gzip file via another API it doesn't create the "Content-Encoding" metadata automatically, so I have to copy the file adding the contentEncoding parameter using S3Client.

Add a method to execute Copy on S3Template passing the source/destination buckets and keys and possibly the Content-Encoding parameter.

Alternative is using S3Client directly, but more verbose.

May I try to work on this?

abilng:

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Allow copying objects between buckets/keys via S3Operations without dropping down to S3Client.

Closes #1157

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/ObjectMetadata.java
+++ b/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/ObjectMetadata.java
@@ -260,6 +260,95 @@ public class ObjectMetadata {
 		}
 	}
 
+	void apply(CopyObjectRequest.Builder builder) {
+		builder.metadataDirective(MetadataDirective.REPLACE);
+		if (acl != null) {
+			builder.acl(acl);
+		}
+		if (cacheControl != null) {
+			builder.cacheControl(cacheControl);
+		}
+		if (contentDisposition != null) {
+			builder.contentDisposition(contentDisposition);
+		}
+		if (contentEncoding != null) {
+			builder.contentEncoding(contentEncoding);
+		}
+		if (contentLanguage != null) {
+			builder.contentLanguage(contentLanguage);
+		}
+		if (contentType != null) {
+			builder.contentType(contentType);
+		}
+		if (expires != null) {
+			builder.expires(expires);
+		}
+		if (grantFullControl != null) {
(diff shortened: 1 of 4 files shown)


spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/ObjectMetadata.java
```diff
@@ -260,6 +260,95 @@ void apply(PutObjectRequest.Builder builder) {
 		}
 	}
 
+	void apply(CopyObjectRequest.Builder builder) {
+		builder.metadataDirective(MetadataDirective.REPLACE);
```
MatejNedic: Should be configurable as well. Someone might want to override only if no value is set so .COPY

spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/ObjectMetadata.java
```diff
@@ -260,6 +260,95 @@ void apply(PutObjectRequest.Builder builder) {
...
+		if (acl != null) {
```
MatejNedic, line 269: User ObjectProvider and set to builder instead of many ifs

MatejNedic (changes requested):
Hey @abilng ,

Thanks on PR, few changes needed.

spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/ObjectMetadata.java
maciejwalkowiak: Many `ifs` is consistent with other parts of this class. I think we can keep it in this PR and refactor independently.

spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/ObjectMetadata.java
abilng: fixed

Issue #1042, opened 2024-02-09
Enable SqsTemplate to send more than 10 messages in a batch


tomazfernandes:
As discussed [here](https://github.com/awspring/spring-cloud-aws/issues/879#issuecomment-1933580325), currently `SqsTemplate` fails if more than 10 messages are provided in a batch with an error from AWS SDK.

We can add support for this use case by partitioning the message list in batches of 10, while also sending the batches in parallel.

This change would likely be around here:

https://github.com/awspring/spring-cloud-aws/blob/562b9334821ea99d7eeac886e60dd59ef73451b3/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java#L350-L356

levys17:
@tomazfernandes  I started to work on it...If it's ok, can you assign this issue to me?

techadri:
@tomazfernandes I'd be happy to provide a PR for this topic, in case you (@levys17) aren't working on it anymore.

levys17:
> @tomazfernandes I'd be happy to provide a PR for this topic, in case you (@levys17) aren't working on it anymore.

I didn't have time to finish that and I've barely worked on it in last weeks....so for me it'll be nice if you do that :D

techadri:
@tomazfernandes 
Currently, `doSendBatchAsync` sends all messages in a single batch request using the `sqsAsyncClient`, ensuring that FIFO queues maintain the order of messages. If we change to partitioning messages and sending batches in parallel, only messages within each batch will retain order, but the order of the batches may vary. What are your thoughts on this?

tomazfernandes:
Oh, that's a great point.

I think the ideal solution would be partitioning batches per message group for fifo queues and then sending each batch in parallel.

A simpler solution would be sending batches in parallel for Standard queues and sequentially for FIFO queues.

WDYT?

joseiedo:
Hello. I couldn't find existing PRs and the error was still happening. Opened a PR #1659 following the approach suggested here, I'm happy to receive feedbacks!

PR #1659, opened 2026-07-24
Add support for SqsTemplate to send more than 10 messages at once

joseiedo:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
AWS SDK doesn't have support to send a batch of messages of more than 10 items. This PR add an alternative where the received batch is partitioned in smaller batches of 10. For FIFO queues the batches are sent sequentially (if one fails, the others aren't sent), while for non-FIFO they are parallelized.

solves #1042

I added integration test and unit tests with a sample.

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
Check if there's concerns with this approach and update documentation explaining how this works in order to make it clear for users.


spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
```diff
@@ -365,13 +369,98 @@ private SendMessageRequest doCreateSendMessageRequest(Message message, QueueAttr
...
+	private <T> CompletableFuture<SendResult.Batch<T>> sendStandardBatches(String endpointName,
+			Collection<Message> messages, Map<String, org.springframework.messaging.Message<T>> originalMessagesById) {
+		List<CompletableFuture<SendResult.Batch<T>>> futures = CollectionUtils.partition(messages, 10).stream()
+				.map(partition -> sendSingleBatch(endpointName, partition, originalMessagesById)).toList();
+		return combineBatchFutures(futures);
+	}
+
+	private <T> CompletableFuture<SendResult.Batch<T>> sendFifoBatches(String endpointName,
+			Collection<Message> messages, Map<String, org.springframework.messaging.Message<T>> originalMessagesById) {
+		Map<String, List<Message>> groupedByMessageGroup = messages.stream().collect(Collectors.groupingBy(msg -> {
```
tomazfernandes, line 443: I think we could improve on this partitioning so it becomes more efficient with many small groups. For instance, 4 groups with 3 messages each would translate into 4 requests when it could become two parallel requests. Also, `SqsTemplate` assigns a random UUID as `MessageGroupId` by default, so a batch of 50 messages would translate into 50 single-entry requests.

I suggest packing the whole message groups into shared requests on a best effort basis e.g.:
- a message group with 4 messages would combine with a message group with 6 messages in a single request with 10 entries
- two message groups with 6 messages each creates two parallel requests.
- 50 messages with random message groups generate 5 batches of 10 messages

Since each small group stays whole within a single request, ordering is preserved and the packed requests can all go out in parallel, and groups larger than 10 keep the sequential chain + skip-on-failure behavior you already implemented.

What do you think?

tomazfernandes (commented):
Thanks for the PR @joseiedo. I see you're still pushing commits, so I'll leave just one comment about the partitioning strategy. Let me know your thoughts.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
joseiedo: Hi @tomazfernandes, this totally makes sense.
I like this strategy, I don't see a problem since the  combinations will only happen if their size sum is not > 10. 

I will work on that, thanks for the review.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/CollectionUtils.java
```diff
@@ -26,7 +26,7 @@ public static <T> List<Collection<T>> partition(Collection<T> messagesToAck, int
 		List<T> messagesToUse = getAsList(messagesToAck);
 		int totalSize = messagesToUse.size();
 		return IntStream.rangeClosed(0, (totalSize - 1) / pageSize)
-				.mapToObj(index -> messagesToUse.subList(index * pageSize, Math.min((index + 1) * pageSize, totalSize)))
+				.mapToObj(index -> (Collection<T>) messagesToUse.subList(index * pageSize, Math.min((index + 1) * pageSize, totalSize)))
```
joseiedo: I'm not sure why suddenly this was happening, but the pipeline couldn't compile successfully if I didn't used this cast. 

https://github.com/awspring/spring-cloud-aws/actions/runs/30174435089/job/89720833992

joseiedo:
Hi @tomazfernandes 

I added a binpack algorithm here, let me know your thoughts about it!

tomazfernandes (commented):
Thanks for the update @joseiedo, the partitioning logic looks great.

One thing left to adjust is that currently if one sending operation in a batch fails, we'll throw the error and some messages will have been sent and others won't.

For the single batch (<= 10 messages) this works because it's all-or-nothing anyway, but for multiple batches we need a way of letting the user know what was sent and what was not.

My suggestion would be to convert exceptions to `Failed` entries at the partition boundary, so every future entering `combineBatchFutures` and the sequential chains always completes normally. Something along the lines of:

```java
private <T> CompletableFuture<SendResult.Batch<T>> sendPartitionedBatch(String endpointName, Collection<Message> partition, Map<String, org.springframework.messaging.Message<T>> originalMessagesById) {
    return sendSingleBatch(endpointName, partition, originalMessagesById).exceptionally(t -> createFailedBatchResult(partition, t, endpointName, originalMessagesById));
}
```

Where `createFailedBatchResult` would mirror `createSkippedResult`, with the unwrapped cause (strip `CompletionException`) in `errorMessage` and `additionalInformation`. With the default `THROW` strategy the user still gets `SendBatchOperationFailedException`, just carrying the full result now.

What do you think?

joseiedo:
Hi @tomazfernandes, I did the adjustment. For the `additionalInformation` field I wasn't sure what to put, so I added the class name of cause. Let me know if we can add more details or if there is a better way to do anything 🤓

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
```diff
@@ -365,13 +371,159 @@ private SendMessageRequest doCreateSendMessageRequest(Message message, QueueAttr
...
+	}
+
+	private <T> SendResult.Batch<T> createFailedBatchResult(Collection<Message> partition, Throwable throwable,
+			String endpointName, Map<String, org.springframework.messaging.Message<T>> originalMessagesById) {
+		Throwable cause = throwable;
+		if (cause instanceof java.util.concurrent.CompletionException) {
+			cause = cause.getCause();
+		}
+		String errorMessage = cause != null && cause.getMessage() != null ? cause.getMessage() : "Unknown error";
+		Map<String, Object> additionalInformation = Map.of(SqsTemplateParameters.ERROR_CODE_PARAMETER_NAME,
```
tomazfernandes: Since we don't have an actual error code from AWS in this path, let's please create a new `SqsTemplateParameter` such as `EXCEPTION_PARAMETER_NAME` and hold the cause instead of reusing `ERROR_CODE_PARAMETER_NAME`.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
```diff
@@ -365,13 +371,159 @@ private SendMessageRequest doCreateSendMessageRequest(Message message, QueueAttr
...
+	}
+
+	/**
+	 * Bin-pack small FIFO groups into shared batches using first-fit decreasing algorithm. Each group is kept whole
+	 * within a single batch. Groups are sorted by size descending before packing to minimize the number of batches.
+	 * @param smallGroups groups with size <= maxBatchSize
+	 * @param maxBatchSize the maximum number of messages per batch (SQS limit is 10)
+	 * @return packed batches, each containing one or more whole groups
+	 */
+	protected static List<List<Message>> binPackSmallFifoGroups(List<List<Message>> smallGroups, int maxBatchSize) {
```
tomazfernandes: Why are we making this method protected rather than private?

spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/integration/SqsIntegrationTests.java
```diff
@@ -342,6 +342,15 @@ void receivesMessageBatch() throws Exception {
 		assertThat(latchContainer.acknowledgementCallbackBatchLatch.await(10, TimeUnit.SECONDS)).isTrue();
 	}
 
+	@Test
+	void shouldSendMoreThan10MessagesAtOnce() {
+		List<Message<String>> messages = IntStream.range(0, 25)
+				.mapToObj(i -> MessageBuilder.withPayload("moreThan10-payload-" + i).build()).toList();
+		SendResult.Batch<String> result = sqsTemplate.sendMany(RECEIVES_MESSAGE_QUEUE_NAME, messages);
```
tomazfernandes: Let's create a separate queue for this test please.

tomazfernandes (commented):
Thanks for the updates @joseiedo, looks good.

I've left a few minor final comments.

Please also update the docs for this new feature (since 4.2.0).

If possible, add a test with the automatic group id scenario: more than 10 messages without group id and asserting that it was batched correctly, e.g. 25 messages should result in 3 requests.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
joseiedo: I did to be able to write tests. However, now that I'm thinking it's redundant to test this method isolated when we're having tests for the complete flow. I will change to private!

joseiedo:
Hi @tomazfernandes, changes done!

For the documentation I replaced only the part where it talks about the limitation, let me know if you think it's necessary to have a section with a more detailed explanation or if it's ok.
I also added the test case you mentioned.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/operations/SqsTemplate.java
```diff
@@ -365,13 +371,158 @@ private SendMessageRequest doCreateSendMessageRequest(Message message, QueueAttr
...
+	}
+
+	private <T> SendResult.Batch<T> createFailedBatchResult(Collection<Message> partition, Throwable throwable,
+			String endpointName, Map<String, org.springframework.messaging.Message<T>> originalMessagesById) {
+		Throwable cause = throwable;
+		if (cause instanceof java.util.concurrent.CompletionException) {
+			cause = cause.getCause();
+		}
+		String errorMessage = cause != null && cause.getMessage() != null ? cause.getMessage() : "Unknown error";
+		Map<String, Object> additionalInformation = Map.of(SqsTemplateParameters.EXCEPTION_PARAMETER_NAME, cause);
```
tomazfernandes: If `cause` is null, `Map.of` will throw here. Let's fall back to the original exception if that's the case.

docs/src/main/asciidoc/sqs.adoc
```diff
@@ -319,8 +319,10 @@ SendResult.Batch<T> sendMany(String queue, Collection<Message<T>> messages);
...
-This limitation is documented in the https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessageBatch.html[AWS SQS API Reference for SendMessageBatch]. As of now, Spring Cloud AWS does not automatically split larger collections into smaller batches of 10 or fewer messages. Users are responsible for ensuring the batch size complies with this AWS limit.
+To send a collection of objects in a single message, the collection must be wrapped in an object.
+The underlying AWS SQS API has a limitation that only up to 10 messages can be sent in a single batch request using `SendMessageBatch`.
+Since 4.2.0, Spring Cloud AWS automatically splits larger collections into batches of 10 or fewer messages.
+For standard queues, all batches are sent in parallel. For FIFO queues, batches are sent sequentially to preserve the message order, with messages grouped by message group ID.
```
tomazfernandes: For FIFO we send the batches in parallel, grouped by message group id.

> to have a section with a more detailed explanation or if it's ok.

I think it's worth adding a bit more detail in a separate section if you're up for it. We don't need to explain the whole mechanism in detail, but an overview covering: 
- No behavior changes for batches < 10;
- How messages are partitioned into batches of up to 10 at a high level;
- Standard (parallel) versus FIFO (grouped by message group ID, ordering preserved per group);

And the failure modes:
- Partial failures;
- Skipped batches after a failure within a FIFO group;
- Exceptions surfacing as failed entries;
- How this interacts with SendBatchFailureHandlingStrategy;

Let me know if that works for you.

tomazfernandes (changes requested):
@joseiedo left a couple of comments, let me know what you think. We should be good to merge after this.

docs/src/main/asciidoc/sqs.adoc
joseiedo: Hi @tomazfernandes, I added a section for this. If it's too much or too little or the wording could be better, let me know!

tomazfernandes (approved):
Thanks for the PR and updates @joseiedo, looking forward to more!

PR #1663, opened 2026-07-31
Flush buffered acknowledgements from the shutdown wait

### LINKED ISSUE #1661: BatchingAcknowledgementProcessor: buffered acknowledgements are never flushed on shutdown (remaining half of #925)
**Type**: Bug

**Component**:
SQS

**Describe the bug**

**TL;DR** — When a container is stopped while `acksBuffer` holds `1 .. ackThreshold - 1` messages, those acknowledgements are never sent: shutdown blocks for the full `acknowledgementShutdownTimeout` (20s by default), the buffer is then cleared, and the messages are redelivered after the visibility timeout even though the listener completed them successfully. `executeAllAcks()` is reachable only from the scheduled flusher, which is already dead by then, so no unconditional final flush exists — the wait added by #1082 has nothing left to wait for. Workaround: `acknowledgementInterval(Duration.ZERO)`.

Versions:

- `spring-cloud-aws` **4.1.0** (also present in current `main`, `a1235192a54a9a19c0d19806e602509781a53bc0`)
- Spring Boot 4.1.0, JDK 25
- Standard (non-FIFO) queue, `AcknowledgementMode.ON_SUCCESS`, `acknowledgementInterval` / `acknowledgementThreshold` left at the Standard-SQS defaults (1s / 10)

This is the remaining half of #925. PR #1082 made `doStop()` **wait** for the acknowledgement buffer to drain, and the *threshold* path can still drain full batches during that window — but nothing is left that can drain a **sub-threshold remainder**. Under sustained load the wait therefore reliably times out and those buffered acknowledgements are discarded instead of being sent.

`AbstractOrderingAcknowledgementProcessor.stop()` sets `running = false` *before* calling `doStop()`. From that moment on:

1. The scheduled (interval) flusher refuses to re-arm itself, so only the already-pending execution can still fire — after that the chain is dead:

   https://github.com/awspring/spring-cloud-aws/blob/a1235192a54a9a19c0d19806e602509781a53bc0/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/acknowledgement/BatchingAcknowledgementProcessor.java#L467-L471

2. That single remaining execution only flushes if it passes this guard:

   https://github.com/awspring/spring-cloud-aws/blob/a1235192a54a9a19c0d19806e602509781a53bc0/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/acknowledgement/BatchingAcknowledgementProcessor.java#L501-L507

   Whether that last run flushes is a race rather than an unconditional failure. It is deterministically lost whenever a *threshold* flush bumped `lastAcknowledgement` (via `doExecute`) **after** the pending scheduled task had already been armed: the task still fires at its original time, which is now earlier than `lastAcknowledgement + ackInterval`, so the guard fails, nothing is flushed, and — per (1) — nothing is rescheduled. Under sustained load that is the normal state of affairs, and it also keeps happening *during* the shutdown window, because the buffering thread deliberately stays alive and keeps moving messages from `acks` into the buffer (`shouldKeepPollingAcks()`), running the threshold executor on each one. At low throughput, by contrast, the scheduled path itself drives `lastAcknowledgement`, so the final run passes the guard and flushes normally.

3. The only remaining drain path is the threshold executor, which requires `>= ackThreshold` messages in a group:

   https://github.com/awspring/spring-cloud-aws/blob/a1235192a54a9a19c0d19806e602509781a53bc0/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/acknowledgement/BatchingAcknowledgementProcessor.java#L432-L438

   Polling has already stopped, so no new messages arrive and the buffer can never reach the threshold again.

4. `waitOnAcknowledgementsIfTimeoutSet()` therefore spins for the full timeout, warns, and `acksBuffer` is cleared:

   https://github.com/awspring/spring-cloud-aws/blob/a1235192a54a9a19c0d19806e602509781a53bc0/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/acknowledgement/BatchingAcknowledgementProcessor.java#L241-L268

**Net effect:** whenever the buffer holds `1 .. ackThreshold - 1` messages at shutdown, those messages are never deleted from SQS, shutdown is delayed by the full `acknowledgementShutdownTimeout` (20s by default), and the messages are redelivered once the visibility timeout expires — even though the listener completed them successfully.

Observed in production (Kubernetes, standard queue, `maxConcurrentMessages` 10, `maxMessagesPerPoll` 10):

```
WARN io.awspring.cloud.sqs.listener.acknowledgement.BatchingAcknowledgementProcessor
     Acknowledgements did not finish in 20000 ms. Proceeding with shutdown.
```

<details>
<summary>CloudWatch evidence for the stranded messages (click to expand)</summary>

- 2 occurrences in one evening, both coinciding with pod scale-in
- `NumberOfMessagesReceived - NumberOfMessagesDeleted` opened a small positive gap on that day, consistent in magnitude with 2 events × `1 .. ackThreshold - 1` stranded messages. The two adjacent days stayed inside that metric's normal noise band
- `ApproximateAgeOfOldestMessage` spiked to exactly our visibility timeout during the matching hour, and was flat otherwise

</details>

The load dependence matches the mechanism above. This service never logged the warning while per-pod throughput was low: at that rate the *scheduled* path (not the threshold path) drives `lastAcknowledgement`, so the final scheduled run passes the guard and flushes normally. After we reduced the replica count, per-pod throughput roughly tripled and the warning started appearing.

This is particularly relevant for `maxReceiveCount = 2` setups (as reported in #925): a message stranded on two consecutive deliveries goes to the DLQ despite having been processed successfully both times.

**Sample**

I don't have a standalone reproducer application, but the condition is deterministic and should be expressible as a unit test alongside the existing `BatchingAcknowledgementProcessorTests`. Note that the shutdown tests added by #1082 (`testAckOnShutdown`) don't cover it: they use `acknowledgementInterval = Duration.ZERO`, so the scheduled flusher never runs at all, and they acknowledge exactly 100 messages — an exact multiple of the threshold, leaving no remainder.

1. Configure `acknowledgementInterval = 1s`, `acknowledgementThreshold = 10`, `acknowledgementShutdownTimeout = 20s`.
2. Acknowledge exactly 10 messages, so the threshold executor flushes and sets `lastAcknowledgement = now`.
3. Immediately acknowledge N more messages with `1 <= N <= 9`, and let them reach `acksBuffer`.
4. Call `stop()` within `ackInterval` of step 2.

Expected: the N messages are acknowledged — there is 20s of budget available.
Actual: `hasAcksLeft()` stays `true`, the wait times out after 20s, and the N messages are dropped by `acksBuffer.clear()`.

**Possible fix direction:** let the scheduled execution keep flushing (or perform one final unconditional `executeAllAcks()`) while `waitOnAcknowledgementsIfTimeoutSet()` is still within its budget — `doStop()` only disposes the `taskScheduler` *after* that wait returns, so the scheduler is still usable for exactly this window. The buffering thread is already deliberately kept alive for the same window via `shouldKeepPollingAcks()` — arguably the flusher should be too, otherwise the wait added by #1082 has nothing to wait for.

**Workaround** for anyone hitting this: set `acknowledgementInterval(Duration.ZERO)`, which selects `ImmediateAcknowledgementProcessor`. Thanks to the `null` threshold check at [`StandardSqsComponentFactory#L70`](https://github.com/awspring/spring-cloud-aws/blob/a1235192a54a9a19c0d19806e602509781a53bc0/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/StandardSqsComponentFactory.java#L67-L72), setting the threshold as well is not necessary.

hyeongguen-song:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Fixes #1661.

Messages left in the acknowledgement buffer below `acknowledgementThreshold` are
never flushed on shutdown. The processor spins for the whole
`acknowledgementShutdownTimeout`, logs

```
Acknowledgements did not finish in 20000 ms. Proceeding with shutdown.
```

and then clears the buffer, so those messages are never deleted from SQS and are
redelivered once the visibility timeout expires.

The buffer is now drained from the shutdown wait loop:

```java
private void flushRemainingAcks() {
    if (!this.acks.isEmpty()) {
        return;
    }
    this.context.lock();
    try {
        this.context.executeAllAcks();
    }
    finally {
        this.context.unlock();
    }
}
```

`running` in `AbstractOrderingAcknowledgementProcessor` is also made `volatile`.

There are two independent ways to lose the remainder, and draining from the
shutdown wait covers both:

**With a scheduled execution.** `AbstractOrderingAcknowledgementProcessor.stop()`
sets `running = false` before calling `doStop()`, so `scheduleNextExecution()`
refuses to arm the next execution. The polling thread keeps running until the
shutdown timeout elapses (`shouldKeepPollingAcks()`), so it goes on moving
messages into the buffer, and each threshold execution pushes
`lastAcknowledgement` forward. The one already-armed execution then fires with
`lastAcknowledgement` newer than the time it was armed, skips the flush because
`now` is not after `lastAcknowledgement + ackInterval`, and cannot re-arm.

**With no scheduled execution at all.** `ScheduledAcknowledgementExecution.start()`
only arms anything when `ackInterval != Duration.ZERO`. With
`acknowledgementInterval = ZERO` and a positive threshold — a configuration
`doStart()` explicitly allows — the threshold path is the only flush mechanism,
so a sub-threshold remainder is lost on *every* shutdown.

Making the drain part of the shutdown path itself covers both, instead of
depending on a scheduled execution existing and being armed.

Two details in the implementation:

- The flush waits for the ack queue to drain, so batches are not needlessly split.
- It is retried on every iteration of the wait loop rather than executed once. The
  polling thread may be holding a message it has already polled but not yet added
  to the buffer, in which case the queue looks empty while the message is in
  neither the queue nor the buffer.

`running` was written under `lifecycleMonitor` but read unsynchronized from the
polling and scheduler threads via `isRunning()`, so neither thread had any
guarantee of observing `stop()`.

This is load-dependent in the scheduled case, which is why it is easy to miss. It
needs the threshold path to be the dominant flush path, i.e. messages arriving
faster than one `acknowledgementInterval` per batch. We hit it in production after
a scaling change raised per-pod throughput several-fold: pod terminations started
leaving a small number of messages undeleted, visible as a gap between
`NumberOfMessagesReceived` and `NumberOfMessagesDeleted` and as a spike in
`ApproximateAgeOfOldestMessage` matching the visibility timeout.

Two new tests, one per loss path. Both are behaviour-only and do not reach into
internals:

- `givenScheduledAcknowledgement_whenStoppedWithRemainderBelowThreshold_...` acks
  exactly `acknowledgementThreshold` messages, waits on a latch in the
  acknowledgement executor so the threshold flush is known to have completed, then
  acks a sub-threshold remainder and stops. The interval outlasts the shutdown
  timeout, so a scheduled execution exists but provably cannot fire during the
  test.
- `givenZeroIntervalAndThreshold_whenStoppedWithRemainderBelowThreshold_...` uses
  `acknowledgementInterval = ZERO`, where no scheduled execution exists.

Both fail on `main` and pass with this change. Measured on JDK 17.0.19, 15
consecutive runs each:

| | result |
| --- | --- |
| with this change | 15/15 pass |
| `main` + `volatile` only | 15/15 reproduce the bug |
| `main` unchanged | 15/15 reproduce the bug |

On failure each test spins the full 10 s shutdown timeout and loses exactly the
5-message remainder.

The three existing shutdown tests could not catch either path: they all use
`acknowledgementInterval(Duration.ZERO)` so no scheduled execution is ever
created, and they use exactly 100 messages against a threshold of 10, so no
remainder is ever left in the buffer.

Full module: `./mvnw -pl spring-cloud-aws-sqs -am test` gives **636 tests, 0
failures, 0 errors, 6 skipped**, spotless checks enabled, including the LocalStack
integration tests.

Rebased onto `main` after #1662 landed, and re-verified with no local patches
applied.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

No documentation change needed: this restores the documented behaviour rather
than changing it.

## :crystal_ball: Next steps

Users who cannot wait for a release can switch to an
`ImmediateAcknowledgementProcessor`, which has no buffer and no scheduler, by
setting `acknowledgementInterval(Duration.ZERO)` and leaving
`acknowledgementThreshold` unset — `StandardSqsComponentFactory` selects it when
the interval is `ZERO` and the threshold is `null` or `0`. Anyone who has
explicitly configured a positive threshold needs to set it back to `0` as well,
otherwise they still get the batching processor and the second path above.

### DIFF
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/acknowledgement/AbstractOrderingAcknowledgementProcessor.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/acknowledgement/AbstractOrderingAcknowledgementProcessor.java
@@ -64,7 +64,8 @@ public abstract class AbstractOrderingAcknowledgementProcessor<T>
 	private AsyncAcknowledgementResultCallback<T> acknowledgementResultCallback = new AsyncAcknowledgementResultCallback<T>() {
 	};
 
-	private boolean running;
+	// Written under lifecycleMonitor, but read without synchronization from the polling and scheduler threads
+	private volatile boolean running;
 
 	private String id;
 
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/acknowledgement/BatchingAcknowledgementProcessor.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/acknowledgement/BatchingAcknowledgementProcessor.java
@@ -250,6 +250,7 @@ public class BatchingAcknowledgementProcessor<T> extends AbstractOrderingAcknowl
 					if (LocalDateTime.now().isAfter(endTime)) {
 						throw new TimeoutException();
 					}
+					flushRemainingAcks();
 					Thread.sleep(200);
 				}
 				logger.debug("All acknowledgements completed.");
@@ -266,6 +267,28 @@ public class BatchingAcknowledgementProcessor<T> extends AbstractOrderingAcknowl
 				logger.warn(
 						"Error thrown when waiting for acknowledgement tasks to finish in {}. Continuing with shutdown.",
 						this.parent.getId(), e);
+			}
+		}
+
+		/**
+		 * Flushes whatever is left in the buffer, so that messages below the acknowledgement threshold are acknowledged
+		 * on shutdown instead of being discarded when the buffer is cleared. Waits for the ack queue to drain first so
+		 * that batches are not needlessly split.
+		 *
+		 * This has to be retried on every iteration of the shutdown wait loop rather than executed once: the polling
+		 * thread may be holding a message it has already polled but not yet added to the buffer, in which case the
+		 * queue looks empty while the message is in neither the queue nor the buffer.
+		 */
+		private void flushRemainingAcks() {
+			if (!this.acks.isEmpty()) {
+				return;
+			}
+			this.context.lock();
+			try {
+				this.context.executeAllAcks();
+			}
+			finally {
+				this.context.unlock();
 			}
 		}
 
--- a/spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/listener/acknowledgement/BatchingAcknowledgementProcessorTests.java
+++ b/spring-cloud-aws-sqs/src/test/java/io/awspring/cloud/sqs/listener/acknowledgement/BatchingAcknowledgementProcessorTests.java
@@ -68,6 +68,8 @@ class BatchingAcknowledgementProcessorTests {
 
 	private static final Duration ACK_INTERVAL_ZERO = Duration.ZERO;
 
+	private static final Duration ACK_INTERVAL_THIRTY_SECONDS = Duration.ofSeconds(30);
+
 	private static final int MAX_ACKNOWLEDGEMENTS_PER_BATCH_TEN = 10;
 
 	private static final Integer ACK_THRESHOLD_TEN = 10;
@@ -187,6 +189,71 @@ class BatchingAcknowledgementProcessorTests {
 		processor.stop();
 		logger.debug("Processor stopped, waiting on latch");
 		assertThat(ackLatch.await(1, TimeUnit.SECONDS)).isEqualTo(shouldWaitAllAcks);
+	}
+
+	@Test
+	void givenScheduledAcknowledgement_whenStoppedWithRemainderBelowThreshold_shouldAcknowledgeRemainingMessages()
+			throws Exception {
+		List<Message<String>> messages = buildMessages(ACK_THRESHOLD_TEN + 5);
+		List<Message<String>> thresholdBatch = messages.subList(0, ACK_THRESHOLD_TEN);
+		List<Message<String>> remainder = messages.subList(ACK_THRESHOLD_TEN, messages.size());
+		Collection<Message<String>> acknowledgedMessages = Collections.synchronizedList(new ArrayList<>());
+		CountDownLatch thresholdFlushLatch = new CountDownLatch(1);
+
+		// The interval outlasts the test, so the remainder can only be flushed by the shutdown drain and never by a
+		// scheduled execution. A scheduled execution does exist here, unlike in the interval ZERO scenario below.
+		BatchingAcknowledgementProcessor<String> processor = createProcessor(ACK_INTERVAL_THIRTY_SECONDS,
+				messagesToAck -> {
+					acknowledgedMessages.addAll(messagesToAck);
+					thresholdFlushLatch.countDown();
+					return CompletableFuture.completedFuture(null);
+				});
+		processor.start();
+
+		processor.doOnAcknowledge(thresholdBatch);
+		assertThat(thresholdFlushLatch.await(10, TimeUnit.SECONDS)).isTrue();
+		processor.doOnAcknowledge(remainder);
+		processor.stop();
+
+		assertThat(acknowledgedMessages).containsExactlyInAnyOrderElementsOf(messages);
+	}
+
+	@Test
+	void givenZeroIntervalAndThreshold_whenStoppedWithRemainderBelowThreshold_shouldAcknowledgeRemainingMessages()
+			throws Exception {
+		List<Message<String>> messages = buildMessages(ACK_THRESHOLD_TEN + 5);
+		Collection<Message<String>> acknowledgedMessages = Collections.synchronizedList(new ArrayList<>());
+
+		// No scheduled execution is ever created with interval ZERO, so the shutdown drain is the only thing that can
+		// flush the remainder.
+		BatchingAcknowledgementProcessor<String> processor = createProcessor(ACK_INTERVAL_ZERO, messagesToAck -> {
+			acknowledgedMessages.addAll(messagesToAck);
+			return CompletableFuture.completedFuture(null);
+		});
+		processor.start();
+
+		processor.doOnAcknowledge(messages);
+		processor.stop();
+
+		assertThat(acknowledgedMessages).containsExactlyInAnyOrderElementsOf(messages);
+	}
+
+	private static List<Message<String>> buildMessages(int count) {
+		return IntStream.range(0, count).mapToObj(index -> MessageBuilder.withPayload(String.valueOf(index)).build())
+				.collect(Collectors.toList());
+	}
+
+	private static BatchingAcknowledgementProcessor<String> createProcessor(Duration acknowledgementInterval,
+			AcknowledgementExecutor<String> acknowledgementExecutor) {
+		BatchingAcknowledgementProcessor<String> processor = new BatchingAcknowledgementProcessor<>();
+		processor.configure(SqsContainerOptions.builder().acknowledgementInterval(acknowledgementInterval)
+				.acknowledgementThreshold(ACK_THRESHOLD_TEN).acknowledgementOrdering(AcknowledgementOrdering.PARALLEL)
+				.acknowledgementShutdownTimeout(Duration.ofSeconds(10)).build());
+		processor.setTaskExecutor(new SimpleAsyncTaskExecutor());
+		processor.setAcknowledgementExecutor(acknowledgementExecutor);
+		processor.setMaxAcknowledgementsPerBatch(MAX_ACKNOWLEDGEMENTS_PER_BATCH_TEN);
+		processor.setId(ID);
+		return processor;
 	}
 
 	@Test


tomazfernandes:
Thanks @hyeongguen-song, great analysis and writeup. I have three asks so we can close this class of bug for good if you're up for it.

1. First is for us to flush from the shutdown wait instead of keeping the scheduler armed. We'd drop the `shouldKeepSchedulingAcks()` change, and instead, in `waitOnAcknowledgementsIfTimeoutSet()`'s loop: `if acks.isEmpty()`, take the lock and `executeAllAcks()` to flush any remaining acks in the buffer,

This would make the drain a guarantee of the shutdown path itself, and also covers `interval = ZERO + threshold > 0`, where no scheduled execution ever exists and a remainder is lost on every shutdown.

This would need to be inside the loop: the polling thread can hold a just-polled message when the queue looks empty.

What do you think?

2. Let's make `running` volatile in `AbstractOrderingAcknowledgementProcessor`. It's read unsynchronized from the polling and scheduler threads. The current test you added is nondeterministic: with the fix reverted it fails only ~1 in 3 runs here because of stale true reads.

3. Let's add these test scenarios:
- Test A: sub-threshold remainder with a live timer config:
1. Ack exactly threshold (10) messages.
2. Await that threshold flush completing (latch in the acknowledgement executor).
3. Ack a sub-threshold remainder (e.g. 5).
4. stop().
5. Assert all 15 messages were acked

- B: interval = ZERO, threshold = 10, ack 15, stop(), assert all 15 acked.

Let me know your thoughts.

hyeongguen-song:
@tomazfernandes

All three make sense to me — pushed. Details below.

**1. Flush from the shutdown wait.** Agreed, and your `interval = ZERO` point is
the decisive one: my approach did nothing for that configuration. I wrote your
Test B against my previous commit to check, and it failed exactly as you'd
expect — full 10 s spin, remainder lost:

```
givenZeroIntervalAndThreshold_... FAILED  (Time elapsed: 10.82 s)
could not find: payload=10, 11, 12, 13, 14
```

`ScheduledAcknowledgementExecution.start()` only arms when
`ackInterval != Duration.ZERO`, so there was nothing for `shouldKeepSchedulingAcks()`
to keep alive. Making the drain part of the shutdown path covers both, so I
dropped that change entirely.

Implemented as you described, gated on `acks.isEmpty()` and inside the loop:

```java
private void flushRemainingAcks() {
    if (!this.acks.isEmpty()) {
        return;
    }
    this.context.lock();
    try {
        this.context.executeAllAcks();
    }
    finally {
        this.context.unlock();
    }
}
```

The futures go through `manageFuture()`, so `hasUnfinishedAcks()` sees them and
the wait loop keeps waiting until the flush actually completes.

**2. `volatile`.** Done, with a comment on the field. Agreed on the reasoning —
the write is under `lifecycleMonitor` but `isRunning()` reads it unsynchronized,
so neither the polling nor the scheduler thread had a guarantee of observing
`stop()`.

I could not reproduce the ~1-in-3 rate, though. On JDK 17.0.19 / arm64, 15
consecutive runs each:

| | result |
| --- | --- |
| with this change | 15/15 pass |
| `main` + `volatile` only | 15/15 reproduce |
| `main` unchanged | 15/15 reproduce |

So the visibility of `running` made no observable difference here — presumably
whether the JIT hoists the `isRunning()` read out of the loop differs between our
setups. That doesn't change anything about the fix being right; it just means the
old test's determinism rested on unspecified behaviour, as you said. Worth
mentioning in case the difference matters to you.

**3. Tests.** Replaced my test with your A and B. A uses a latch in the
acknowledgement executor to establish that the threshold flush completed before
the remainder is acked, and an interval that outlasts the shutdown timeout, so a
scheduled execution exists but provably cannot fire during the test. Both fail on
`main` and pass with the change.

Full module is green: 636 tests, 0 failures, 6 skipped, spotless checks enabled,
LocalStack integration tests included.

---

One thing I noticed while working through your "the polling thread can hold a
just-polled message" point, which I have deliberately left out of scope. The same
window is also reachable from the wait loop's exit condition:

```java
Message<T> polledMessage = this.acks.poll(1, TimeUnit.SECONDS);   // removed from acks
if (polledMessage != null) {
    addMessageToBuffer(polledMessage);                            // not in the buffer yet
```

In between, the message is in neither `acks` nor `acksBuffer`, so `hasAcksLeft()`
can report false, the wait loop exits, `waitAcknowledgementsToFinish()` sets
`isTimeoutElapsed` and clears the buffer, and the polling thread then adds the
message to a buffer nothing will flush. It is dropped with no warning at all,
unlike the timeout case. The window is narrow, but it widens under contention on
`context.lock()`. Happy to open a separate issue if you think it's worth
addressing.

tomazfernandes:
Thanks for the issue and PR @hyeongguen-song.

This addresses both the sub-threshold remainder scenario you initially brought up, the interval = ZERO case, and makes `running` volatile to fix stale reads during shutdown.

> Happy to open a separate issue if you think it's worth addressing.

Yes, let's squash this one as well. It's a narrow window that can only affect one message per shutdown, but as you said under contention it widens up.

PR #1666, opened 2026-08-06
Preserve generic SQS listener payload types

### LINKED ISSUE #1597: SQS Listener deserializes POJOs with generics types incorrectly
**Type**: Bug

**Component**:
SQS but may be also others

**Describe the bug**
`@SqsListener` methods fail to correctly deserialize payloads with generic types. E.g.:

1. `Message<List<TestPayload>>` — list elements are deserialized as `LinkedHashMap` instead of `TestPayload`
2. `Wrapper<TestPayload>` (generic outer type) — the generic field is deserialized as `LinkedHashMap` instead of `TestPayload`

Problem appears from version 4.0.0.

**Sample**
Test application provided [sqs-deserialization-main.zip](https://github.com/user-attachments/files/26371527/sqs-deserialization-main.zip).

brun0-4ugusto:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

This PR fixes generic payload deserialization for `@SqsListener` methods.

Previously, payload type inference retained only the raw `Class<?>`. Types such as `GenericWrapper<TestEvent>` and `Message<List<TestEvent>>` were therefore reduced to `GenericWrapper.class` and `List.class`. Without the complete generic type, Jackson deserialized nested values as `LinkedHashMap` instances.

This change:

- Introduces `MethodPayloadMetadata`, containing the raw payload class and an optional conversion hint.
- Preserves the listener's `MethodParameter` as the `SmartMessageConverter` conversion hint.
- Propagates the payload class and conversion hint through the endpoint, listener container, message source, and `MessageConversionContext`.
- Uses `SmartMessageConverter#fromMessage(Message, Class, Object)` when a conversion hint is available.
- Falls back to the regular `MessageConverter` contract for converters that do not implement `SmartMessageConverter`.
- Preserves custom payload type mapper precedence over inferred listener metadata.
- Supports nested generic payloads and batch elements, including:
  - `GenericWrapper<TestEvent>`
  - `GenericWrapper<List<TestEvent>>`
  - `Message<List<TestEvent>>`
  - `List<GenericWrapper<TestEvent>>`
  - `List<Message<GenericWrapper<TestEvent>>>`
- Keeps existing `MethodPayloadTypeInferrer` implementations source-compatible through a default metadata adaptation method.
- Updates the reference documentation and the `MessageSource` Javadoc to describe where payload conversion occurs.

`@SqsListener` payload deserialization currently loses generic type information because the inferred listener type is transported only as a raw `Class<?>`.

As a result, Jackson cannot determine the concrete type of generic fields or collection elements and falls back to `LinkedHashMap`.

The problem is also visible before listener invocation because SQS payload conversion happens at the `MessageSource` level. Therefore, interceptors, error handlers, and acknowledgement callbacks may also receive incorrectly typed generic values.

Spring's `SmartMessageConverter` already supports a conversion hint. Its base converters can use a `MethodParameter` hint to recover the complete generic `Type`, so this PR propagates that existing Spring metadata instead of introducing custom Jackson-specific type resolution.

Fixes #1597

Added and executed focused tests covering:

- Payload metadata inference for simple, wrapper, message, collection, and batch listener parameters.
- Backwards compatibility for custom `MethodPayloadTypeInferrer` implementations.
- Propagation through endpoint, container, message source, and conversion context.
- `SmartMessageConverter` hint invocation and regular `MessageConverter` fallback.
- Custom payload type mapper precedence.
- Conversion hint cleanup when the payload type is reconfigured.
- Generic and nested generic deserialization with both Jackson 3 and the legacy Jackson 2 converter.
- LocalStack integration scenarios for wrappers, nested collections, message wrappers, batches, interceptors, and acknowledgement callbacks.

Validation results:

- Unit and converter tests passed.
- All `SqsPayloadTypeInferenceIntegrationTests` passed with LocalStack.
- Spotless validation for all changed Java files passed.
- The complete SQS module suite executed 658 tests. 

Commands used:
`./mvnw -pl spring-cloud-aws-sqs -am test` Full module: Tests run: 658, Failures: 0, Errors: 0, Skipped: 5

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

Subject to maintainer feedback, a separate follow-up PR could expand generic payload inference for listener methods inherited from generic superclasses.

For example:

```java
abstract class GenericListener<T> {

    @SqsListener("events")
    void listen(GenericWrapper<T> event) {
    }
}

class TestEventListener extends GenericListener<TestEvent> {
} 
```

When Spring discovers the listener method, its original declaration still describes the payload as GenericWrapper<T>. To deserialize it as GenericWrapper<TestEvent>, the MethodParameter must also be resolved against the concrete listener class (TestEventListener). This allows Spring's type resolution infrastructure to substitute T with TestEvent before passing the conversion hint to the SmartMessageConverter.
I have already explored and implemented this extension in a separate branch, including tests for inherited generic listener methods and nested or batch generic payload shapes. It is deliberately not included in this PR so that the initial fix remains focused and easier to review.
We can open a discussion about the expected scope and compatibility requirements for inherited generic listeners. If the maintainers agree with the direction, the existing implementation can be refined and submitted as a separate follow-up PR.

### DIFF
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/config/MethodPayloadTypeInferrer.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/config/MethodPayloadTypeInferrer.java
@@ -40,4 +40,20 @@ public interface MethodPayloadTypeInferrer {
 	@Nullable
 	Class<?> inferPayloadType(Method method, @Nullable List<HandlerMethodArgumentResolver> argumentResolvers);
 
+	/**
+	 * Infer payload metadata from the given method and its argument resolvers.
+	 * <p>
+	 * The default implementation adapts existing {@link MethodPayloadTypeInferrer} implementations by returning the
+	 * inferred class without a conversion hint.
+	 * @param method the listener method
+	 * @param argumentResolvers the argument resolvers available for this method, may be null or empty
+	 * @return the inferred payload metadata, or null if it cannot be determined
+	 */
+	@Nullable
+	default MethodPayloadMetadata inferPayloadMetadata(Method method,
+			@Nullable List<HandlerMethodArgumentResolver> argumentResolvers) {
+		Class<?> payloadClass = inferPayloadType(method, argumentResolvers);
+		return payloadClass != null ? new MethodPayloadMetadata(payloadClass, null) : null;
+	}
+
 }
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/config/MethodPayloadTypeInferrer.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/config/MethodPayloadTypeInferrer.java
@@ -33,11 +33,29 @@ public interface MethodPayloadTypeInferrer {
 
 	/**
 	 * Infer the payload class from the given method and its argument resolvers.
+	 * @deprecated in favor of {@link #inferPayloadMetadata(Method, List)}
 	 * @param method the listener method
 	 * @param argumentResolvers the argument resolvers available for this method, may be null or empty
 	 * @return the inferred payload class, or null if it cannot be determined
 	 */
 	@Nullable
+	@Deprecated
 	Class<?> inferPayloadType(Method method, @Nullable List<HandlerMethodArgumentResolver> argumentResolvers);
+
+	/**
+	 * Infer payload metadata from the given method and its argument resolvers.
+	 * <p>
+	 * The default implementation adapts existing {@link MethodPayloadTypeInferrer} implementations by returning the
+	 * inferred class without a conversion hint.
+	 * @param method the listener method
+	 * @param argumentResolvers the argument resolvers available for this method, may be null or empty
+	 * @return the inferred payload metadata, or null if it cannot be determined
+	 */
+	@Nullable
+	default MethodPayloadMetadata inferPayloadMetadata(Method method,
+			@Nullable List<HandlerMethodArgumentResolver> argumentResolvers) {
+		Class<?> payloadClass = inferPayloadType(method, argumentResolvers);
+		return payloadClass != null ? new MethodPayloadMetadata(payloadClass, null) : null;
+	}
 
 }
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/AbstractMessageListenerContainer.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/AbstractMessageListenerContainer.java
@@ -74,6 +74,9 @@ public abstract class AbstractMessageListenerContainer<T, O extends ContainerOpt
 	@Nullable
 	private Class<?> payloadDeserializationType;
 
+	@Nullable
+	private Object payloadConversionHint;
+
 	/**
 	 * Create an instance with the provided {@link ContainerOptions}
 	 * @param containerOptions the options instance.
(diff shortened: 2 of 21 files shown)


brun0-4ugusto:
Hi, everyone! How are you?

Are there any updates on this PR? Please let me know if you need any additional information from me.

I’m looking forward to this feature because I’d like to use it in a project at my company. 😊

Thank you, and have a lovely day!

@tomazfernandes

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/source/AbstractMessageConvertingMessageSource.java
tomazfernandes, line 87: Let's have this method delegate to the new overload with a `null` conversion hint.

That way both setters converge on the 3-arg hook, whose default already delegates to this 2-arg hook, so delegation flows in one direction and existing 2-arg overrides remain invoked from every entry point.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/source/AbstractSqsMessageSource.java
```diff
@@ -137,8 +137,16 @@ private void configureConversionContextAndAcknowledgement() {
 
 	@Override
 	protected void doConfigurePayloadTypeOnContext(Class<?> payloadType, MessageConversionContext context) {
-		ConfigUtils.INSTANCE.acceptIfInstance(context, SqsMessageConversionContext.class,
-				ctx -> ctx.setPayloadClass(payloadType));
+		doConfigurePayloadTypeOnContext(payloadType, null, context);
```
tomazfernandes: For backwards compatibility, let's invert the delegation here: keep the payload class assignment in the 2-arg override, and have the 3-arg override call it before setting the hint.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/listener/AbstractMessageListenerContainer.java
```diff
@@ -190,6 +193,21 @@ public void setPhase(int phase) {
 	 */
 	public void setPayloadDeserializationType(@Nullable Class<?> payloadDeserializationType) {
 		this.payloadDeserializationType = payloadDeserializationType;
```
tomazfernandes: Let's delegate to `setPayloadDeserializationType` with a null conversion hint, and have the 2-arg overload assign both fields directly instead of calling this method.

That makes the richer overload the single write site, so both fields are updated together.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/config/MethodPayloadTypeInferrer.java
```diff
@@ -40,4 +40,20 @@ public interface MethodPayloadTypeInferrer {
 	@Nullable
```
tomazfernandes, line 41: Let's deprecate this method, routing implementations to the new method instead.

tomazfernandes (commented):
@brun0-4ugusto thanks for the PR, looking good.

Left a few comments around backwards compatibility for existing overrides and keeping the delegation between the old and new overloads one-directional.

brun0-4ugusto:
Thank you for the feedback @tomazfernandes ! I really appreciated the points you raised in the review. I’m already working on the requested changes, and I’ll update the PR with the fixes later this week.

brun0-4ugusto:
Hi @tomazfernandes ! Thanks again for the review. I’ve addressed the comments and updated the PR accordingly.

Regarding MethodPayloadTypeInferrer, I deprecated [`inferPayloadType(...)`](https://github.com/awspring/spring-cloud-aws/pull/1666/changes#diff-0a6942b574dbaa0a302b788a9ecb97199d4553f4b48338f379904019a0b63a55R42) and routed the framework implementation through the [`new inferPayloadMetadata(...)`](https://github.com/awspring/spring-cloud-aws/pull/1666/changes#diff-0a6942b574dbaa0a302b788a9ecb97199d4553f4b48338f379904019a0b63a55R55) method. 

I intentionally kept `inferPayloadType(...)` as the SAM of the `@FunctionalInterface` to preserve backward compatibility with existing custom implementations and lambdas. Making `inferPayloadMetadata(...)` the new abstract method would change the functional interface contract and could break existing implementations.

Could you please take another look when you have a chance? In particular, I’d appreciate your feedback on whether preserving the existing SAM is the direction you had in mind, or whether you would prefer changing it despite the compatibility impact.

Thank you!  :)

tomazfernandes (approved):

tomazfernandes:
Thanks for the PR @brun0-4ugusto. Let's keep the SAM as is to keep compatibility. We'll flip in the next major version when we remove the deprecated method.

Looking forward to more.

PR #1676, opened 2026-08-08
Speed up SQS error handler tests

tomazfernandes:
## Motivation

`SqsErrorHandlerIntegrationTests` currently takes about 99 seconds on CI.

## Changes

**Shorter backoff sequences.** All eight backoff listeners now fail for three retries rather than four, five or six. Exponential jitter goes from timeouts of 2, 4, 8, 16, 32, 64 down to 2, 4, 8, which the mocked random reduces to waits of 1, 2 and 4 seconds. The plain exponential listener drops its initial value from 2 seconds to 1. Three retries still exercise the growth of the sequence across three intervals.

**Removed the redundant minimum wait.** The listeners already compare the elapsed time reported by SQS against the expected backoff total and only count the latch down when it is met, so `await().atLeast(...)` in the test methods asserted the same property a second time while setting a wall-clock floor. The tests now wait only for the latch, with a ceiling.

Assertion accuracy is unaffected. Each test still requires the observed redelivery interval to match the computed backoff for its strategy, over a shorter sequence.

## Result

Run on its own the class goes from about 108 seconds to 19. Across six consecutive local runs of the full module, all green with the test count unchanged at 684, the module averages 1 minute 21 seconds.


tomazfernandes:
CI results after the rerun, comparing against the same job type on main.

The class goes from 98.7s to 43.9s and the SQS module from 2:58 to 1:57.

PR #1695, opened 2026-08-30
feat: auto-configure virtual threads for SQS listener containers

### LINKED ISSUE #1693: Enhancement: Auto-configure Virtual Threads for SQS/Kinesis Listener Containers
### Description
Spring Boot 3.2+ introduced the `spring.threads.virtual.enabled` property to automatically configure Virtual Threads for standard listeners (like Spring Kafka and Spring RabbitMQ). 

Currently, Spring Cloud AWS SQS containers can use Virtual Threads if manually configured via container options, but they do not automatically hook into this standard Spring Boot property out-of-the-box.

### Proposed Solution
1. In `SqsAutoConfiguration`, detect if `spring.threads.virtual.enabled` is set to `true`.
2. If enabled and running on JDK 21+, configure the default `SqsMessageListenerContainerFactory` to use `SimpleAsyncTaskExecutor` with `virtualThreads = true` for both `componentsTaskExecutor` and `acknowledgementResultTaskExecutor`.
3. Provide similar automatic fallback in Kinesis message-driven adapters.

This improves startup/concurrency performance out-of-the-box on Java 21 stacks with minimal configuration.

Theprathamshah:
- [x] Enhancement

## :scroll: Description
This pull request adds native support for Java 21+ Virtual Threads to SQS listener container factories.
* Modifies `SqsAutoConfiguration` to detect the standard Spring Boot property `spring.threads.virtual.enabled`.
* When enabled, configures `SimpleAsyncTaskExecutor` (with `virtualThreads = true`) for both the message processing executor (`componentsTaskExecutor`) and the acknowledgment executor (`acknowledgementResultTaskExecutor`).

Closes #1693.
Integrates Spring Boot's global standard `spring.threads.virtual.enabled` toggle directly into Spring Cloud AWS SQS listeners, allowing users to leverage high-throughput non-blocking concurrency (Project Loom) out-of-the-box on Java 21+ stacks.

1. Added unit tests in `SqsAutoConfigurationTest` verifying SQS containers operate on virtual threads when enabled and fallback to standard platform thread execution when disabled.
2. Verified all tests compile and pass successfully.

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sqs/SqsAutoConfiguration.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/io/awspring/cloud/autoconfigure/sqs/SqsAutoConfiguration.java
@@ -57,6 +57,8 @@ import org.springframework.boot.context.properties.PropertyMapper;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
+import org.springframework.core.env.Environment;
+import org.springframework.core.task.SimpleAsyncTaskExecutor;
 import software.amazon.awssdk.services.s3.S3AsyncClient;
 import software.amazon.awssdk.services.sqs.SqsAsyncClient;
 import software.amazon.awssdk.services.sqs.model.Message;
@@ -186,7 +188,7 @@ public class SqsAutoConfiguration {
 			ObjectProvider<AcknowledgementResultCallback<Object>> acknowledgementResultCallback,
 			ObjectProvider<AsyncAcknowledgementResultCallback<Object>> asyncAcknowledgementResultCallback,
 			ObjectProvider<JacksonMessageConverterMigration> messageConverterFactory,
-			MessagingMessageConverter<?> messagingMessageConverter) {
+			MessagingMessageConverter<?> messagingMessageConverter, Environment environment) {
 
 		SqsMessageListenerContainerFactory<Object> factory = new SqsMessageListenerContainerFactory<>();
 		factory.configure(this::configureProperties);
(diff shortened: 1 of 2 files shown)


tomazfernandes:
@Theprathamshah , thanks for the PR. 

Let's keep this PR scoped to SQS, and please open a separate one for Kinesis and other integrations.

Theprathamshah:
@tomazfernandes Sure let me create separate PR for kinesis.

Theprathamshah:
@tomazfernandes I've separated the changes as discussed. This PR focuses exclusively on the SQS listener container auto-configuration (the Kinesis changes are in #1696). 

Both are ready for review whenever you have time. Thanks!

Theprathamshah:
Hi @tomazfernandes @MatejNedic ,just following up to see if you have any feedback or suggestions on this PR when you get a moment.

Also, could you please approve the CI workflow when convenient so we can ensure the automated checks pass? Thanks!

tomazfernandes:
@Theprathamshah I approved the CI workflow for this PR. There seems to be some failing tests, please take a look.

Theprathamshah:
Hi @tomazfernandes, thanks for pointing that out! I've updated the tests to use reflection for isVirtual() and added @EnabledForJreRange(min = JRE.JAVA_21) so they compile and run cleanly across the entire JDK 17 / 21 / 24 matrix.

tomazfernandes (changes requested):
Thanks for the update @Theprathamshah, tests pass now.

The issue I see with this approach is that in this PR the auto-configuration reads `spring.threads.virtual.enabled` directly and does not check the JDK version. 

On JDK 17 with the property set to `true`, `SimpleAsyncTaskExecutor.setVirtualThreads(true)` throws `UnsupportedOperationException` and the application context fails to start.  CI does not catch it because the only test that sets the property to `true` is restricted to JDK 21+.

Spring Boot treats the property as a no-op below 21, so applications on 17 that carry it in shared configuration would start failing after upgrading, which would be a regression.

Please use Spring Boot's threading condition (`@ConditionalOnThreading(Threading.VIRTUAL)` or `Threading.VIRTUAL.isActive(environment)`), which [checks both the property and the JDK version](https://github.com/spring-projects/spring-boot/blob/dc1daa53bf5719e9906d65cccaf42950ff9fab78/core/spring-boot/src/main/java/org/springframework/boot/thread/Threading.java#L45-L52), as Boot's [Kafka](https://github.com/spring-projects/spring-boot/blob/dc1daa53bf5719e9906d65cccaf42950ff9fab78/module/spring-boot-kafka/src/main/java/org/springframework/boot/kafka/autoconfigure/KafkaAnnotationDrivenConfiguration.java#L125-L131) and [RabbitMQ](https://github.com/spring-projects/spring-boot/blob/dc1daa53bf5719e9906d65cccaf42950ff9fab78/module/spring-boot-amqp/src/main/java/org/springframework/boot/amqp/autoconfigure/RabbitAnnotationDrivenConfiguration.java#L77-L84) integrations do. 

Please also add a test that runs on Java 17, sets the property to `true`, and and asserts that the containers fall back to the default platform thread executors (no custom executors set in the container options). 

Let me know if you have any questions.

PR #1699, opened 2026-09-04
Map SNS message attributes to SQS listener headers

### LINKED ISSUE #1111: Missing SNS attributes when sending a message via SnsTemplate and receiving via @SqsListener
**Type**: Bug

**Component**: SNS, SQS

**Description**
Reference documentation [mentions](https://docs.awspring.io/spring-cloud-aws/docs/3.1.1/reference/html/index.html#sns-messages) that when we receive SNS messages through the `@SqsListener`, the message includes all attributes of the `SnsNotification`. However, the attributes are not passed to message headers.

**Sample**
Having an SQS queue that is subscribed to SNS topic (raw message delivery is **disabled**), when sending to SNS via `snsTemplate` as following
```
 snsTemplate.convertAndSend(topicName, pojo, Map.of("my-attribute", "attribute-value"));
```
and receiving the message via `@SqsListener` as following
```
 @SqsListener("${queueName}")
 public void processGreetingsEvent(Message<?> message){
     System.out.println(message.getHeaders().get("my-attribute"))
 }
 ```
 then **null** is printed instead of **"attribute-value"**.
 
 The same effect is seen when receiving as following
 ```
 @SqsListener("${queue.name}")
 public void processGreetingsEvent(@SnsNotificationMessage Pojo pojo, @Headers Map<String, Object> headers) {
     headers.get("my-attribute"))
 }
 ```

### PR DESCRIPTION
**Pull request by azunox**

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Maps SNS `MessageAttributes` embedded in a non-raw SNS notification envelope to Spring Message headers when the notification is received through SQS.

The change:

- Detects SNS notification envelopes before extracting attributes.
- Maps String, Number, and Binary SNS attribute values to their corresponding header value types.
- Preserves existing SQS message attributes when an SNS attribute has the same name.
- Leaves regular SQS messages and non-SNS JSON payloads unchanged.
- Avoids parsing message bodies that do not contain `MessageAttributes`.

When raw message delivery is disabled for an SNS-to-SQS subscription, SNS message attributes are stored inside the SNS JSON envelope rather than as native SQS message attributes.

`SqsHeaderMapper` previously mapped only native SQS message attributes. Consequently, attributes published through `SnsTemplate` were unavailable from `Message#getHeaders()` or an `@Headers` argument in an `@SqsListener`.

This change promotes the nested SNS attributes to Spring Message headers while retaining native SQS attributes as the authoritative value in case of a name collision.

Closes #1111

Added focused `SqsHeaderMapperTests` covering:

- String, Number, and Binary SNS message attribute mapping.
- Precedence of native SQS message attributes over nested SNS attributes.
- Rejection of `MessageAttributes` found in non-SNS JSON payloads.

Executed:

`./mvnw -pl spring-cloud-aws-sqs -am -Dtest=SqsHeaderMapperTests -Dsurefire.failIfNoSpecifiedTests=false test`

Result: 33 tests run, 0 failures, 0 errors, 0 skipped. The full repository and LocalStack integration test suites were not executed.

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

Run the complete CI and integration test suites through the repository pull request workflow.

### DIFF
--- a/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SqsHeaderMapper.java
+++ b/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SqsHeaderMapper.java
@@ -23,6 +23,7 @@ import io.awspring.cloud.sqs.listener.SqsHeaders;
 import io.awspring.cloud.sqs.support.converter.legacy.LegacyJackson2SqsMessagingMessageConverter;
 import java.nio.ByteBuffer;
 import java.time.Instant;
+import java.util.Base64;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
@@ -42,6 +43,10 @@ import software.amazon.awssdk.services.sqs.SqsAsyncClient;
 import software.amazon.awssdk.services.sqs.model.Message;
 import software.amazon.awssdk.services.sqs.model.MessageAttributeValue;
 import software.amazon.awssdk.services.sqs.model.MessageSystemAttributeName;
+import tools.jackson.core.JacksonException;
+import tools.jackson.core.type.TypeReference;
+import tools.jackson.databind.JsonNode;
+import tools.jackson.databind.json.JsonMapper;
 
 /**
  * A {@link HeaderMapper} implementation for SQS {@link Message}s. Enables creating additional SQS related headers from
@@ -58,6 +63,11 @@ import software.amazon.awssdk.services.sqs.model.MessageSystemAttributeName;
 public class SqsHeaderMapper implements ContextAwareHeaderMapper<Message> {
 
 	private static final Logger logger = LoggerFactory.getLogger(SqsHeaderMapper.class);
+
+	private static final TypeReference<Map<String, SnsNotification.MessageAttribute>> SNS_MESSAGE_ATTRIBUTES_TYPE = new TypeReference<>() {
+	};
+
+	private final JsonMapper jsonMapper = new JsonMapper();
 
 	private BiFunction<Message, MessageHeaderAccessor, MessageHeaders> additionalHeadersFunction = ((message,
 			accessor) -> accessor.toMessageHeaders());
@@ -163,6 +173,7 @@ public class SqsHeaderMapper implements ContextAwareHeaderMapper<Message> {
 		accessor.copyHeadersIfAbsent(getMessageAttributesAsHeaders(source));
 		accessor.copyHeadersIfAbsent(createDefaultHeaders(source));
 		accessor.copyHeadersIfAbsent(createAdditionalHeaders(source));
+		accessor.copyHeadersIfAbsent(getSnsMessageAttributesAsHeaders(source));
 
 		MessageHeaders messageHeaders = accessor.toMessageHeaders();
 		logger.trace("Mapped headers {} for message {}", messageHeaders, source.messageId());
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SqsHeaderMapper.java [hunk 1]
```diff
@@ -59,6 +64,11 @@ public class SqsHeaderMapper implements ContextAwareHeaderMapper<Message> {
 
 	private static final Logger logger = LoggerFactory.getLogger(SqsHeaderMapper.class);
 
+	private static final TypeReference<Map<String, SnsNotification.MessageAttribute>> SNS_MESSAGE_ATTRIBUTES_TYPE = new TypeReference<>() {
+	};
+
+	private final JsonMapper jsonMapper = new JsonMapper();
```
MatejNedic, line 70: JsonMapper should be the same one that is passed in autoconfiguration.

spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/support/converter/SqsHeaderMapper.java
```diff
@@ -163,6 +173,7 @@ public MessageHeaders toHeaders(Message source) {
 		accessor.copyHeadersIfAbsent(getMessageAttributesAsHeaders(source));
 		accessor.copyHeadersIfAbsent(createDefaultHeaders(source));
 		accessor.copyHeadersIfAbsent(createAdditionalHeaders(source));
+		accessor.copyHeadersIfAbsent(getSnsMessageAttributesAsHeaders(source));
```
MatejNedic, line 176: This runs on every SQS message, whether it's an SNS notification or not. `SqsHeaderMapper` is the default mapper for all SQS traffic, so this adds an SNS-envelope check (and a JSON parse on match) to the hot path for everyone, even deployments that never receive SNS messages.

Custom `SnsAwareSqsHeaderMapper` should be implemented and  wired via existing hooks such as `AbstractMessagingMessageConverter.setHeaderMapper` and `AbstractMessagingMessageConverter.configureHeaderMapper`.

Container options can be used then to configure this custom mapper for Queues which will have messages sent via `SnsTemplate`. This makes it cheaper for performance.

MatejNedic (changes requested):
Hey @azunox ,

Thanks on PR! I see few things I personally would do differently. 

@tomazfernandes  what do you think?

azunox:
Hello! Thanks for the review, @MatejNedic 

I agree that SNS-specific parsing should not run for all SQS messages.

I propose moving the logic into an opt-in SnsAwareSqsHeaderMapper that receives the application's configured JsonMapper, with focused tests and a configuration example for queues receiving SNS notifications.

does this approach look right to you before I update the PR?

tomazfernandes:
Hey @azunox, thanks for the PR, and thanks @MatejNedic for the suggestions.

I'll take a closer look shortly.

tomazfernandes:
Hey @azunox and @MatejNedic, agreed on the new Mapper opt-in direction.

This can be a Jackson 3-only feature since Jackson 2 support is deprecated, let's have that explicit in the docs.

azunox:
Thanks for confirming the direction, @tomazfernandes, and thanks for the suggestions, @MatejNedic. 
I've updated the PR accordingly:

- Moved SNS envelope parsing into an opt-in `SnsAwareSqsHeaderMapper`. The default `SqsHeaderMapper` no longer checks or parses message bodies for SNS attributes.
- Made the new mapper accept the application's configured Jackson 3 `JsonMapper` through its constructor, rather than creating its own instance.
- Added a dedicated listener factory example using `setHeaderMapper` and container options, passing the same `JsonMapper` to both the message converter and the header mapper. Only listeners selecting that factory opt in to SNS attribute mapping.
- Explicitly documented that this is a Jackson 3-only feature and that no Jackson 2 implementation is provided, as Jackson 2 support is deprecated.

The changes are split into three commits: `b60fc7e9` (opt-in mapper), `c214eca0` (configured `JsonMapper` and configuration example), and `82144bf1` (Jackson 3 documentation).

Focused tests cover attribute type conversion, SQS header precedence, preservation of additional headers and message IDs, non-SNS and malformed bodies, opt-in converter configuration, and use of the supplied mapper's parsing configuration.

Validation:

```bash
./mvnw -B -pl spring-cloud-aws-sqs -am \
  -Dtest=SqsHeaderMapperTests,SnsAwareSqsHeaderMapperTests \
  -Dsurefire.failIfNoSpecifiedTests=false test
```

All **44 tests passed**, with no failures, errors, or skipped tests. The build's Spotless checks also passed. The full repository and LocalStack integration test suites were not run.

Please let me know if you'd like any further adjustments.

